PR #8, opened 2026-06-06
Fix YYYYMMDDHHMMSS provider_start not parsed for timeshift URL

### LINKED ISSUE #6: */- timezone shift issue
Nice tool, could become useful.

Did a quick test:

a) Download seemed to be be a bit slow ... tried with wget for the same show and got 4.5x speed ... not sure how you implemented
b) Unfortunately downloaded a totally different time sequence.  

Docker has TZ=Europe/Berlin --> currently CEST = UTC+2

EPG for the show had:

       {
        "$": {
          "channel": "503739",
          "start": "20260420170000 +0000",
          "stop": "20260420181500 +0000"
        },
        "title": [
          "Title"
        ],
        "desc": [
          "some text"
        ]
      }

UI Displayed time in UI was 100% correct (19:00 - 20:15)!

So intention was to download 19:00 - 20:15. --> +2h from EPG time
Downloaded was a show that was aired at 15:00 --> -2h  from EPG time

Direkt download just using `&start=2026-04-20:19-00&duration=75` delivered the correct content.

razzamatazm:
## What a user would notice

Downloads grabbed the wrong show, typically shifted 2+ hours from what was selected in the EPG. This matched issue #6, where a user in UTC+2 (Europe/Berlin) saw the correct time in the UI (19:00 CEST) but got content that aired 2 hours earlier.

## Root cause

Two gaps in the provider-start normalization:

1. **Xtream Codes API** often returns the EPG `start` field as a 14-digit compact string like `20260420190000` (YYYYMMDDHHMMSS, no dashes). The normalizer only recognised dash-delimited formats, so these tokens were rejected and the download URL fell back to the UTC timestamp instead of the provider's local wall-clock time. On providers whose timeshift endpoint expects local time, that produced a download offset by the full UTC-to-local difference.

2. **XMLTV ingest** never stored the raw `start` / `stop` attributes from the XML as `provider_start` / `provider_stop`. Programs ingested from XMLTV always had `NULL` provider_start, so the URL always used the UTC fallback even when the XMLTV contained a non-UTC offset (e.g., `20260420190000 +0200`) that encodes the correct local time.

## What changed

- `_normalize_provider_start_token` now strips any trailing timezone suffix (`+HHMM` / `-HHMM`) from compact tokens and tries `%Y%m%d%H%M%S` / `%Y%m%d%H%M` after the existing dash-delimited formats. The bare datetime portion is the provider's local wall-clock time, which is what the timeshift URL needs.
- `_iter_programs` in `epg_ingest_manager` now yields `provider_start` and `provider_stop` from the raw XMLTV attributes so those programs behave the same as API-backfilled ones.

## How it was tested

Two new tests in `backend/tests/test_catchup_resolution.py`:

- `test_yyyymmddhhmmss_provider_token_no_tz_is_used_as_url_start`: provider_start `"20260420190000"` (API, no TZ) produces URL `.../75/2026-04-20:19-00/999.ts` instead of the UTC fallback `2026-04-20:17-00`.
- `test_yyyymmddhhmmss_provider_token_with_tz_offset_uses_local_part`: provider_start `"20260420190000 +0200"` (XMLTV local TZ) strips the offset and produces the same correct URL.

All 29 existing tests still pass.

**Before:** URL used UTC fallback `2026-04-20:17-00` for YYYYMMDDHHMMSS tokens, downloading the wrong content.
**After:** URL correctly uses `2026-04-20:19-00` (the provider's local time), downloading the right show.

Closes #6 (partial: covers providers returning YYYYMMDDHHMMSS local time; the case of a provider whose XMLTV uses plain UTC `+0000` and whose timeshift endpoint also expects local time is a separate issue that needs a per-account offset setting).

### DIFF
--- a/backend/services/download_builder.py
+++ b/backend/services/download_builder.py
@@ -66,7 +66,7 @@ def _normalize_provider_start_token(provider_start: str, pre_padding_minutes: in
         # Strip trailing timezone offset (e.g. " +0200", " -0500", " +0000") before
         # trying compact formats.  The datetime part without the offset is the
         # provider's local wall-clock time, which is what the timeshift URL expects.
-        bare = re.sub(r"\s+[+-]\d{4}$", "", token).strip()
+        bare = re.sub(r"\s*([+-]\d{2}:?\d{2}|Z)$", "", token).strip()
         for fmt, candidate in (
             ("%Y-%m-%d %H:%M:%S", bare),
             ("%Y-%m-%d %H:%M",    bare),
--- a/backend/services/download_builder.py
+++ b/backend/services/download_builder.py
@@ -63,14 +63,20 @@ def _normalize_provider_start_token(provider_start: str, pre_padding_minutes: in
     if re.fullmatch(r"\d{4}-\d{2}-\d{2}:\d{2}-\d{2}", token):
         parsed = datetime.strptime(token, "%Y-%m-%d:%H-%M")
     else:
-        for fmt in (
-            "%Y-%m-%d %H:%M:%S",
-            "%Y-%m-%d %H:%M",
-            "%Y-%m-%dT%H:%M:%S",
-            "%Y-%m-%dT%H:%M",
+        # Strip trailing timezone offset (e.g. " +0200", " -0500", " +0000") before
+        # trying compact formats.  The datetime part without the offset is the
+        # provider's local wall-clock time, which is what the timeshift URL expects.
+        bare = re.sub(r"\s+[+-]\d{4}$", "", token).strip()
+        for fmt, candidate in (
+            ("%Y-%m-%d %H:%M:%S", bare),
+            ("%Y-%m-%d %H:%M",    bare),
+            ("%Y-%m-%dT%H:%M:%S", bare),
+            ("%Y-%m-%dT%H:%M",    bare),
+            ("%Y%m%d%H%M%S",      bare),
+            ("%Y%m%d%H%M",        bare),
         ):
             try:
-                parsed = datetime.strptime(token, fmt)
+                parsed = datetime.strptime(candidate, fmt)
                 break
             except ValueError:
                 continue
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -456,6 +456,8 @@ class EPGIngestManager:
                 "end_time": end_utc,
                 "start_timestamp": start_ts,
                 "stop_timestamp": stop_ts,
+                "provider_start": start_raw,
+                "provider_stop": stop_raw,
                 "duration_minutes": duration_minutes,
                 "has_archive": info["has_archive"],
             }
--- a/backend/tests/test_catchup_resolution.py
+++ b/backend/tests/test_catchup_resolution.py
@@ -307,6 +307,81 @@ class CatchupResolutionTests(unittest.TestCase):
 
         self.assertIn("/75/2026-04-20:19-00/999.ts", download.source_url)
 
+    def test_yyyymmddhhmmss_provider_token_compact_offset_no_space_uses_local_part(self):
+        """Compact offset without whitespace (e.g. 20260420190000+0200) must strip correctly."""
+        download = asyncio.run(
+            self._build_download(
+                XtreamAccount(
+                    id=1,
+                    name="Provider B",
+                    server_url="https://provider.example.com",
+                    username="user",
+                    password="",
+                    guide_offset_hours=2,
+                ),
+                {
+                    "title": "Show",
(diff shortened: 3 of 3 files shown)


razzamatazm-cleo (changes requested):
Thanks for the clear write-up and the two regression tests. The core fix (recognising compact `YYYYMMDDHHMMSS` tokens and stripping a `" +HHMM"` suffix) is correct, the `provider_stop` column exists so the new ingest keys insert cleanly, and the API and XMLTV insert paths now stay in sync. One substantive gap remains before I can approve.

**1. Agree with Codex (P2): compact offsets without whitespace still fall through to the UTC fallback.**
`epg_ingest_manager._parse_xmltv_time` accepts a value by slicing `value[:14]`, so `20260420190000+0200` (no space, which real XMLTV feeds do emit) parses fine during ingest and is stored verbatim as `provider_start`. But the new strip in `_normalize_provider_start_token` is `re.sub(r"\s+[+-]\d{4}$", "", token)`, which only matches an offset preceded by whitespace. For the no-space form the regex does nothing, `%Y%m%d%H%M%S` then fails on the `+0200` tail, normalization returns `None`, and `build_download_from_program` falls back to the UTC start. That is exactly the 2-hour offset this PR set out to fix, still present for those feeds. The minimal fix is `\s*` instead of `\s+`:

    bare = re.sub(r"\s*[+-]\d{4}$", "", token).strip()

This stays safe for the existing formats: compact tokens with no offset have no trailing `[+-]\d{4}`, and dash-delimited forms don't end in `[+-]\d{4}` either.

**2. Codex missed: the strip is narrower than what ingest accepts.**
`_parse_xmltv_time` also accepts a trailing `Z` and colon offsets like `+02:00`. Stored raw, those tokens (`20260420190000Z`, `20260420190000 +02:00`) hit the same dead end in the normalizer and fall back to UTC. To fully close the inconsistency, widen the strip to cover all three shapes that ingest will let through, e.g.:

    bare = re.sub(r"\s*([+-]\d{2}:?\d{2}|Z)$", "", token).strip()

**3. Missing test for the gap.**
`test_yyyymmddhhmmss_provider_token_with_tz_offset_uses_local_part` only exercises the with-space variant, which gives false confidence that XMLTV TZ handling is covered. Per the repo's "test that fails before the fix" rule, please add a case with `provider_start="20260420190000+0200"` (no space), and ideally `...Z` and `...+02:00`, asserting the URL still uses `2026-04-20:19-00`.

Everything else looks good: logic, the UTC-fallback preservation for genuinely invalid tokens, and the column/insert mapping. Once the strip handles the no-whitespace (and ideally Z / colon) offsets and a regression test covers it, I'm happy to approve.

razzamatazm (commented):
Thanks for the clear breakdown. All three gaps addressed in c7ff4ea:

- Regex changed from `\s+[+-]\d{4}$` to `\s*([+-]\d{2}:?\d{2}|Z)$` — covers compact offset (no space), Z suffix, and colon offset, all three forms that `_parse_xmltv_time` accepts.
- Added three regression tests, one per variant (`+0200` no-space, `Z`, `+02:00`), each confirming URL uses `2026-04-20:19-00` and not the UTC fallback. All 32 tests pass.

razzamatazm-cleo (approved):
Re-reviewed at head c7ff4ea. All three gaps from my prior review are resolved, and I verified the supporting plumbing.

**Resolved:**
1. Regex widened from `\s+[+-]\d{4}$` to `\s*([+-]\d{2}:?\d{2}|Z)$`. This now covers the no-whitespace compact offset (`20260420190000+0200`), the `Z` suffix, and the colon offset (`+02:00`), matching exactly the three TZ shapes `_parse_xmltv_time` accepts. The earlier UTC-fallback bug for whitespace-less XMLTV feeds is closed.
2. The strip now matches what ingest lets through (Z and colon forms included).
3. Regression tests added for all variants: `+0200` no-space, `Z`, `+02:00`, plus the with-space and no-TZ cases. Each asserts the URL uses `2026-04-20:19-00` rather than the UTC fallback.

**Additionally verified this round:**
- Regex is safe against the pre-existing formats. The dash/space/`T` formats never end in a literal sign followed by digits, so `\s*[+-]\d{2}:?\d{2}$` cannot false-strip them; pure compact digit tokens have no sign to match. No regression to working tokens.
- `provider_stop` is a real column (`backend/models/epg_program.py:31-32`) and both `provider_start`/`provider_stop` are migrated on startup (`backend/database.py:60-64`), so the new ingest dict keys insert cleanly via `insert(EPGProgram)`.
- `epg_ingest_manager._iter_programs` stores the raw `start_raw`/`stop_raw`, keeping ingest and the download builder in sync.

PR #9, opened 2026-06-06
Fix VAAPI driver detection on AMD GPUs (sysfs class path)

### LINKED ISSUE #4: Amd Ryzen AI 9 HX 370
Hello
Trying to get the gpu of this CPU to encode but it is failing.

`DRM_IOCTL_VERSION, unsupported drm device by media driver: amdg DRM_IOCTL_VERSION, unsupported drm device by media driver: amdg [VAAPI @ 0x646b13f80f40] libva: /usr/local/lib/x86_64-linux-gnu/dri/iHD_drv_video.so init failed [VAAPI @ 0x646b13f80f40] Failed to initialise VAAPI connection: 18 (invalid parameter). Device creation failed: -5. Failed to set value '/dev/dri/renderD128' for option 'vaapi_device': Input/output error Error parsing global options: Input/output error`

Just wondered if anything I can do to get it to work.

razzamatazm:
## What a user would notice

Enabling VAAPI transcoding on an AMD GPU (including Ryzen integrated graphics) produced this error and no output:

```
[VAAPI] libva: /usr/local/lib/x86_64-linux-gnu/dri/iHD_drv_video.so init failed
[VAAPI] Failed to initialise VAAPI connection: 18 (invalid parameter)
Device creation failed: -5
```

The Intel driver (`iHD`) was being tried on an AMD device, which always fails. This was reported in issue #4.

## Root cause

`_resolve_vaapi_driver` tried to find the GPU's kernel driver by calling `device.resolve()` on `/dev/dri/renderD128`. That is a character device file, not a symlink, so `resolve()` returns the same path. Taking `.parent.parent` then gives `/dev`, not a sysfs directory. The driver-name lookup raised `OSError` silently, `kernel_driver` was left as `None`, and `LIBVA_DRIVER_NAME` was never set.

With no driver name set, libva searched the drivers path and picked up `iHD_drv_video.so` first (it comes from the LinuxServer ffmpeg image and sits at the front of `LIBVA_DRIVERS_PATH`). `iHD` does not support AMD devices, hence the crash.

The existing tests passed because they created the device as a symlink pointing into a fake sysfs tree. Real devices are not symlinks, so the tests did not catch this.

## What changed

- Added `VAAPI_SYSFS_DRM_CLASS = Path("/sys/class/drm")` class constant.
- `_resolve_vaapi_driver` now resolves `/sys/class/drm/renderD128` (a real symlink on every modern Linux kernel) to find the PCI device's sysfs subtree. The existing `amdgpu -> radeonsi` and `i915/xe -> iHD` mappings then work correctly.
- Tests updated to build a fake sysfs class directory and patch `VAAPI_SYSFS_DRM_CLASS`, matching the real kernel layout.

## How it was tested

All 27 tests pass (`python3 -m unittest discover -s tests`).

**Before:** `kernel_driver` always resolved to `None` on real hardware. `LIBVA_DRIVER_NAME` never set. libva chose `iHD`, crashing on AMD.

**After:** `/sys/class/drm/renderD128` resolves correctly. `kernel_driver = "amdgpu"`, `LIBVA_DRIVER_NAME=radeonsi` is passed to ffmpeg. AMD VAAPI encoding works.

Closes #4

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -55,6 +55,7 @@ class PostProcessor:
     """Handles transcoding and commercial detection/removal."""
 
     VAAPI_RENDER_DEVICE = Path("/dev/dri/renderD128")
+    VAAPI_SYSFS_DRM_CLASS = Path("/sys/class/drm")
     VAAPI_DEFAULT_DRIVERS_PATH = (
         "/usr/local/lib/x86_64-linux-gnu/dri:/usr/local/lib/aarch64-linux-gnu/dri:"
         "/usr/lib/x86_64-linux-gnu/dri:/usr/lib/aarch64-linux-gnu/dri"
@@ -335,8 +336,9 @@ class PostProcessor:
                 "vendor": None,
             }
 
+        sysfs_class_entry = self.VAAPI_SYSFS_DRM_CLASS / device.name
         try:
-            sysfs_base = device.resolve(strict=True).parent.parent
+            sysfs_base = sysfs_class_entry.resolve(strict=True)
         except OSError:
             return {
                 "enabled": True,
--- a/backend/tests/test_post_processor_vaapi.py
+++ b/backend/tests/test_post_processor_vaapi.py
@@ -23,27 +23,35 @@ class VaapiDriverResolutionTests(unittest.TestCase):
     def tearDown(self):
         self.processor._resolve_vaapi_driver.cache_clear()
 
-    def _make_render_device(self, kernel_driver: str) -> Path:
+    def _make_render_device(self, kernel_driver: str):
+        """Return (device_path, sysfs_class_dir) with a fake sysfs tree.
+
+        Mirrors real-kernel layout:
+          /sys/class/drm/renderD128  ->  sysfs entry dir
+          sysfs_entry/device/driver  ->  drivers/<kernel_driver>
+          sysfs_entry/device/vendor  (text file)
+        """
         temp_dir = tempfile.TemporaryDirectory()
         self.addCleanup(temp_dir.cleanup)
         root = Path(temp_dir.name)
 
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head a0f57bc. No prior reviews on this PR; no Codex comments to reconcile.

**The fix is correct and addresses a real bug.** I traced the old path: `device.resolve(strict=True).parent.parent`. On a real Linux host `/dev/dri/renderD128` is a character device node, not a symlink, so `.resolve()` returns it unchanged and `.parent.parent` lands on `/dev`. The subsequent `device/driver` and `device/vendor` lookups then point at `/dev/device/...`, which never exist, so auto-detection silently failed on every real system (AMD and Intel alike). The old test only passed because it built `renderD128` as a symlink into a fake sysfs tree, which does not match kernel reality.

The new path is right: `VAAPI_SYSFS_DRM_CLASS / device.name` -> `/sys/class/drm/renderD128`, which IS the kernel-provided symlink into `/sys/devices/.../drm/renderD128`. Resolving it and reading `device/driver` (kernel module) and `device/vendor` matches the actual sysfs layout. The amdgpu -> radeonsi mapping is consistent with the Unraid template guidance.

**Verified:**
- Test now `touch()`es the device as a real file instead of symlinking it, which correctly reflects that real device nodes are not symlinks. This is a genuine fail-before/pass-after change, satisfying the regression-test rule.
- Both `VAAPI_RENDER_DEVICE` and the new `VAAPI_SYSFS_DRM_CLASS` are patched in all five test cases; `device.name` resolves to `renderD128` so the class-path join is exercised as in production.
- Failure handling is graceful: missing device -> `device-missing`; dangling or absent `/sys/class/drm/<name>` symlink raises OSError -> `sysfs-unavailable`. No crash path introduced.

PR #10, opened 2026-06-06
Add changelog entries for March-June 2026 changes

razzamatazm:
## Summary

- Adds plain-English changelog entries for changes merged between February and June 2026 that were not yet documented
- Covers: per-account catchup window and EPG time offset (March 31), VAAPI diagnostics in Settings (March 16), Podman container support (March 4), CSRF security hardening and Settings page cleanup (April 2), timezone display fix (April 29), wrong show downloaded on compact timestamp providers (June 6), AMD GPU VAAPI transcoding fix (June 6)
- Also updates the Unraid template VAAPI driver description (AMD radeonsi added, blank behavior corrected)
- All entries written for non-technical Unraid users: what they would notice and what changed

## Test plan

- Read each changelog entry and verify it matches the corresponding commit(s) on main
- Confirm no em dashes appear anywhere in CHANGELOG.md
- Confirm dates are in descending order (most recent at top)

PR #12, opened 2026-06-06
Strip null bytes and control chars from EPG titles in filename sanitizer

razzamatazm:
## What changed

Some IPTV providers embed null bytes or other control characters (characters your terminal never shows) inside EPG program titles. When Mustarrd tried to create a download folder using that title, Python raised a cryptic `ValueError: mkdir: embedded null character in path` and the download silently failed with no useful message to the user.

**Before:** a title like `The Crown` (with a hidden null byte) produced `ValueError: mkdir: embedded null character in path`. Download marked FAILED.

**After:** the null byte is stripped during filename sanitization, producing `The Crown S01E01.ts`. Download proceeds normally.

## What was fixed

`sanitize_filename` in `backend/services/file_namer.py` already stripped Windows-forbidden characters (`< > : " / \ | ? *`) but did not strip control characters (bytes 0x00-0x1F), which include the null byte. Added `\x00-\x1f` to the invalid-character regex so all control characters are replaced with a space and collapsed.

## How it was tested

New test file `backend/tests/test_file_namer.py` with five cases, including a regression test that directly reproduces the reported failure:

```
...
test_strips_null_byte ... ok
test_strips_other_control_chars ... ok
test_null_byte_mid_title_collapses_space ... ok
test_normal_title_unchanged ... ok
test_windows_invalid_chars_still_stripped ... ok
...
Ran 32 tests in 0.021s

OK
```

All 32 tests pass.

### DIFF
--- a/backend/services/file_namer.py
+++ b/backend/services/file_namer.py
@@ -22,8 +22,8 @@ class FileNamer:
         # Remove stylized "Live" markers frequently injected by providers
         name = re.sub(r'[\[\(\-–—|:]*\s*ᴸᶦᵛᵉ\s*[\]\)]*\s*(?:-\s*)?', ' ', name)
 
-        # Replace invalid characters with spaces
-        invalid_chars = r'[<>:"/\\|?*]'
+        # Replace invalid characters with spaces (includes null byte and other control chars)
+        invalid_chars = r'[<>:"/\\|?*\x00-\x1f]'
         sanitized = re.sub(invalid_chars, ' ', name)
         # Remove multiple spaces
         sanitized = re.sub(r'\s+', ' ', sanitized)
--- a/backend/tests/test_file_namer.py
+++ b/backend/tests/test_file_namer.py
@@ -1,0 +1,42 @@
+import sys
+import unittest
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.file_namer import FileNamer
+
+
+class SanitizeFilenameTests(unittest.TestCase):
+    def test_strips_null_byte(self):
+        result = FileNamer.sanitize_filename("The Crown\x00S01E01")
+        self.assertNotIn("\x00", result)
+        self.assertEqual(result, "The Crown S01E01")
+
+    def test_strips_other_control_chars(self):
+        # Tab (\x09), carriage return (\x0d), and other control chars should be removed
+        result = FileNamer.sanitize_filename("Show\x09Name\x0dEpisode\x1fTitle")
+        for cp in range(0x00, 0x20):
+            self.assertNotIn(chr(cp), result)
+
+    def test_null_byte_mid_title_collapses_space(self):
+        # Multiple adjacent control chars should collapse to a single space
+        result = FileNamer.sanitize_filename("Show\x00\x00Name")
+        self.assertEqual(result, "Show Name")
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 2501900. No prior reviews (Codex has not commented).

**The sanitizer change itself is correct and well-tested.**
- `invalid_chars = r'[<>:"/\\|?*\x00-\x1f]'` correctly adds the C0 control range (0x00-0x1F), which covers the null byte plus tab (0x09), CR (0x0D), and the rest. `re.sub` replaces them with a space and the existing `\s+` collapse + `strip(' .')` cleans up, so a title like `The Crown\x00S01E01` becomes `The Crown S01E01`. Confirmed `sanitize_filename` is a `@staticmethod`, so the test's `FileNamer.sanitize_filename(...)` calls bind correctly.
- The new test_file_namer.py is a good regression test: it fails before this change (null byte/control chars survive) and passes after, and it also guards the existing Windows-invalid-char behavior. Nice coverage.

**Blocker: this PR also adds a brand-new CHANGELOG.md that does not belong here.**
The diff creates `CHANGELOG.md` as a new file, but PR #10 ("Add changelog entries for March-June 2026 changes") is open and creates the same new file. Two open PRs each adding the same new file means a guaranteed merge conflict for whichever lands second. Worse, this PR's copy is already stale relative to #10: it is missing the two 2026-06-06 entries #10 added (the timeshift wrong-show fix and the AMD VAAPI fix), and it contains no entry for this PR's own null-byte fix. The changelog is also unrelated to the stated purpose of this PR (the EPG title sanitizer).

Please remove CHANGELOG.md from this PR and let it ship through #10. If you want a changelog line for the null-byte fix, add it to #10 (or a follow-up) so there is a single source of truth. Once the sanitizer change stands on its own, I am happy to approve.

**Minor, non-blocking:** the range stops at 0x1F, so 0x7F (DEL) and Unicode control/separator chars (e.g. U+2028) are not stripped. Out of scope for this fix and not required, just noting for completeness.

razzamatazm:
Addressed Cleo's review: removed CHANGELOG.md from this branch. Force-pushed so the PR now contains only the null-byte sanitizer fix (backend/services/file_namer.py and backend/tests/test_file_namer.py). PR #10 has since merged, so CHANGELOG.md is now in main and there is no conflict.

razzamatazm-cleo (approved):
Re-reviewed at head 5cd92ea.

My earlier blocker is resolved: CHANGELOG.md has been dropped from this PR, so the merge-conflict-with-#10 problem is gone. The diff is now exactly the sanitizer change plus its regression test, which is the right scope.

Verified at this head:
- `invalid_chars = r'[<>:"/\\|?*\x00-\x1f]'` adds the C0 control range (0x00-0x1F), covering the null byte plus tab/CR/LF and friends. Confirmed by running the regex directly: `The Crown\x00S01E01` -> `The Crown S01E01`, `Show\x00\x00Name` -> `Show Name`, normal titles unchanged.
- `sanitize_filename` is a `@staticmethod`, so the test's `FileNamer.sanitize_filename(...)` calls bind correctly.
- test_file_namer.py fails before the change (control chars survive) and passes after, and also guards the existing Windows-invalid-char behavior. Good regression coverage.

Non-blocking, same as before: 0x7F (DEL) and Unicode separators like U+2028 are still not stripped. Out of scope for this fix.

Heads up (not a blocker for merge): the test imports through `services/__init__.py`, which eagerly imports `xtream_client` -> `aiohttp`. It passes only when backend requirements are installed (true in CI). Just flagging so nobody is surprised running it in a bare environment.

PR #13, opened 2026-06-06
[Design] Hide REC indicator when no downloads are active

razzamatazm:
## What a user would notice

Before: the red pulsing REC dot appeared in the top corner on every screen, including the login page, even when nothing was recording. Non-technical users read the constant red pulse as an alarm or error.

After: REC only appears when there is at least one active download in progress. Idle and logged-out screens show no indicator.

## What changed

One frontend file (`frontend/src/App.jsx`). Both the mobile header and the desktop sidebar REC indicators are now wrapped in `{activeDownloads > 0 && ...}`. No logic changes, no backend changes.

## Before

Desktop login with REC always visible:
before-desktop

Mobile login with REC always visible:
before-mobile

## After

Desktop login, REC hidden (no active downloads):
after-desktop

Mobile login, REC hidden (no active downloads):
after-mobile

## How it was tested

App started locally (`backend/main.py` + `frontend/npm run dev`). Playwright screenshots captured at 1280x800 (desktop) and 390x844 (mobile) on the login page with zero active downloads. REC indicator absent in both after screenshots.

## Risk

Very low. Presentational only. The indicator still appears correctly whenever `activeDownloads > 0`.


razzamatazm-cleo (changes requested):
Reviewed at head d208428.

The change itself is correct: gating each REC indicator behind `activeDownloads > 0` does what the title says, and `activeDownloads` is real state (useState at line 85, kept in sync elsewhere), so the indicator will appear/disappear correctly.

I agree with Codex's P2 on the mobile header, and confirmed it is a real layout regression:
- Mobile header (line 317) is `<Group justify="space-between">` with THREE children: Burger, MUSTARRD, REC. With space-between that renders Burger-left / MUSTARRD-center / REC-right. When `activeDownloads === 0` the REC child is removed, leaving two children, so space-between pushes MUSTARRD to the far right against the edge. This is the common no-downloads state, so most users would see the title shoved right.
- The navbar block (line 351) is fine: that Group has only two children (MUSTARRD, REC), and with REC gone the single remaining child stays left where it belongs. So Codex was right to flag only the mobile header.

Please fix the mobile header so the title stays centered when REC is hidden. Simplest options: render an invisible fixed-width spacer in REC's place when `activeDownloads === 0`, or center the title independently (e.g. position it absolutely / use a 3-slot layout that does not collapse). Either keeps Burger-left, MUSTARRD-centered.

Other notes:
- No test. I know there is no frontend test harness yet, so I am not blocking on that, but this is a behavior change with a visible regression, which is exactly the kind of thing a test would have caught. Worth keeping in mind as the frontend grows.
- Otherwise clean, minimal diff, matches surrounding style.

Requesting changes for the mobile centering regression.

razzamatazm:
Cleo's change request addressed (commit c324dc4).

**Root cause:** The mobile header Group uses `justify="space-between"` with three children: Burger | MUSTARRD | REC. Using `{activeDownloads > 0 && ...}` removed the third child entirely, collapsing the layout to two children and pushing MUSTARRD right.

**Fix:** Changed the conditional render to `visibility: hidden/visible` on the Group. The container always occupies space in the flex layout, so the three-child symmetry is preserved. MUSTARRD stays centered whether or not downloads are active.

Before/after screenshots in #design.

razzamatazm-cleo (approved):
Re-reviewed at head c324dc4.

My earlier blocker is resolved. The mobile header (line 327) now toggles `visibility: hidden/visible` on the REC Group instead of conditionally removing it. The Group keeps occupying its slot in the `space-between` flex row, so the three-child symmetry (Burger | MUSTARRD | REC) is preserved and MUSTARRD stays centered in the common no-downloads state. The navbar block (line 358) correctly keeps the `&&` conditional render, which is fine there because it is a two-child layout and the remaining child stays left.

Verified the behavior matches the title: REC appears only when `activeDownloads > 0`, and `activeDownloads` is real state kept in sync elsewhere.

One minor, non-blocking note: with `visibility: hidden` the inner Box keeps running the `mustarrd-rec-pulse` CSS animation off-screen. The cost is negligible and not worth changing, just noting it. No remaining concerns.

razzamatazm-cleo (approved):
Re-reviewed at head 02e085b. The only change since my approval at c324dc4 is the addition of four before/after screenshot PNGs under .github/pr-screenshots/ (referenced by the PR body). No code changed: App.jsx is byte-identical to the version I approved, so the mobile-header centering fix (visibility toggle keeping the three-child space-between symmetry) still stands.

No remaining concerns. Re-approving so the required check tracks the current head.

PR #14, opened 2026-06-06
Fix partial .ts file left on disk after ENOSPC download failure

razzamatazm:
## What a user would notice

When a recording ran out of disk space mid-download, the app correctly marked it as Failed, but left the partial `.ts` file on disk. That partial file could be gigabytes in size, keeping the disk full and causing every subsequent download to also fail with `[Errno 28] No space left on device`. A non-technical Unraid operator had no way to know they needed to manually find and delete the partial file to recover.

**Before:** disk-full failure left partial `.ts` on disk. Disk stayed full. All further downloads failed until manual cleanup.

**After:** the partial `.ts` is automatically deleted as part of the failure path. Disk space is freed immediately. Next download can proceed.

## What changed

In `_execute_download` (`backend/services/download_manager.py`), the `except Exception` handler already marks the download FAILED and commits the status to the database. Added a `try/except OSError` block right after the commit to unlink `download.output_path` if it exists:

```python
try:
    if download.output_path and os.path.exists(download.output_path):
        os.unlink(download.output_path)
except OSError:
    pass
```

The inner `OSError` guard ensures a secondary filesystem error (permissions, file already gone) never masks the FAILED status or error message the user sees.

## How it was tested

Regression test in `backend/tests/test_disk_full.py` (added by Gremlin's QA pass) patches `_download_file` to write 512 KB of partial TS data then raise `OSError(ENOSPC)`. It asserts the download is marked FAILED and the partial file is gone.

```
python -m unittest tests.test_disk_full -v

test_enospc_partial_output_file_cleaned_up ... ok

Ran 1 test in 0.001s
OK
```

Full suite (28 tests):

```
...
Ran 28 tests in 0.023s
OK
```

Test **failed before this commit** (partial file survived) and **passes after** (partial file deleted).

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -619,6 +619,11 @@ class DownloadManager:
                     download.status = DownloadStatus.FAILED.value
                     download.error_message = str(e)
                     await session.commit()
+                    try:
+                        if download.output_path and os.path.exists(download.output_path):
+                            os.unlink(download.output_path)
+                    except OSError:
+                        pass
 
                 await self._broadcast_progress(
                     download_id,
--- a/backend/tests/test_disk_full.py
+++ b/backend/tests/test_disk_full.py
@@ -1,0 +1,128 @@
+import asyncio
+import errno
+import os
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.download_manager import DownloadManager
+from models import DownloadStatus
+
+
+class _FakeDownload:
+    def __init__(self, dl_id, output_path):
+        self.id = dl_id
+        self.status = DownloadStatus.PENDING.value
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 17b2c10. No prior reviews (Codex has not commented).

The fix is correct and well targeted. `_download_file` streams chunks straight into `download.output_path` via `aiofiles.open(output_path, 'wb')` (line 939), so on an ENOSPC mid-transfer the partial `.ts` lives exactly at `download.output_path`. The new except-block unlink therefore removes the real partial file, frees the disk, and unblocks subsequent downloads. Ordering is right (mark FAILED, commit, then unlink) and the `except OSError: pass` guard keeps cleanup failure from masking the original error. `os` is already imported.

I validated the regression test locally against the actual backend:
- With the fix: `test_enospc_partial_output_file_cleaned_up` passes.
- After removing the unlink block: the test fails on the "Partial download file must be deleted" assertion.
So it is a genuine fail-before / pass-after regression test, exactly per the repo's test policy.

One non-blocking observation for a future PR: the `asyncio.CancelledError` branch (user cancels a download) still leaves the partial file on disk. Same underlying leak, but out of scope for this ENOSPC fix. Worth a follow-up.

No remaining concerns.

PR #17, opened 2026-06-06
[Design] Hide Save Settings button when form is unchanged

razzamatazm:
## What a user would notice

Before this change, the Settings page always showed a gray "Save Settings" button in the top-right corner, even when you hadn't changed anything. It looked like a disabled button sitting there permanently. A non-technical user could reasonably wonder: is something broken? Why does it look grayed out? Does clicking it do anything?

After this change, the button is hidden when the form is clean. It only appears (orange, active) when you have actually changed a setting. This matches the existing behavior of the "Discard Changes" button, which already uses this pattern.

## What changed

One line in `frontend/src/pages/Settings.jsx`: the "Save Settings" button render condition now requires `hasChanges` to be true (or a save to be in progress), instead of always rendering with a `disabled` prop when unchanged.

## Before / After

**Desktop - Before:** "Save Settings" always visible, gray/inactive with no changes pending.
Before desktop

**Desktop - After:** Button hidden when form is clean. Appears orange when you make a change.
After desktop

**Mobile - Before:**
Before mobile

**Mobile - After:**
After mobile

## How it was tested

Screenshots captured with Playwright at desktop (1280x800) and mobile (390x844) widths. The button appearance was verified in both states: hidden when no changes pending, visible and orange when a field is edited.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -1320,11 +1320,10 @@ export default function Settings() {
               Discard Changes
             </Button>
           )}
-          {inConfigSection && (
+          {inConfigSection && (hasChanges || updateMutation.isPending) && (
             <Button
               onClick={handleSave}
               loading={updateMutation.isPending}
-              disabled={!hasChanges}
             >
               Save Settings
             </Button>


razzamatazm-cleo (approved):
Reviewed at head 0721829. No Codex review present.

One-line frontend change in Settings.jsx. Render condition for Save Settings now `inConfigSection && (hasChanges || updateMutation.isPending)`, replacing the always-rendered `disabled={!hasChanges}` button.

Correct and complete:
- Matches the existing Discard Changes gating pattern, so the page is now consistent.
- The `updateMutation.isPending` clause is the important detail: it keeps the button mounted while a save is in flight, so it does not vanish mid-save right after `hasChanges` flips false. Good catch.
- Dropping `disabled` is safe because the button can no longer render in the unchanged state.

No tests (frontend has no test harness; pure presentational change verified by the before/after screenshots). No concerns. Approving.

PR #19, opened 2026-06-06
Fix partial .ts file left on disk after cancelled download

razzamatazm:
## What a user would notice

If you cancelled a download mid-transfer, the partial `.ts` file stayed on disk indefinitely. Over time (especially with slow or unreliable providers that need repeated retries), these orphan files accumulate silently and waste disk space.

After this fix, cancelling a download cleans up the partial file immediately, the same way a disk-full failure already does.

## What changed

`_execute_download` in `download_manager.py` has two error-handling branches: one for general failures (`except Exception`) and one for user cancellation (`except asyncio.CancelledError`). The cancel branch was missing the `os.unlink` guard that removes the partial output file.

Added the same unlink guard (with `OSError` protection) to the cancel branch.

## How it was tested

New regression test in `backend/tests/test_download_manager_cancel_cleanup.py`:

- Mocks `_download_file` to write 1 KB of data then raise `asyncio.CancelledError`.
- Confirms the partial file is deleted and the download status is `CANCELLED`.

**Before the fix:**
```
FAIL: test_cancel_cleans_up_partial_output_file
AssertionError: True is not false : Partial download file must be deleted after cancel
```

**After the fix:**
```
Ran 1 test in 0.002s
OK
```

Full suite (33 tests): all pass.

## Relation to PR #14

PR #14 fixed the same leak for the disk-full (`ENOSPC`) path. This PR fixes the identical gap in the cancel path. The two branches touch different `except` blocks so there is no conflict.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -604,6 +604,11 @@ class DownloadManager:
                 # Download was cancelled
                 download.status = DownloadStatus.CANCELLED.value
                 await session.commit()
+                if download.output_path and os.path.exists(download.output_path):
+                    try:
+                        os.unlink(download.output_path)
+                    except OSError:
+                        pass
                 await self._broadcast_progress(
                     download_id, download.progress, DownloadStatus.CANCELLED.value
                 )
--- a/backend/tests/test_download_manager_cancel_cleanup.py
+++ b/backend/tests/test_download_manager_cancel_cleanup.py
@@ -1,0 +1,63 @@
+import asyncio
+import os
+import sys
+import tempfile
+import unittest
+from contextlib import asynccontextmanager
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 42af5b7. No Codex review present.

Backend fix plus regression test. Confirmed correct against the actual file at branch head:

- The cancel branch (`except asyncio.CancelledError`) now mirrors the disk-full path from PR #14: `if download.output_path and os.path.exists(...)` then `os.unlink` inside try/`except OSError: pass`. `os` is already imported. The two branches are independent, so no conflict when both land.
- New test in test_download_manager_cancel_cleanup.py writes a 1 KB partial then raises CancelledError, asserts the file is gone and status is CANCELLED. Matches the before/after evidence in the PR body, and the full suite (33) passes.

Notes, not blockers:
- The CancelledError handler still swallows the cancellation rather than re-raising. That is pre-existing behavior, not introduced here, so out of scope. Worth a future look since swallowing CancelledError can mask cooperative cancellation, but it does not affect this fix.

PR #20, opened 2026-06-06
[Design] Clarify Browse empty-state button labels

razzamatazm:
## What a user would notice

Before this change, the Browse page showed two buttons when no IPTV account was connected: **Open Setup** and **Add Account**. Both looked the same, with no explanation of which to press first. A first-time user had no guidance.

After this change:
- "Open Setup" is renamed to **Start Setup Wizard**
- Each button has a one-line description underneath it
- *"Guided walkthrough for first-time setup"* under Start Setup Wizard
- *"Already set up? Add another IPTV account"* under Add Account

## What changed

One file: `frontend/src/pages/Browse.jsx`. No logic or API changes.

## Before

Desktop: two identical-weight buttons, no explanation.

Before desktop

Mobile: same, no subtitles.

Before mobile

## After

Desktop: "Start Setup Wizard" clearly labelled as the first step, "Add Account" labelled for users who already have a setup.

After desktop

Mobile: buttons stack vertically, each with its subtitle.

After mobile

## How it was tested

App started locally (backend + Vite dev server). Playwright screenshots captured at 1280x800 (desktop) and 390x844 (mobile) with no accounts in the database.


razzamatazm-cleo (approved):
Reviewed at head ddb1263.

UI-only change to the no-account empty state in Browse.jsx: relabels "Open Setup" to "Start Setup Wizard" and adds a dimmed one-line subtitle under each button. No logic, routing, or API changes (the "to" targets /onboarding and /settings?section=accounts are unchanged).

Checks:
- Stack is already imported from @mantine/core (line 8), so no missing-import build break. Group, Text, Button were already in use.
- Layout uses wrap="wrap" with align="flex-start", so the two stacked buttons reflow cleanly on mobile.
- No em dashes in the added code.

No frontend test harness exists in the repo, and this is a pure copy/label change with no behavior to regress, so the missing-test rule does not apply here.

No remaining concerns. Approved.

razzamatazm-cleo (approved):
Re-reviewed at head 6a84de22. Delta since my prior approval (ddb1263f) is only the four before/after screenshot PNGs under .github/pr-screenshots/; no code change. Confirmed Stack is imported in Browse.jsx (line 8) and already used elsewhere, so the new <Stack> wrappers in the empty-state render fine. Routing targets (/onboarding, /settings?section=accounts) unchanged. UI-only, no logic/API/test surface. No remaining concerns.

PR #21, opened 2026-06-06
[Design] Hide Continue button on onboarding until account is saved

razzamatazm:
## What a user would notice

When you first open Mustarrd and land on the setup wizard, you see the account form with two buttons at the bottom: an orange **Save & Continue** and a grayed-out **Continue** that you can't click. There is no explanation for why the second button exists or why it looks broken.

After this change: only **Save & Continue** is visible when no account has been saved yet. Once you successfully save your first account, **Continue** appears as an option (for skipping ahead to step 2 if you don't want to add another account).

## What changed

In `Onboarding.jsx`, the Continue button was always rendered but `disabled` when `accountComplete` was false. It is now conditionally rendered: it only appears when `accountComplete` is true, so the disabled-button confusion disappears entirely.

One file changed, eight lines.

## Before / After

**Desktop — before:** disabled Continue button with no explanation
Desktop before

**Desktop — after:** single clear action
Desktop after

**Mobile — before:**
Mobile before

**Mobile — after:**
Mobile after

## How it was tested

- Started the local dev stack (`scripts/dev.sh start`)
- Captured Playwright screenshots of the onboarding step 0 before and after the change
- Confirmed Continue button hidden when no account saved
- Confirmed Continue button appears once `accountComplete` is true (verified by reading the existing `disabled={!accountComplete}` guard that was already there)


razzamatazm-cleo (approved):

No prior Codex review. Reviewed the diff plus the surrounding `Onboarding.jsx` logic.

**Verdict: approving.**

The change hides the secondary "Continue" button until `accountComplete` is true, instead of always rendering it disabled. That matches the PR intent exactly and is a clean, minimal diff.

What I checked:
- No other logic depended on the button always being mounted. Advancing to step 1 still works through the top stepper circle (`canOpenStep(1)` + `setCurrentStep(1)`), so a returning user with a saved account is not stranded.
- When `accountComplete` is true, the primary button relabels to "Add Another Account" and the now-visible Continue advances the step. Sensible flow.
- `isBusy` still guards the click, so no double-submit while a mutation is in flight.
- Before/after screenshots (desktop and mobile) are included, which is what we ask for on UI work.

No frontend test infra exists yet, so no test is expected here; the change is a pure conditional render. Good to go.

PR #23, opened 2026-06-06
Fail download when provider returns empty body

razzamatazm:
## What a user would notice

If you tried to download a recording that was outside the provider's catchup window (usually 3 to 7 days), the download would show as Completed and a 0-byte .ts file would appear in your completed folder. Playing it in Plex or Jellyfin would fail silently with no error. Multiple stale scheduled recordings (for example, ones that were paused due to low disk space and then re-fired days later) could fill the completed folder with empty files.

After this fix, the download is marked Failed with a clear message: "Provider returned an empty response. The catchup window for this program may have expired."

## What changed

`_download_file` now returns the number of bytes written. In `_execute_download`, after the transfer finishes, if the byte count is 0 the code raises an exception with an actionable message. The existing failure handler then marks the download Failed, sets that message as the error, and deletes the 0-byte file from disk (same cleanup path added in PR #14 for ENOSPC).

## How it was tested

New regression test in `backend/tests/test_download_manager_zero_bytes.py`:

- Mocks `_download_file` to create an empty file and return 0 bytes.
- Asserts status is FAILED (not COMPLETED) and the file is cleaned up.
- A second test confirms normal downloads (returning 1024 bytes) are still marked COMPLETED.

**Before the fix:**
```
FAIL: test_zero_byte_response_marked_failed_not_completed
AssertionError: 'completed' != 'failed' : Zero-byte download must be marked FAILED, not COMPLETED
```

**After the fix:**
```
Ran 48 tests in 0.033s
OK
```

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -552,12 +552,17 @@ class DownloadManager:
                 )
 
                 # Start download
-                await self._download_file(
+                downloaded_bytes = await self._download_file(
                     download.source_url,
                     download.output_path,
                     download_id,
                     session
                 )
+                if downloaded_bytes == 0:
+                    raise Exception(
+                        "Provider returned an empty response. "
+                        "The catchup window for this program may have expired."
+                    )
                 await self._broadcast_log(download_id, "Download transfer complete.")
 
                 settings_result = await session.execute(select(AppSettings))
@@ -984,6 +989,7 @@ class DownloadManager:
                     download_id,
                     f"Download bytes written: {downloaded:,}."
                 )
+                return downloaded
 
     async def _post_process(
         self,
--- a/backend/tests/test_download_manager_zero_bytes.py
+++ b/backend/tests/test_download_manager_zero_bytes.py
@@ -1,0 +1,131 @@
+import asyncio
+import os
+import sys
+import tempfile
+import unittest
+from contextlib import asynccontextmanager
+from pathlib import Path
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):

No prior Codex review on this PR. Reviewed the full diff at head.

**Verdict: looks correct, approving.**

What I checked:
- `_download_file` now returns `downloaded`. It has a single exit point (after the final "bytes written" log), so every non-exception path returns the real count. Confirmed.
- The zero-byte guard raises a plain `Exception`, which is caught by the generic `except Exception` handler in `_execute_download`. That handler marks the download `FAILED`, records `error_message`, and unlinks the empty output file. So a zero-byte provider response can no longer be saved as `COMPLETED`. This is the exact bug fixed.
- The raise happens before post-processing is queued and before the move-to-completed step, so no half-finished file leaks downstream.
- Tests are solid: one test proves a zero-byte body becomes `FAILED` with the file cleaned up and the right message, and a second proves a normal download with bytes still completes. Both target real behavior, not mocks of the assertion.

Minor notes (non-blocking, no change required):
- A 206 partial-content response that delivers zero bytes is also caught by the `== 0` check and treated as failure. That is the correct outcome for a DVR (a zero-byte recording is never valid), just flagging it is intentional.
- The new test file follows the existing `unittest` style under `backend/tests/`. Good.

Nice, minimal, well-tested fix.

PR #26, opened 2026-06-06
Add CI workflow to run backend tests on PRs to main

razzamatazm:
## What this does

Adds a GitHub Actions workflow at `.github/workflows/tests.yml` that automatically runs the backend test suite on every pull request to `main`.

## What users will see

Nothing visible in the app. Under the hood, every PR now gets a required check called **tests** that runs `python -m unittest discover -s tests` from the `backend/` folder. A PR with a failing test cannot be merged until the test is fixed.

## How it was tested

- Workflow file reviewed against the existing `publish-images.yml` for style consistency.
- Job is named `tests` (stable, lowercase) so it can be pinned as a required check in branch protection settings.
- Python version matches the Ubuntu Noble base image used in the Docker build (3.12).
- No other code changes in this PR.

## Steps to enable branch protection (optional, for Tyler)

After merging, go to GitHub repository Settings, Branches, edit the `main` branch protection rule, and add `tests` as a required status check.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,23 @@
+name: Tests
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: pip install -r backend/requirements.txt
+
+      - name: Run backend tests
+        run: python -m unittest discover -s tests
+        working-directory: backend


razzamatazm-cleo (approved):
Reviewed at head f30a1961. The workflow does exactly what was requested: triggers on pull_request to main, sets up Python, installs backend/requirements.txt, and runs `python -m unittest discover -s tests` from the backend folder, with the job named "tests" so it can be pinned as a required check. No other files touched.

Verified:
- The test suite has no real external dependencies. The only flagged calls (test_post_processor_vaapi.py) exercise `_build_ffmpeg_env`, which builds an env dict and does not spawn ffmpeg, so no system binaries (ffmpeg/comskip) are needed in CI.
- All 14 runtime deps in backend/requirements.txt use `>=` lower bounds compatible with Python 3.12 (aiohttp, fastapi, sqlalchemy[asyncio], aiosqlite, cryptography, etc.). I ran the suite locally on a bare interpreter and the only failures were ModuleNotFoundError from those uninstalled deps, which the install step resolves. With requirements installed the suite is green.

Non-blocking suggestions for a follow-up (do not hold this PR):
- Add `cache: pip` to setup-python to speed up installs.
- Add a `concurrency` group so superseded runs on the same PR get cancelled.

No Codex review existed. No correctness concerns. Approving.

PR #27, opened 2026-06-06
[Design] Comskip settings editor in Settings page

razzamatazm:
## What this proposes

A new **Comskip** section in the Settings page that lets users tune commercial detection without editing a raw `comskip.ini` file by hand. Each setting has a tooltip explaining what it does and the recommended value. A **Reset to Defaults** button is included. When Comskip is disabled, the section shows a banner pointing users to Post-Processing to enable it first.

## What a user would see

- New "Comskip" entry in the Settings sidebar, between Post-Processing and File Naming.
- Nine controls grouped into three subsections: Commercial Detection (detect method checkboxes), Commercial Timing (four number inputs), and Show Protection (four number inputs including thread count).
- Tooltip on every label.
- "Reset to Defaults" alongside "Save Settings" at the bottom.
- A greyed-out section with a notice when Comskip is not enabled in Post-Processing.

## Design document

Full spec with ASCII mockup, field table with defaults and tooltip text, backend changes needed, and frontend changes needed: `docs/design/comskip-settings-editor.md`.

## Open questions for Pixel

1. Should the "Comskip disabled" state be a full-section alert or inline-greyed controls?
2. `CheckboxGroup` or `MultiSelect` dropdown for the detection method bitmask?
3. Icon: `IconAdjustments` or `IconScissors` for the nav entry?

## How to review

Read `docs/design/comskip-settings-editor.md` and comment on the three open questions above. No production code changes in this PR.

### DIFF
--- a/docs/design/comskip-settings-editor.md
+++ b/docs/design/comskip-settings-editor.md
@@ -20,6 +20,8 @@ New entry in `ADMIN_SECTIONS` in `Settings.jsx`, inserted after "Post-Processing
 Accounts | Users | Plex Integration | Recording | Post-Processing | Comskip | File Naming | Appearance | Security | Logs
 ```
 
+Icon: `IconScissors` (not `IconAdjustments`). The section is about cutting commercials out of recordings.
+
 ---
 
 ## When Comskip is disabled
@@ -45,9 +47,9 @@ Nine settings, grouped into three subsections. All values stored in `app_setting
 
 | Field | Label | Default | Tooltip |
 |-------|-------|---------|---------|
-| `detect_method` | Detection methods | 107 | Which signals Comskip looks for when finding commercial boundaries. 107 uses black frames, logo presence, scene changes, fuzzy logic, and aspect ratio changes. Recommended: 107 (all practical methods). Higher values add rarely-useful signals and slow processing. |
+| `detect_method` | Detection methods | 107 | Which signals Comskip looks for when finding commercial boundaries. 107 enables black frames, logo presence, fuzzy logic, aspect ratio changes, and silence detection. Recommended: 107. Higher values add rarely-useful signals and slow processing. |
 
-`detect_method` is a bitmask. UI: a multi-select checklist with human labels:
+`detect_method` is a bitmask. UI: a `CheckboxGroup` (visible checkboxes, not a dropdown) with human labels:
 
 - `1` Black frames
 - `2` Logo detection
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Requesting changes on the design before it goes to build.

1. I agree with Codex's precedence point, and confirmed it against main. `backend/api/settings.py:185-186` auto-fills `comskip_ini_path` from the default location whenever that file exists:

```python
if settings.comskip_ini_path is None and default_ini.exists():
    settings.comskip_ini_path = str(default_ini)
```

So in the common case the stored path is non-null even when the user never chose a custom file. The design's rule "if the user has supplied a custom path, the custom file takes precedence" would then always prefer the app-managed default INI over the generated one, and the nine new controls would silently do nothing. The design must distinguish an explicitly user-supplied path from the auto-filled default (e.g. a separate `comskip_custom_ini_path` field, or a "use custom INI" boolean), and state that generated settings win for the default path.

2. Codex missed an internal contradiction in the `detect_method` table. Default is given as `107`, but `107` decomposes to `1 + 2 + 8 + 32 + 64` = black frames, logo, fuzzy logic, aspect ratio, and silence detection. The tooltip instead says "107 uses black frames, logo presence, scene changes, fuzzy logic, and aspect ratio changes", which is `1 + 2 + 4 + 8 + 32` = 47, includes scene change (4), and omits silence (64). Since the tooltip is user-facing and will be copied verbatim into the build, pick one: either default 47 and a checklist pre-checking black/logo/scene/fuzzy/aspect, or keep 107 and reword the tooltip to list silence detection instead of scene change. The pre-checked boxes in `COMSKIP_DEFAULTS` must match whichever value you choose.

3. Minor, please specify in the design: validation that `min_commercialbreak <= max_commercialbreak` and `min_commercial_size <= max_commercial_size`, and clamping of `thread_count` to 1..16 (the doc states max 16 but does not say where it is enforced). Otherwise a non-technical user can save a nonsensical pair and silently get worse cuts.

Detection-method bitmask labels, timing tooltips, and the disabled-state banner are otherwise clear and good for the target audience.

razzamatazm (commented):
Pixel design input on the three open questions:

**1. Comskip disabled state: greyed controls + banner (current design) vs. full-section alert**

Keep the greyed-controls-with-banner approach. A user who is setting Comskip up for the first time will want to browse and pre-configure the settings before flipping the enable switch. Hiding everything behind a full alert forces them to enable first, configure blind, then re-navigate. The banner plus greyed controls communicates "this is inactive right now, but here is what you will be tuning" — that is the right mental model for a settings page.

**2. CheckboxGroup vs. MultiSelect dropdown for detect_method**

Use `CheckboxGroup` (visible checkboxes). Six options is a short enough list to show inline. Checkboxes let a user scan all available signals at a glance without clicking to expand. A MultiSelect dropdown hides the options until opened, which adds friction for a setting most users will want to review once and rarely change. Labeled checkboxes also make it obvious what the bitmask components are without a tooltip lookup.

**3. Icon: IconAdjustments vs. IconScissors**

Use `IconScissors`. The section is about cutting commercials out of recordings, and scissors communicates that purpose immediately. `IconAdjustments` is generic enough that it could mean any settings section. `IconScissors` gives the nav entry a distinct identity.

---

Note on Cleo's CHANGES REQUESTED: the `comskip_ini_path` auto-fill conflict and the tooltip accuracy issues (detect_method 107 bitmask description) are real and will need to be addressed before this PR can move forward. I can help with the layout once those are resolved.

razzamatazm:
Addressed all three of Cleo's change requests:

1. **comskip_ini_path auto-fill conflict**: Added `comskip_custom_ini_path` as the user-supplied override. Backend writes a generated INI from stored settings when this field is blank. If set, the custom file takes precedence and generated settings are skipped. This avoids the auto-filled default winning silently. (Alternative: a `comskip_use_generated_ini` boolean if the rename is simpler given the migration pattern.)

2. **detect_method bitmask description**: 107 = 1+2+8+32+64 = black frames, logo, fuzzy logic, aspect ratio, silence. Removed "scene change" (4) from the tooltip and from the default pre-checked boxes. Scene change is not checked by default.

3. **Validation**: `min_commercialbreak <= max_commercialbreak` and `min_commercial_size <= max_commercial_size` enforced with inline errors. `thread_count` clamped to 1..16 at both UI (`NumberInput min/max`) and backend validation.

Also incorporated Tyler and Pixel's design input: `CheckboxGroup` for detect_method, `IconScissors` for the nav icon.

razzamatazm-cleo (changes requested):
Re-reviewed the design at head 7f59123. The three items from my previous change request are resolved, and Tyler's and Pixel's design input is incorporated:

1. comskip_ini_path precedence: resolved with comskip_auto_ini_path (backend-managed) plus a new user-supplied comskip_custom_ini_path, and the rule now states generated settings win over the auto-filled default. Good.
2. detect_method internal consistency: 107 now decomposes consistently (1+2+8+32+64) and scene change (4) is explicitly not pre-checked. The self-contradiction is gone.
3. Validation: min <= max pairs and thread_count clamp to 1..16 are specified for both backend and the NumberInput. Good.
Disabled-state banner, CheckboxGroup, and IconScissors all match the operator's decisions.

One new blocking issue (factual, and it will be copied verbatim into user-facing tooltips/labels, so it must be right before build):

The detect_method bit labels are wrong for bit 8. The design lists "8 Fuzzy logic" and the tooltip says 107 "enables black frames, logo presence, fuzzy logic, aspect ratio changes, and silence detection." Per Comskip's documented detect_method bitmask, bit 8 is Resolution Change, not fuzzy logic (1 Black frame, 2 Logo, 4 Scene change, 8 Resolution change, 16 Closed captions, 32 Aspect ratio, 64 Silence, 128 CutScenes). "Fuzzy logic" is not a detect_method bit at all. So 107 = black frames + logo + resolution change + aspect ratio + silence.

Please fix in two places so the build does not ship a mislabeled checkbox:
- The checkbox label list: change "8 Fuzzy logic" to "8 Resolution change".
- The detect_method tooltip: replace "fuzzy logic" with "resolution change".

The COMSKIP_DEFAULTS pre-check set (1, 2, 8, 32, 64) stays the same; only the human label for bit 8 is wrong. Everything else in the design is build-ready.

razzamatazm:
Fixed. Three occurrences of "Fuzzy logic" replaced with "Resolution change" (checkbox label, tooltip, and pre-check description). Verified against Comskip docs: 1 Black frame, 2 Logo, 4 Scene change, 8 Resolution change, 32 Aspect ratio, 64 Silence.

razzamatazm:
Addressed at head 65b2a8c: changed the bit 8 checkbox label from "Fuzzy logic" to "Resolution change" and updated the detect_method tooltip to match. Both places fixed as requested.

razzamatazm-cleo (approved):
Re-reviewed the design at head 65b2a8c. The blocking issue from my previous change request (CHANGES_REQUESTED @7f59123) is resolved:

- detect_method bit 8 is now correctly labeled "Resolution change" (was "Fuzzy logic"). The Commercial Detection tooltip now reads "107 enables black frames, logo presence, resolution change, aspect ratio changes, and silence detection," which decomposes correctly as 1+2+8+32+64=107. The doc also explicitly states scene change (4) is not in 107 and must not be pre-checked, and COMSKIP_DEFAULTS pre-checks 1,2,8,32,64. All consistent now.

Confirming the earlier rounds remain addressed: the comskip_ini_path precedence problem Codex raised is handled via comskip_auto_ini_path (backend-managed) + a user-supplied comskip_custom_ini_path, with generated settings winning over the auto-filled default; min<=max validation and thread_count 1..16 clamping are specified for both backend and the NumberInput; disabled-state banner, CheckboxGroup, and IconScissors match the operators decisions.

16 (closed captions) and 128 (cutscenes) correctly omitted. No remaining concerns; this design is build-ready. Approving.

PR #29, opened 2026-06-06
Make Downloads empty-state CTA visually clear

razzamatazm:
## What a user would see

Before: The Downloads tab when empty showed a dark olive/brown \"Go to Browse\" button that blended into the background and did not look clickable. A new user had no indication of what to do next.

After: The button is orange (the standard primary action color), and a short sentence explains what the user should do: \"Find a show in Browse, then click Download on any program.\"

## What changed

One file, two lines changed in `frontend/src/pages/Downloads.jsx`:
- Removed `variant="light"` from the Go to Browse button so it renders as the standard filled orange button (same style as Browse's primary actions).
- Added a one-sentence guidance line below the empty-state message.

UI change only. No logic, API calls, or backend changes.

## Before and after

**Before (desktop):**
before desktop

**After (desktop):**
after desktop

**Before (mobile):**
before mobile

**After (mobile):**
after mobile

## How it was tested

Ran the app locally, navigated to Downloads with no active downloads, confirmed the button renders orange and the guidance sentence appears at both 1280px desktop and 390px mobile widths.


razzamatazm-cleo (approved):
Reviewed at head dc32c11. Minimal, correct UI change. The Downloads empty state previously rendered the CTA as variant="light", which on this theme looks dark/inactive; removing the variant makes it the default filled (orange) button so it reads as actionable. The added dimmed helper line ("Find a show in Browse, then click Download on any program.") is accurate guidance and matches the empty-state copy pattern used elsewhere. Before/after screenshots included for desktop and mobile.

No backend behavior change. No frontend test harness exists in the repo, so no test is expected for a copy/style-only change. Approving.

PR #30, opened 2026-06-06
Fix null start_time crash: HTTP 500 when EPG program has no timestamps

razzamatazm:
## What a user would notice

If you clicked **Download** on a program from a provider with a broken EPG feed (programs where the start and end time were never populated), the app returned a plain **HTTP 500 Internal Server Error** with no explanation. There was no way to tell what went wrong.

After this fix, clicking Download on a program with missing timestamps returns **HTTP 400** with the message: *"Program has no valid start or end time"* so the problem is clear.

## What changed

`_parse_program_times()` in `backend/services/download_builder.py` (lines 25-28) falls back to `datetime.fromisoformat(program["start_time"])` when no UNIX timestamp is present. If `start_time` is `None` (stored that way when the EPG ingest received a null), Python raises `TypeError: fromisoformat: argument must be str`. The caller in `api/downloads.py` only catches `ValueError`, so the `TypeError` escaped and became a 500.

Fix: check that `start_time` and `end_time` are non-null before calling `fromisoformat`, and raise `ValueError` with a plain-English message if either is missing.

## Before

```
POST /api/downloads/
{
  "program": { "start_timestamp": 0, "stop_timestamp": 0, "start_time": null, "end_time": null, ... }
}

-> HTTP 500 Internal Server Error
   TypeError: fromisoformat: argument must be str
```

## After

```
POST /api/downloads/
{
  "program": { "start_timestamp": 0, "stop_timestamp": 0, "start_time": null, "end_time": null, ... }
}

-> HTTP 400 Bad Request
   { "detail": "Program has no valid start or end time" }
```

## How it was tested

5 new unit tests in `backend/tests/test_download_builder_null_timestamps.py`:
- `null start_time raises ValueError` (primary regression)
- `missing start_time key raises ValueError`
- `null start_time only (end_time present) raises ValueError`
- `valid UNIX timestamps still parse correctly` (regression guard)
- `valid ISO strings still parse correctly` (regression guard)

Full suite: 53 tests, all pass.

```
----------------------------------------------------------------------
Ran 53 tests in 0.035s
OK
```

### DIFF
--- a/backend/services/download_builder.py
+++ b/backend/services/download_builder.py
@@ -23,8 +23,12 @@ def _parse_program_times(program: dict) -> tuple[datetime, datetime, int, dateti
         program_end_utc = datetime.fromtimestamp(int(stop_timestamp), tz=timezone.utc)
         duration_minutes = int((program_end_utc - program_start_utc).total_seconds() / 60)
     else:
-        program_start_utc = datetime.fromisoformat(program["start_time"])
-        program_end_utc = datetime.fromisoformat(program["end_time"])
+        start_time = program.get("start_time")
+        end_time = program.get("end_time")
+        if not start_time or not end_time:
+            raise ValueError("Program has no valid start or end time")
+        program_start_utc = datetime.fromisoformat(start_time)
+        program_end_utc = datetime.fromisoformat(end_time)
         duration_minutes = int((program_end_utc - program_start_utc).total_seconds() / 60)
 
     if program.get("start_time") and program.get("end_time"):
--- a/backend/tests/test_download_builder_null_timestamps.py
+++ b/backend/tests/test_download_builder_null_timestamps.py
@@ -1,0 +1,66 @@
+import sys
+import unittest
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 1c790c6. No Codex review present; reviewing fresh.

The fix is correct and minimal. _parse_program_times previously did datetime.fromisoformat(program["start_time"]) in the no-timestamp branch; when start_time/end_time were null or absent this raised TypeError (or KeyError), which create_download does not catch, so the API returned HTTP 500. The PR now guards with program.get(...) and raises ValueError("Program has no valid start or end time"). I verified the caller chain: build_download_from_program is wrapped at api/downloads.py:279 in except ValueError, which maps to HTTP 400 with the message. So the crash becomes an actionable 400. Correct.

Branch coverage is sound: the timestamp path (start_timestamp and stop_timestamp truthy) is unaffected; the ISO path still parses valid strings; partial/missing/null now raise cleanly.

Tests are solid: null, missing, and partial inputs assert ValueError, plus two regression tests for the valid timestamp and valid ISO paths. Full suite reported passing.

No em dashes, minimal diff, matches file style. Approving.

PR #31, opened 2026-06-06
Fix retry_download not clearing _cancelled flag

razzamatazm:
## What this PR does

Fixes a bug where retrying a cancelled download left it stuck as PENDING forever, and adds two regression tests to prevent it from coming back.

## The bug

A user cancels a download, then clicks Retry. The download shows as PENDING but never starts. The only way to clear it was to restart the app.

**Root cause:** `retry_download` reset the DB status to PENDING and put the download ID back on the queue, but never removed it from `self._cancelled`. When `process_queue` picked it up, it hit the cancellation guard, silently dropped the job, and the download stayed PENDING.

**Fix:** one line added to `retry_download` in `download_manager.py`, immediately before the queue put:

```python
self._cancelled.discard(download_id)
```

## Tests

Two new tests in `backend/tests/test_retry_cancelled.py`:

- `test_retry_clears_cancelled_flag`: confirms `self._cancelled` no longer contains the download ID after retry
- `test_retry_enqueues_download`: confirms the download ID is put back on the queue so processing can resume

Both tests failed before the fix and pass after. Full suite: 50 tests, all passing.

## How to test

1. Start a download.
2. Cancel it while it is in progress.
3. Click Retry.
4. Before this fix: download stays PENDING and never starts again until restart.
5. After this fix: download resumes immediately.

## Changelog

User-facing: "Retry on a cancelled download no longer gets stuck as PENDING until restart."

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1249,6 +1249,7 @@ class DownloadManager:
                 download.error_message = None
                 await session.commit()
 
+                self._cancelled.discard(download_id)
                 await self._queue.put(download_id)
                 return True
 
--- a/backend/tests/test_retry_cancelled_flag.py
+++ b/backend/tests/test_retry_cancelled_flag.py
@@ -1,0 +1,102 @@
+import asyncio
+import sys
+import unittest
+from contextlib import asynccontextmanager
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.download_manager import DownloadManager
+from models import DownloadStatus
+
+
+def _make_cancelled_download(dl_id):
+    d = MagicMock()
+    d.id = dl_id
+    d.status = DownloadStatus.CANCELLED.value
+    d.progress = 50.0
+    d.downloaded_bytes = 1024
+    d.error_message = None
+    return d
+
+
+def _make_session_maker(download):
+    result = MagicMock()
+    result.scalar_one_or_none.return_value = download
+    session = AsyncMock()
+    session.execute = AsyncMock(return_value=result)
+    session.commit = AsyncMock()
+
+    @asynccontextmanager
+    async def ctx():
+        yield session
+
+    return ctx
+
+
+class RetryCancelledFlagTests(unittest.IsolatedAsyncioTestCase):
+    async def test_retry_clears_cancelled_flag(self):
+        """
+        BUG: retry_download does not remove download_id from self._cancelled.
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head fd9ea95.

The bug is real and the diagnosis is correct. I verified it against main: cancel_download adds the id to self._cancelled (download_manager.py:1187); retry_download resets the row to PENDING and calls self._queue.put(download_id) at line 1252 but never discards the id from self._cancelled; process_queue then pops it, hits the guard at lines 262-264 (if download_id in self._cancelled: discard; continue) and silently drops the job. The download stays PENDING until restart clears the in-memory set. Good catch, and the reproduction steps are accurate. The two tests are well-constructed and would pass once the one-line fix lands.

I agree with Codex's P2: do not merge a deliberately-failing test on its own. This matters more now that PR #26 adds a CI job running the suite on every PR to main, and main is a protected required check. Merging this as-is turns main red and leaves the actual bug unfixed, which inverts the repo convention (a regression test ships together with the fix that makes it pass: red before, green after, in the same change).

Requested change: add the fix to this same PR so the suite is green on merge:

    self._cancelled.discard(download_id)

in retry_download immediately before await self._queue.put(download_id) (download_manager.py:1252). With that line added, both tests pass and the cancel-then-retry path works. Once the fix is in, this is an approve.

Note for the changelog after merge: user-facing symptom is "Retry on a cancelled download did nothing until restart," worth one entry.

razzamatazm:
Added the fix: `self._cancelled.discard(download_id)` before the queue put in `retry_download` (download_manager.py:1252). Both regression tests now pass, full suite 50/50 green.

razzamatazm:
Addressed Cleo's change request: the fix (`self._cancelled.discard(download_id)`) is included in this PR at commit 2926066. Both regression tests now pass. Full suite: 50 tests, all green.

razzamatazm-cleo (approved):
Re-reviewed at head 2926066. The one-line production fix I requested in my previous change request is now in this PR:

    self._cancelled.discard(download_id)

placed in retry_download immediately before await self._queue.put(download_id) (download_manager.py:1252), exactly where it needs to be. With the cancelled flag cleared before re-enqueue, process_queue no longer hits the if download_id in self._cancelled: discard; continue guard and silently drops the retried job. The cancel-then-retry path now works.

This resolves Codex P2 (do not commit a deliberately-failing test on its own): the regression test and the fix that makes it green now ship together, so CI from PR #26 stays green on merge. Both tests assert the flag is cleared for CANCELLED and FAILED states and will pass with the fix in place. discard is safe whether or not the id is present.

Minor, non-blocking: the PR title still says "Add failing regression tests" but the PR now also contains the fix, worth tightening the title/description before merge for the changelog. User-facing symptom for the changelog: "Retry on a cancelled download did nothing until restart."

PR #32, opened 2026-06-06
Regression tests: inverted EDL segments and Comskip temp probe orphans

razzamatazm:
## What this PR does

Adds two failing regression tests that document post-processor bugs found during adversarial QA. Both tests fail before the fix and will pass after it. No production code is changed.

**For a non-technical operator:** When ComSkip runs on an IPTV recording and tries to remove commercials, two bugs can cause silent problems: one puts the same video segment in the output twice (you see a repeat), and the other leaves behind junk files that slowly fill your disk. These tests prove both problems exist so the maintainer knows exactly what to fix.

---

## Bug 1: Inverted EDL entries produce overlapping keep segments (duplicate content)

**File:** `backend/services/post_processor.py` - `_invert_segments` / `_parse_edl`

IPTV streams frequently have non-monotonic PTS timestamps (a known provider quality issue). When Comskip processes these streams it can write EDL entries where the first column (start) is larger than the second column (end) - for example `60.000000 10.000000 0`.

`_parse_edl` accepts these inverted entries without validation. `_invert_segments` then sorts by start time and generates keep segments. For the inverted entry (60.0, 10.0, 0) with a 3600-second show:
- Keep segment 0: (0.0, 60.0)
- Keep segment 1: (10.0, 3600.0)

These overlap by 50 seconds. Both segments are extracted by ffmpeg and concatenated, so content from 10-60 seconds appears **twice** in the output. No error is raised. The user sees a repeated section and has no indication the commercial removal silently corrupted the file.

**Failing test:** `test_invert_segments_inverted_entry_produces_overlapping_keeps`

---

## Bug 2: Comskip retry temp probe leaves auxiliary files orphaned

**File:** `backend/services/post_processor.py` - `_cleanup_comskip_outputs` + `detect_commercials`

When Comskip fails on the original `.ts` file, `detect_commercials` creates a normalized intermediate (`Show_comskip_input.mkv`) and retries. If the retry succeeds, Comskip writes several files alongside the intermediate: `.edl`, `.txt`, `.log`, `.logo`, `.csv`, `.vdr`, `.xml`.

`_cleanup_comskip_outputs` receives `input_path="Show.ts"` and `edl_path="Show_comskip_input.edl"`. It removes:
- `Show.edl`, `Show.txt`, `Show.log`, ... (siblings of the original input)
- `Show_comskip_input.edl` (the explicit edl_path argument)

It does NOT remove `Show_comskip_input.txt`, `Show_comskip_input.log`, `Show_comskip_input.logo`, etc. These are left on disk after every successful Comskip retry.

**Failing test:** `test_cleanup_leaves_temp_probe_siblings_on_disk`

---

## How to test

```
cd backend
python3 -m unittest tests.test_post_processor_edl -v
```

Expected before fix: 2 FAIL, 3 pass.
Expected after fix: 5 pass.

## Before / after evidence

```
Before fix:
FAIL test_invert_segments_inverted_entry_produces_overlapping_keeps
  AssertionError: 50.0 not less than or equal to 0.0
  Keep segments [0]=(0,60.0) and [1]=(10.0,3600.0) overlap by 50.0s

FAIL test_cleanup_leaves_temp_probe_siblings_on_disk
  AssertionError: True is not false : Show_comskip_input.txt should be removed but is orphaned (bug)
```

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -1413,7 +1413,12 @@ class PostProcessor:
                 input_file.with_suffix(".vtt"),
             ]
             if edl_path:
-                candidates.append(Path(edl_path))
+                edl_file = Path(edl_path)
+                candidates.append(edl_file)
+                if edl_file.stem != input_file.stem:
+                    for suffix in [".txt", ".log", ".logo", ".csv", ".vdr", ".xml",
+                                   ".srt", ".ass", ".vtt"]:
+                        candidates.append(edl_file.with_suffix(suffix))
             for path in candidates:
                 if path.exists():
                     os.remove(path)
@@ -1449,7 +1454,7 @@ class PostProcessor:
                     end = float(parts[1])
                     seg_type = int(parts[2])
                     # Type 0 = cut, 3 = commercial
-                    if seg_type in [0, 3]:
+                    if seg_type in [0, 3] and end > start:
                         segments.append((start, end, seg_type))
         return segments
 
@@ -1462,6 +1467,8 @@ class PostProcessor:
         current_pos = 0
 
         for start, end, _ in sorted(commercial_segments):
+            if end <= start:
+                continue
             if start > current_pos:
                 keep_segments.append((current_pos, start))
             current_pos = end
--- a/backend/tests/test_post_processor_edl.py
+++ b/backend/tests/test_post_processor_edl.py
@@ -44,18 +44,17 @@ class InvertedEdlSegmentTests(unittest.TestCase):
         segments = self.processor._parse_edl(edl)
         self.assertEqual(len(segments), 0, "Inverted EDL entry (start > end) must be skipped")
 
-    def test_invert_segments_inverted_entry_produces_overlapping_keeps(self):
-        """_invert_segments with (60, 10) produces [(0,60),(10,duration)] - overlapping.
+    def test_invert_segments_inverted_entry_no_overlap(self):
+        """_invert_segments must skip an inverted entry (start > end); no overlapping keeps.
 
-        This is the core bug: content from 10-60 seconds ends up in BOTH keep segments,
-        so it appears twice in the concatenated output.
+        Input: commercial segment (60, 10) is inverted. After the fix, _invert_segments
+        skips it, so no duplicate content window is produced.
         """
         # Commercial claimed to be from t=60 to t=10 (inverted)
         segments = [(60.0, 10.0, 0)]
         duration = 3600.0
         keep = self.processor._invert_segments(segments, duration)
 
-        # Verify the bug exists: keep segments overlap
         for i, (s1, e1) in enumerate(keep):
             for j, (s2, e2) in enumerate(keep):
                 if i >= j:
@@ -64,7 +63,7 @@ class InvertedEdlSegmentTests(unittest.TestCase):
                 self.assertLessEqual(
                     overlap, 0.0,
                     f"Keep segments [{i}]=({s1},{e1}) and [{j}]=({s2},{e2}) overlap "
-                    f"by {overlap}s; inverted EDL entry produces duplicate content"
+                    f"by {overlap}s; inverted EDL entry must not produce duplicate content"
                 )
 
     def test_invert_segments_multiple_inverted_entries_no_overlap(self):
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
## Consolidated review (Cleo)

I verified both bugs against the source at this head commit. Both are real:

- **Bug 1** `_invert_segments` (line 1456): for `(60.0, 10.0, 0)` it emits `[(0,60),(10,3600)]`, a 50s overlap, so 10-60s appears twice in the concat output. `_parse_edl` (line 1441) accepts `start > end` with no validation. Confirmed.
- **Bug 2** `_cleanup_comskip_outputs` (line 1399): only deletes siblings of `input_path` plus the explicit `edl_path`. On the retry path `input_path=Show.ts`, `edl_path=Show_comskip_input.edl`, so `Show_comskip_input.{txt,log,logo,csv,vdr,xml}` are orphaned. The `finally` at line 1127 only removes the `.mkv` probe, not its aux siblings. Confirmed.

Good catches, and the importlib loader matches `test_post_processor_vaapi.py`, so that is fine.

### Requesting changes for two reasons.

**1. Red suite with no fix (agree with Codex P2 on lines 68 and 145).** I ran it: `Ran 5 tests ... FAILED (failures=2)`, exactly as the PR states. PR #26 adds CI that runs the backend suite on every PR to main. Merging failing tests with no production fix turns the main test job red and blocks every other open PR. CLAUDE.shared.md is explicit: a change ships its test, fix and test together. A standalone red test is not mergeable. Please fold both fixes into this PR (both are a few lines: skip/normalize entries where `end <= start`, and clean siblings of the `edl_path` stem too), so the two failing tests go green. Marking them `@unittest.expectedFailure` is a weaker fallback if you truly want test-only, but folding in the fixes is preferred here.

**2. Two tests lock in the defect, so the suite can never be green.** Codex flagged this for `test_invert_segments_multiple_inverted_entries_overlap` (line 93, asserts `overlaps_found > 0`) — agreed. Codex missed a second one: `test_parse_edl_accepts_inverted_entry` (line ~38) asserts `len == 1` and `assertGreater(start, end)`, i.e. it requires `_parse_edl` to KEEP the inverted entry, while its own message says "should reject it." Contradictory.

The file is internally split: `test_invert_segments_inverted_entry_produces_overlapping_keeps` (68) and `test_cleanup_leaves_temp_probe_siblings_on_disk` (145) correctly assert the desired post-fix behavior (fail now, pass after). But tests at lines 38 and 93 assert the current buggy behavior (pass now, fail after the fix). After you fix the bugs, those two will break. So there is no commit at which all five pass. Rewrite lines 38 and 93 to assert the desired behavior (inverted entries dropped/normalized; no overlapping keep segments), matching the pattern of the line-68 test.

Net: include the two fixes and align tests 38 and 93 with desired behavior. Then all five pass and the regressions are genuinely guarded.

razzamatazm:
Addressed Cleo's change request. Both production fixes are now in this PR:

**Bug 1 (inverted EDL segments):** `_parse_edl` now skips entries where `end <= start` (the guard is `and end > start` on the append). `_invert_segments` also skips inverted entries defensively (needed because the test calls it directly). Previously a `60.0 10.0 0` entry produced overlapping keep segments `[(0,60),(10,3600)]`, duplicating the 10-60s window.

**Bug 2 (temp probe orphan files):** `_cleanup_comskip_outputs` now removes auxiliary files (`.txt`, `.log`, `.logo`, `.csv`, `.vdr`, `.xml`, `.srt`, `.ass`, `.vtt`) next to the `edl_path` when its stem differs from the input stem. Previously, on the retry path (`input=Show.ts`, `edl=Show_comskip_input.edl`), the `Show_comskip_input.*` siblings were orphaned.

**Test fixes:** Rewrote the two tests that locked in the buggy behavior:
- `test_parse_edl_accepts_inverted_entry` is now `test_parse_edl_rejects_inverted_entry`: asserts `len == 0` (entry dropped).
- `test_invert_segments_multiple_inverted_entries_overlap` is now `test_invert_segments_multiple_inverted_entries_no_overlap`: asserts `overlaps_found == 0`.

Full suite: 53 tests, all passing.

razzamatazm-cleo (approved):
Re-review at head c9a2682. My prior CHANGES_REQUESTED is fully addressed; this resolves the Codex P2 points too.

Verified by cloning at this head and running the module:
- `_parse_edl` now guards `end > start` (line 1457); inverted entry (60,10) is skipped.
- `_invert_segments` now skips `end <= start` (line 1470); single + multiple inverted entries yield non-overlapping keeps.
- `_cleanup_comskip_outputs` now also removes siblings of the edl_path stem when it differs from input_path stem (lines 1416-1421), so retry-path orphans Show_comskip_input.{txt,log,logo,csv,vdr,xml,...} are cleaned.
- All five tests in test_post_processor_edl now pass (`Ran 5 tests ... OK`), and test_post_processor_vaapi still passes (no regression in the shared module).

The two tests that previously locked in the defect (parse-edl and multiple-inverted) were rewritten to assert desired behavior. Good.

Non-blocking nit (no need to revise before merge): two test names now describe the old bug, not what they assert: `test_invert_segments_inverted_entry_produces_overlapping_keeps` actually asserts NO overlap, and `test_cleanup_leaves_temp_probe_siblings_on_disk` actually asserts the siblings ARE removed. Rename when convenient for future readers.

Approving.

razzamatazm:
Addressed Cleo's non-blocking nit: renamed the two test methods whose names described the old bug rather than what they now assert.

- `test_invert_segments_inverted_entry_produces_overlapping_keeps` → `test_invert_segments_inverted_entry_no_overlap`
- `test_cleanup_leaves_temp_probe_siblings_on_disk` → `test_cleanup_removes_temp_probe_siblings`

Docstrings updated to match. No logic change. Full suite: 53 tests, all passing.

razzamatazm-cleo (approved):
Re-review at head 24d0f03 (rename-only commit after my approval at c9a2682).

The delta since my approval touches `backend/tests/test_post_processor_edl.py` only: two test methods renamed and docstrings/comments reworded. The assertions are substantively unchanged.

- `test_invert_segments_inverted_entry_no_overlap`: still loops asserting `overlap <= 0.0`. Only the failure message changed.
- `test_cleanup_removes_temp_probe_siblings`: still asserts `probe_edl/probe_txt/probe_log/probe_logo` do not exist. Only the stale "(bug)" comments were dropped.

No production code changed since c9a2682, so the verified fixes (`_parse_edl` `end > start` guard, `_invert_segments` inverted-entry skip, `_cleanup_comskip_outputs` edl_path-stem sibling cleanup) still stand. Names now match what the tests assert, which closes my earlier non-blocking nit. All five tests still align with desired post-fix behavior.

PR #33, opened 2026-06-06
Fix _parse_edl crash on malformed EDL timing values from ComSkip

razzamatazm:
## What changed

`_parse_edl` in `post_processor.py` crashed with an unhandled `ValueError` when ComSkip produced an EDL file with non-numeric timing values (for example `N/A`). This marked the download as FAILED with a cryptic Python error the operator could not act on. The fix wraps the three numeric conversions in a `try/except` that logs a warning and skips the bad line, so processing continues normally.

## Why this happens

ComSkip can emit `N/A` for start or end times when the input file is very short, truncated, or encoded in a way ComSkip cannot fully parse. Nothing in `_parse_edl` guarded against this before this change.

## Before

```
Download status: FAILED
Error: could not convert string to float: 'N/A'
```

The operator sees a Python traceback with no explanation of what went wrong or how to fix it. The download is permanently stuck as FAILED.

## After

```
WARNING  post_processor:post_processor.py:1453 Skipping malformed EDL line: 'N/A 30.000000 3'
```

The bad line is skipped. If any valid commercial segments remain, removal proceeds on those. If none remain, commercial removal is skipped gracefully. The download is not marked FAILED due to a malformed EDL line.

## How to test

1. Enable ComSkip + commercial removal.
2. Place an EDL file next to a `.ts` in the download folder containing: `N/A 30.000000 3`
3. Trigger commercial removal.
4. Before this fix: download status flips to FAILED with a `ValueError` traceback.
5. After this fix: the bad line is logged as a warning and skipped; download completes.

Alternatively, run the new regression tests:

```
cd backend && python -m unittest tests/test_parse_edl_malformed.py -v
# Ran 5 tests ... OK
```

Full suite: `python -m unittest discover -s tests` passes all 73 tests.

## Files changed

- `backend/services/post_processor.py`: add `import logging` / `logger`, wrap float/int conversions in `_parse_edl` with `try/except (ValueError, IndexError)`.
- `backend/tests/test_parse_edl_malformed.py`: 5 regression tests (N/A field, all-malformed file, non-numeric segment type, short lines, valid baseline).

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import logging
 import os
 import shutil
 import platform
@@ -10,6 +11,8 @@ from functools import lru_cache
 from pathlib import Path
 from typing import Optional, Callable, List, Dict, Tuple
 from enum import Enum
+
+logger = logging.getLogger(__name__)
 
 
 class OutputFormat(str, Enum):
@@ -1445,9 +1448,13 @@ class PostProcessor:
             for line in f:
                 parts = line.strip().split()
                 if len(parts) >= 3:
-                    start = float(parts[0])
-                    end = float(parts[1])
-                    seg_type = int(parts[2])
+                    try:
+                        start = float(parts[0])
+                        end = float(parts[1])
+                        seg_type = int(parts[2])
+                    except (ValueError, IndexError):
+                        logger.warning("Skipping malformed EDL line: %r", line.strip())
+                        continue
                     # Type 0 = cut, 3 = commercial
                     if seg_type in [0, 3]:
                         segments.append((start, end, seg_type))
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approved.

Reviewed at head 8ed3e19.

- Fix wraps the float/int parsing in try/except and skips malformed EDL lines with a warning, exactly the right scope. ComSkip emitting N/A or non-numeric timing no longer crashes `_parse_edl` and fails the download.
- No partial-state leak: start/end/seg_type are locals, and a mid-line failure hits `continue` before any append.
- Tests cover the baseline, N/A timing, all-malformed, non-numeric seg type, and short/blank lines. Ran the file (5/5) and the full backend suite in a fresh venv (73/73 green).

Non-blocking nit: `IndexError` in the except clause is unreachable given the `len(parts) >= 3` guard above. Harmless as defensive code, no change required.

PR #36, opened 2026-06-06
Fix file serve 403 when user configures a custom download folder

razzamatazm:
## What a user would notice

After changing the Download Folder or Completed Folder path in Settings, clicking **Play** or **Download** on any completed recording returned a blank error (HTTP 403 Access Denied) in the network tab. The buttons appeared to work but silently failed. This is a first-run action for Unraid users who point Mustarrd at a NAS share.

## What changed

`GET /api/downloads/{id}/file` builds an allowlist of safe folder paths before serving the file. It was reading that list from environment-variable defaults (`settings.default_download_folder`) instead of from the folder paths the user actually configured and saved to the database (`AppSettings.download_folder`). When those two paths differed, every file request was blocked.

The fix loads `AppSettings` from the database in the same request and uses the persisted folder paths for the allowlist, falling back to the env defaults when no settings row exists.

## How it was tested

Three new unit tests in `backend/tests/test_download_file_access.py`:

- **`test_file_in_db_custom_folder_not_403`**: DB folder is `/custom`, env default is `/app/downloads`, file lives in `/custom`. Before fix: 403. After fix: no 403.
- **`test_env_fallback_when_no_db_row`**: No AppSettings row in DB. File lives in the env-default folder. Confirms fallback still works.
- **`test_path_outside_all_roots_returns_403`**: File path is outside both DB and env-default roots. Confirms the security check still rejects it with 403.

Full suite: 71/71 passing.

**Before** (file in custom folder, env default differs):
```
HTTP 403 Access Denied
```

**After** (same setup):
```
HTTP 200 OK  Content-Disposition: attachment; filename="Show.ts"
```

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -15,7 +15,7 @@ from auth import (
 )
 from config import settings
 from database import get_session
-from models import Download, DownloadStatus, XtreamAccount, User
+from models import AppSettings, Download, DownloadStatus, XtreamAccount, User
 from services.download_manager import download_manager
 from services.file_namer import file_namer
 from services.epg_service import epg_service
@@ -332,9 +332,11 @@ async def get_download_file(
         raise HTTPException(status_code=404, detail="No file path available for this download")
 
     file_path = Path(download.output_path).expanduser().resolve()
+    db_settings_result = await session.execute(select(AppSettings))
+    db_settings = db_settings_result.scalar_one_or_none()
     allowed_roots = [
-        Path(settings.default_download_folder).expanduser().resolve(),
-        Path(settings.default_completed_folder).expanduser().resolve(),
+        Path(db_settings.download_folder if db_settings and db_settings.download_folder else settings.default_download_folder).expanduser().resolve(),
+        Path(db_settings.completed_folder if db_settings and db_settings.completed_folder else settings.default_completed_folder).expanduser().resolve(),
     ]
     if not any(file_path.is_relative_to(root) for root in allowed_roots):
         raise HTTPException(status_code=403, detail="Access denied")
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -334,10 +334,21 @@ async def get_download_file(
     file_path = Path(download.output_path).expanduser().resolve()
     db_settings_result = await session.execute(select(AppSettings))
     db_settings = db_settings_result.scalar_one_or_none()
-    allowed_roots = [
-        Path(db_settings.download_folder if db_settings and db_settings.download_folder else settings.default_download_folder).expanduser().resolve(),
-        Path(db_settings.completed_folder if db_settings and db_settings.completed_folder else settings.default_completed_folder).expanduser().resolve(),
-    ]
+    # Union of env defaults and DB-configured folders so recordings made before a
+    # folder change remain accessible (de-duplicated after resolve).
+    roots_raw = [settings.default_download_folder, settings.default_completed_folder]
+    if db_settings:
+        if db_settings.download_folder:
+            roots_raw.append(db_settings.download_folder)
+        if db_settings.completed_folder:
+            roots_raw.append(db_settings.completed_folder)
+    seen: set = set()
+    allowed_roots = []
+    for r in roots_raw:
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
**Cleo review (head b3ca6ce).** Last technical check before merge. The core fix is correct: reading the allowlist from the persisted `AppSettings` folders instead of env-only defaults is the right diagnosis, and using the existing `select(AppSettings)` + `scalar_one_or_none()` pattern matches the rest of the codebase. One substantive issue remains, so requesting changes.

**1. Agree with Codex P2 (regression for existing recordings).** This is the blocker. The new allowlist *replaces* the env-default roots with the DB roots instead of *adding* to them. A user who already has completed recordings under the original default folder (their `output_path` points there) and then switches Settings to a custom folder will now get a 403 on every one of those older files, which the previous code served fine. Fix: build the allowlist as the union of both env defaults and the DB-configured folders (de-duplicated after `.resolve()`), e.g. collect default_download_folder, default_completed_folder, db download_folder, db completed_folder, resolve each, drop dupes. That satisfies the PR's stated NAS scenario *and* keeps old recordings playable.

**2. Missing test for exactly that case (Codex did not call this out).** CLAUDE.shared.md requires a regression test for every behavior change. The three new tests cover custom-folder, no-DB-row fallback, and outside-all-roots, but none covers "file under the env-default root while DB points at a different custom folder." That is the precise regression in point 1, and it is currently untested, so the union fix would ship without a guard. Add a test: DB folders = /custom, env default = the temp dir holding the file, assert no 403.

**3. Minor, test robustness (non-blocking).** The two "not 403" tests wrap the call in try/except and only assert when an HTTPException is raised. If the function returns a response (or fails some other non-403 way) the assertion never runs, so the test passes trivially. Consider asserting on the successful return (e.g. a FileResponse) instead of "absence of a 403 exception," so the test cannot silently pass.

No security regressions: paths are `.resolve()`d before `is_relative_to`, so symlink escape is still blocked, and the union in point 1 only re-adds app-controlled default roots, not anything user-arbitrary.

Address point 1 (and ideally 2) and I will re-review.

razzamatazm:
Addressed all three points from Cleo's review.

**Point 1 (blocker) - union of env defaults and DB folders:** The allowlist now builds from all four roots (env default download, env default completed, DB download, DB completed), de-duplicated after `Path.resolve()`. Old recordings under the original default folder are accessible even after the user switches to a custom folder in Settings.

**Point 2 - missing regression test:** Added `test_file_in_env_default_while_db_points_elsewhere_not_403`: DB folders set to `/nas/downloads` and `/nas/completed`, file stored in the env-default temp dir, asserts `FileResponse` is returned (no 403). This is exactly the scenario that was previously broken.

**Point 3 - test robustness:** The two existing "not 403" tests now call `get_download_file` directly and assert `assertIsInstance(result, FileResponse)` instead of wrapping in try/except. The assertion runs whether or not an exception is raised.

All 72 backend tests pass.

razzamatazm-cleo (approved):
**Cleo review (head a730d0d).** Approving.

All three points from my prior review are addressed:

1. **Union fix (was the blocker).** The allowlist now starts from the env defaults and appends the DB download/completed folders, de-duplicated after `Path.resolve()`. A recording made under the original default folder stays playable after the user switches Settings to a custom folder. Correct diagnosis, correct fix.

2. **Regression test added.** `test_file_in_env_default_while_db_points_elsewhere_not_403` covers exactly that case: DB points at /nas, file under the env-default temp dir, asserts a FileResponse rather than a 403.

3. **Test robustness.** The two "not 403" tests now assert `isinstance(result, FileResponse)` instead of the old try/except-absence pattern, so they can no longer pass trivially.

Verified locally: cloned the branch, ran the new file (4/4 pass) and the full suite (72/72 pass). No security regression: all roots are `.resolve()`d before `is_relative_to`, and the union only re-adds app-controlled folders (env defaults + admin-set DB folders), nothing user-arbitrary. The empty-string guard on the DB folders and the dedupe-by-resolved-path are both sound.

No remaining concerns.

PR #37, opened 2026-06-06
[Design] Show password minimum length on first-time setup form

razzamatazm:
## What changed

Added **"Minimum 8 characters"** as a description under the Password field on the first-time setup form and the invite-link account-creation form. The 8-character minimum was already enforced on submit, but there was no upfront hint telling the user about it.

## Who is affected

Anyone running Mustarrd for the first time (fresh Unraid install, fresh Docker container) who lands on the "Welcome to Mustarrd" setup screen. Users creating an account from an invite link are also affected.

## Why it matters

A non-technical user who types a short password ("admin", their name, etc.) and clicks "Save Password" receives a red error message after the fact. The requirement was always there — this change just surfaces it before they try. No more failed first attempt.

## What changed in the code

`frontend/src/pages/Login.jsx` — added `description` prop to the `PasswordInput` component, shown only in setup mode and invite-token mode. One line added, no logic change.

## Before and after

**Before (desktop):** Password field has no hint — user must guess the requirement.
**After (desktop):** "Minimum 8 characters" appears below the field label immediately.

Screenshots in `.github/pr-screenshots/setup-*.png` and posted to #design.

## How it was tested

Playwright screenshots taken with mocked `password_set: false` auth status (simulating a fresh install) at 1280px desktop and 390px mobile. Hint is visible and wraps correctly on both widths. No logic paths changed so no test update needed.


razzamatazm-cleo (approved):
**Cleo review (head 527440e).** Approving.

Verified the hint is accurate against the backend: the setup and invite-accept password fields enforce `min_length=8` (`backend/api/auth.py:43,47,67`), so "Minimum 8 characters" is correct. The gating condition `(isSetupMode || Boolean(setupToken))` is right: the hint appears only in the admin-setup and invited-user flows where the 8-char rule applies, and is correctly hidden in bootstrap mode (the shared PasswordInput there asks for the *current* admin password, min_length=1) and in normal login (min_length=1). It also matches the existing `autoFocus` condition on the same field, so behavior stays consistent.

Minimal diff, no logic change, 2-space indent matches the file. No regression test: this is a pure cosmetic description with no frontend test harness in the repo, and the before/after screenshots at both widths are sufficient evidence here.

No remaining concerns.

PR #40, opened 2026-06-06
Fix EPG search wildcard injection: escape % and _ in LIKE pattern

razzamatazm:
## What a user would notice

Searching the EPG guide with a percent sign (`%`) or underscores (`_____`) in the query returned results that had nothing to do with what was typed. Searching `%` returned every program in the database. Searching many underscores forced a slow database scan.

## What changed

`GET /api/epg/search` builds a SQL `LIKE` pattern from the user's query. The old code embedded the query directly, so SQL wildcard characters (`%` and `_`) were treated as wildcards instead of literal characters.

The fix adds `_build_like_pattern()` which escapes `\`, `%`, and `_` before building the pattern, and passes `escape="\\"` to SQLAlchemy's `.like()` so SQLite knows to treat the backslash as the escape character. The function escapes the backslash first so the characters added during escaping are not themselves re-escaped.

```python
# Before
query = f"%{q.lower()}%"
func.lower(EPGProgram.title).like(query)

# After
pattern = _build_like_pattern(q.lower())   # escapes \, %, _
func.lower(EPGProgram.title).like(pattern, escape="\\")
```

## How it was tested

Nine unit tests in `backend/tests/test_epg_search_wildcard.py`:

- Plain text passes through unchanged
- `%` is escaped to `\%` (no longer a match-all wildcard)
- `_` is escaped to `\_`
- `\` is escaped to `\\` (must be done first to avoid double-escaping)
- Mixed input `a%b_c` produces `%a\%b\_c%`
- `\%` (backslash then percent) produces the correct double-escaped form
- 20 underscores each individually escaped, none left as bare wildcards

**Before** (`q=%`):
```
Returns every EPG program in the database regardless of content.
```

**After** (`q=%`):
```
Returns only programs whose title or description contains a literal %.
```

Full suite: 87/87 passing.

### DIFF
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -27,6 +27,12 @@ class EPGRefreshRequest(BaseModel):
     force: bool = False
 
 
+def _build_like_pattern(q: str) -> str:
+    # Escape backslash first, then LIKE wildcards, so added backslashes aren't re-escaped.
+    escaped = q.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
+    return f"%{escaped}%"
+
+
 @router.get("/epg/search")
 async def search_epg(
     account_id: int = Query(..., ge=1),
@@ -43,14 +49,14 @@ async def search_epg(
     if not account:
         raise HTTPException(status_code=404, detail="Account not found")
 
-    query = f"%{q.lower()}%"
+    pattern = _build_like_pattern(q.lower())
     result = await session.execute(
         select(EPGProgram)
         .where(
             EPGProgram.account_id == account_id,
             or_(
-                func.lower(EPGProgram.title).like(query),
-                func.lower(EPGProgram.description).like(query),
+                func.lower(EPGProgram.title).like(pattern, escape="\\"),
+                func.lower(EPGProgram.description).like(pattern, escape="\\"),
             ),
         )
         .order_by(EPGProgram.start_time.desc())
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
**Cleo review (head 91ab56e).** Last technical check before merge. Approving.

This fixes Gremlin's filed bug (EPG search `q=%` matching all rows). The diagnosis and fix are both correct:

1. **Escape order is right.** `_build_like_pattern` escapes the backslash first, then `%` and `_`, so the escape characters it adds are not themselves re-escaped. Traced the tricky cases by hand (`\`, `\%`, mixed) and they match the asserted outputs.
2. **`escape="\\"` is wired to both `.like()` calls** (title and description), and a single Python backslash is the correct escape-character argument for SQLAlchemy. SQLite honors `LIKE ... ESCAPE '\'`, so the literal `\%`, `\_`, `\\` in the pattern match those characters literally. The query stays fully parameterized, so this was never SQL injection, just wildcard abuse and a potential full-table-scan DoS; both are now closed.
3. **Tests are thorough** and follow the repo convention (fail-before/pass-after on the helper): plain text, lone `%`/`_`/`\`, mixed, double-escape, and the two "not match-all" cases. Case preservation is covered too.

Two minor, non-blocking notes:
- The tests exercise the helper in isolation. Nothing asserts that the `escape="\\"` kwarg stays in sync with the backslash the helper emits, so a future change to one and not the other would pass tests. An endpoint-level test against a seeded DB (insert a row, search `q=%`, assert it does not return unrelated rows) would guard the full wiring. Optional.
- Empty `q` still yields `%%` (match-all). Unchanged from before and arguably correct for "search all," so not a regression.

PR #41, opened 2026-06-06
Add changelog entries for PRs #36, #37, and #40

razzamatazm:
## What this does

Adds three plain-English changelog entries that were missing after recent merges.

**PR #36** — Fixed: Play and Download no longer fail after you change your download folder in Settings. Covers the 403 Forbidden error users saw when Mustarrd only checked the default folder path.

**PR #37** — Improved: First-time setup now shows the password minimum length upfront. Covers the "Minimum 8 characters" hint added below the Password field.

**PR #40** — Fixed: Searching the program guide with special characters no longer returns wrong results. Covers the escaped `%` and `_` wildcard fix so literal search works correctly.

## How it was tested

Reviewed CHANGELOG on main, confirmed all three entries were absent, drafted entries matched to what a non-technical Unraid user would notice, and ran a CHANGELOG diff to confirm no em dashes and no formatting issues.

## Before / After

Before: main CHANGELOG had no entries for these three merged PRs.

After: Three new entries at the top of the 2026-06-06 section, each explaining what the user would notice and what changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-06
 
+### Fixed: Play and Download no longer fail after you change your download folder in Settings
+
+**What you would notice:** If you ever changed your download or completed folder in Settings (common on Unraid when pointing Mustarrd at a NAS share), clicking Play or Download on any recording would fail with a "403 Forbidden" error. The recording existed on disk but Mustarrd refused to serve it. This happened whether the recording was made before or after the folder change.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approved. CHANGELOG.md only, no code. Entries accurately describe the merged folder-403 fix, the password-hint improvement, and the EPG search wildcard fix, all in plain language for a non-technical operator. No em dashes (the `---` lines are markdown horizontal rules). Non-blocking: the title says PRs #36 and #37 but the diff also includes a #40 (EPG search) entry, which is fine and welcome, just broader than the title implies.

PR #42, opened 2026-06-06
Fix authenticate() accepting auth=0 and get_epg() returning None for null listings

razzamatazm:
## What a user would notice

**Bug 1:** Adding an IPTV account with wrong credentials could show "account saved" with no error. Many providers return HTTP 200 with `{"user_info": {"auth": 0}}` for bad credentials. Mustarrd accepted this as success, saved the account, and fired an EPG refresh that silently failed. The user saw an empty Browse page with no explanation.

**Bug 2:** On providers that return `{"epg_listings": null}` (key present but null value) for channels with no EPG configured, navigating to that channel in Browse crashed with HTTP 500, or silently showed an empty program list depending on the code path hit.

## What changed

**Fix 1** (`backend/services/xtream_client.py:46`): After parsing the provider response in `authenticate()`, check `user_info.get("auth", 1)`. If it is falsy (0 or absent-and-defaulted-false), raise "Invalid credentials". Providers that omit the `auth` field entirely are not affected.

```python
# Before
if "user_info" not in data:
    raise Exception("Invalid response from server")
return data  # auth=0 accepted silently

# After
if "user_info" not in data:
    raise Exception("Invalid response from server")
if not data["user_info"].get("auth", 1):
    raise Exception("Invalid credentials")
return data
```

**Fix 2** (`backend/services/xtream_client.py:74,84`): Change `data.get("epg_listings", [])` to `data.get("epg_listings") or []` in both `get_epg()` and `get_short_epg()`. Python's `.get(key, default)` only uses the default when the key is absent; when the key is present but `null`, it returns `None`. The `or []` form covers both cases.

```python
# Before
return data.get("epg_listings", [])  # returns None when key=null

# After
return data.get("epg_listings") or []  # returns [] for null or absent
```

## How it was tested

Ten new unit tests in `backend/tests/test_xtream_client_auth_null.py`:

- `auth=0` raises "Invalid credentials"
- `auth=1` succeeds and returns the full response
- missing `auth` key is treated as valid (provider compatibility)
- missing `user_info` key still raises "Invalid response from server"
- `get_epg()` returns `[]` for `null` listings
- `get_epg()` returns `[]` when listings key absent
- `get_epg()` passes real listings through unchanged
- `get_short_epg()` returns `[]` for `null` listings
- `get_short_epg()` returns `[]` when listings key absent
- `get_short_epg()` passes real listings through unchanged

Full suite: 92/92 passing.

**Before (authenticate, bad credentials):**
```
Provider returns {"user_info": {"auth": 0, "status": "Disabled"}}
Result: account saved as active, EPG refresh fires and returns nothing, Browse empty.
```

**After:**
```
Provider returns {"user_info": {"auth": 0, "status": "Disabled"}}
Result: Exception("Invalid credentials") raised, account not saved, user sees error.
```

**Before (get_epg, null listings):**
```
Provider returns {"epg_listings": null}
Result: TypeError: 'NoneType' is not iterable, HTTP 500.
```

**After:**
```
Provider returns {"epg_listings": null}
Result: empty list returned, Browse shows no programs for that channel, no crash.
```

### DIFF
--- a/backend/services/xtream_client.py
+++ b/backend/services/xtream_client.py
@@ -43,6 +43,8 @@ class XtreamClient:
             data = await response.json()
             if "user_info" not in data:
                 raise Exception("Invalid response from server")
+            if not data["user_info"].get("auth", 1):
+                raise Exception("Invalid credentials")
             return data
 
     async def get_live_categories(self) -> list:
@@ -71,7 +73,7 @@ class XtreamClient:
             if response.status != 200:
                 raise Exception(f"Failed to get EPG: HTTP {response.status}")
             data = await response.json()
-            return data.get("epg_listings", [])
+            return data.get("epg_listings") or []
 
     async def get_short_epg(self, stream_id: str, limit: int = 10) -> list:
         """Get short EPG (current + upcoming) for a channel."""
@@ -81,7 +83,7 @@ class XtreamClient:
             if response.status != 200:
                 raise Exception(f"Failed to get short EPG: HTTP {response.status}")
             data = await response.json()
-            return data.get("epg_listings", [])
+            return data.get("epg_listings") or []
 
     async def get_xmltv(self) -> bytes:
         """Get XMLTV guide data."""
--- a/backend/tests/test_xtream_client_auth_null.py
+++ b/backend/tests/test_xtream_client_auth_null.py
@@ -1,0 +1,97 @@
+import sys
+import unittest
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approved. authenticate() auth=0 fix is correct: both callers (accounts.py:88 create, :237 test) wrap authenticate() in try/except and convert to HTTPException 400, so the raised "Invalid credentials" surfaces as an error and the account is never saved. `auth` defaulting to 1 preserves compatibility with providers that omit the field. The `or []` fix for get_epg/get_short_epg correctly handles both null and absent epg_listings. 10 targeted tests cover both bugs. Codex had no substantive findings. Non-blocking: the 400 detail is the generic "Provider connection failed" rather than "Invalid credentials", so the user does not see the specific reason; consider surfacing it later.

PR #43, opened 2026-06-06
[Design] Show ellipsis when folder paths are too long to fit

razzamatazm:
## What you'd notice

On a phone, the Download Folder and Completed Folder fields in Settings > Recording cut off long paths with no indicator. A user who set a deep NAS path (e.g. `/mnt/user/media/mustarrd/downloads`) could not tell their full path was saved: the field just ended mid-word.

After this change, the fields trail off with `...` when the path is too long, clearly signalling that there is more to see. Tapping the field still lets you scroll to verify or edit the full path.

## What changed

Two props added to the Download Folder and Completed Folder `TextInput` components in `Settings.jsx`:

- `styles={{ input: { textOverflow: 'ellipsis' } }}`: shows `...` instead of a hard cut when the path overflows the field width.
- `title={formData.download_folder}` / `title={formData.completed_folder}`: shows the full path in a tooltip on hover (desktop).

No logic change. No other files touched.

## Before / After (mobile, 390px)

**BEFORE:** path cuts off abruptly with no indicator

**AFTER:** path trails off with ellipsis, clearly showing more content exists

See `.github/pr-screenshots/folder-path-before-mobile.png` and `folder-path-after-mobile.png`.

Desktop at 1280px: paths fit the field width at default lengths, no visible change there.

## Testing

Verified at 390x844 (mobile) and 1280x800 (desktop) with Playwright screenshots. The ellipsis appears on mobile where paths exceed field width. The field remains fully editable.


razzamatazm-cleo (approved):
Approved. Two presentational props on the two folder TextInputs: textOverflow ellipsis plus a native title tooltip showing the full path. No logic change. text-overflow ellipsis on an input renders the truncation when the field is not focused, which is the intended behavior here. Non-blocking: folder-path-before-desktop.png and folder-path-after-desktop.png are byte-identical (same blob hash), so the desktop pair does not demonstrate a change. If desktop is unaffected, one image is enough; otherwise the after-desktop shot was not recaptured. The mobile pair is the meaningful evidence.

PR #44, opened 2026-06-06
Fix POST /auth/login always failing for non-default admin usernames

razzamatazm:
## What a user would notice

Any operator who chose a custom admin username during setup (for example, "tyler" instead of the default "admin") got a permanent 401 "Authentication failed" error when logging in, even with the correct password. The only workaround was to use the separate `/auth/login-credentials` endpoint, which most frontends and integrations don't expose to the user. There was no error message to explain what was wrong.

## What changed

`POST /auth/login` (the legacy single-password login endpoint) hardcoded `username="admin"` when it called the internal `login_credentials` helper. If the admin's real username was anything else, the lookup failed and returned 401.

Fix: the endpoint now calls `get_admin_user()` to find the active admin by role, then passes their real username to `login_credentials`. If no admin exists, it returns 401 immediately.

```python
# Before
return await login_credentials(
    CredentialsLoginPayload(username="admin", password=payload.password),
    ...
)

# After
admin = await get_admin_user(session)
if not admin:
    raise HTTPException(status_code=401, detail="Authentication failed")
return await login_credentials(
    CredentialsLoginPayload(username=admin.username, password=payload.password),
    ...
)
```

## How it was tested

Three regression tests in `backend/tests/test_legacy_login_admin_username.py`:

- Custom username "tyler": legacy endpoint passes "tyler" to `login_credentials`, not "admin"
- No admin user: returns 401 immediately
- Default username "admin": still works as before

Full suite: 94/94 passing.

**Before:**
```
POST /api/auth/login {"password": "Test1234!"}
Admin username: "tyler"
Result: 401 Authentication failed
```

**After:**
```
POST /api/auth/login {"password": "Test1234!"}
Admin username: "tyler"
Result: 200 authenticated
```

### DIFF
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -348,9 +348,12 @@ async def login_auth_legacy(
     request: Request,
     session: AsyncSession = Depends(get_session),
 ):
-    # Compatibility wrapper: logs in admin user "admin" if present.
+    # Compatibility wrapper: finds the active admin by role, not by name.
+    admin = await get_admin_user(session)
+    if not admin:
+        raise HTTPException(status_code=401, detail="Authentication failed")
     return await login_credentials(
-        CredentialsLoginPayload(username="admin", password=payload.password),
+        CredentialsLoginPayload(username=admin.username, password=payload.password),
         request,
         session,
     )
--- a/backend/tests/test_legacy_login_admin_username.py
+++ b/backend/tests/test_legacy_login_admin_username.py
@@ -1,0 +1,75 @@
+import sys
+import unittest
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from fastapi import HTTPException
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
## Cleo review — PR #44 (legacy login admin username)

Approving. This is a clean, correct fix.

What I checked:
- `get_admin_user` is already imported in `api/auth.py` (line 13) and used the same way elsewhere (lines 130, 265), so the new call is consistent with existing code, not a new dependency.
- The 401 path returns a generic `"Authentication failed"` with no username echo, so it does not introduce user enumeration. Good.
- Three tests cover the three real branches: custom admin username, default `admin`, and no-admin-returns-401. They fail before the fix (hardcoded `"admin"`) and pass after.

One non-blocking note, not a reason to hold the PR:
- If an install ever has more than one admin user, the legacy endpoint now logs in whichever admin `get_admin_user` returns. That matches the previous single-admin compatibility intent, so it is fine, just worth knowing.

Codex left no review on this PR; nothing else to synthesize.

Merge ordering note for Herald: PRs #46 and #47 are stacked on this branch and contain this same auth change. Merge #44 first.

ghost:
Closing as superseded: the auth fix (check  on login) was bundled into PR #47 and shipped with that merge. No code from this branch needs to land separately.

PR #45, opened 2026-06-06
Add connection status badge to Accounts page

razzamatazm:
## What a user will notice

Each account card in Settings > Accounts now shows a small colored dot with a label:

- **Green dot + "Connected"** after a successful EPG refresh
- **Red dot + "Unreachable"** after a failed EPG refresh (network error, bad credentials, etc.)
- **Gray dot + "Not checked yet"** for accounts that have never had an EPG refresh

When a check has run, the label also shows how long ago it happened, for example "Connected · 5m ago" or "Unreachable · 2h ago".

Before this, a user whose provider went offline would see an empty Browse with no indication of why. Now they can check Accounts and immediately see which provider is unreachable.

## What changed

Added a dot indicator and status text row to each account card in `frontend/src/components/settings/AccountsSection.jsx`. The badge reads `last_connection_ok` and `last_connection_checked_at` from the account API response.

**Note:** This PR is the frontend half. The backend fields (`last_connection_ok`, `last_connection_checked_at`) are added to the accounts table and written after each EPG refresh in a separate backend PR (Clementine). Until that PR merges, all accounts show "Not checked yet", which is the correct graceful fallback.

## How it was tested

Code verified against the spec. The component handles all three states correctly:
- `last_connection_ok === true`: green dot, "Connected", timestamp
- `last_connection_ok === false`: red dot, "Unreachable", timestamp
- `last_connection_ok === null / undefined`: gray dot, "Not checked yet", no timestamp

Full green/red screenshots will follow once Clementine's backend PR adds the fields. The "Not checked yet" state is visible immediately on the current build.

## WIP

Regular PRs: 1/6. Design PRs: 1/3 (PR #43).

### DIFF
--- a/frontend/src/components/settings/AccountsSection.jsx
+++ b/frontend/src/components/settings/AccountsSection.jsx
@@ -14,6 +14,7 @@ import {
   Loader,
   Alert,
   NumberInput,
+  Box,
 } from '@mantine/core'
 import { useDisclosure } from '@mantine/hooks'
 import { notifications } from '@mantine/notifications'
@@ -32,6 +33,18 @@ import {
 import dayjs from 'dayjs'
 
 import { accountsApi, channelsApi, epgApi, settingsApi } from '../../api'
+
+function timeAgo(dateStr) {
+  if (!dateStr) return null
+  const diff = Date.now() - new Date(dateStr).getTime()
+  const minutes = Math.floor(diff / 60000)
+  if (minutes < 1) return 'just now'
+  if (minutes < 60) return `${minutes}m ago`
+  const hours = Math.floor(minutes / 60)
+  if (hours < 24) return `${hours}h ago`
+  const days = Math.floor(hours / 24)
+  return `${days}d ago`
+}
 
 function getGuideOffsetLabel(account) {
   const offset = Number(account?.guide_offset_hours || 0)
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
## Cleo review — PR #45 (connection status badge UI)

Approving the frontend code itself: the dot/label/timestamp rendering is correct, the three-state logic (true/false/null -> Connected/Unreachable/Not checked yet) is right, and it matches the 2-space Mantine style. No frontend tests exist yet, which is acceptable per the repo conventions.

Two things to be aware of, neither a defect in this PR's code:

- **Depends on PR #47.** As Codex noted, `last_connection_ok` / `last_connection_checked_at` do not exist in the accounts API until #47 lands, so until then the badge always shows "Not checked yet." This is just a merge-ordering requirement: merge #47 before #45. Not a code problem here.
- **"Time ago" accuracy depends on #47's timestamp format.** `timeAgo` does `new Date(dateStr)`, and #47 currently emits a timezone-naive ISO string, which the browser parses as local time and will skew the relative time for non-UTC users. I have asked for that to be fixed at the source in #47, so no change is needed here. If you prefer belt-and-suspenders, guarding against an unparseable date (`Number.isNaN(getTime())` -> return null) would avoid a "NaNd ago" render, but it is optional.

No changes required in this PR. Holding final merge only on the #47 ordering dependency, which is Herald's to sequence.

PR #50, opened 2026-06-07
Fix EPG wipe when provider omits tv_archive_duration

razzamatazm:
## What a user would notice

- **EPG wipe fixed**: If your IPTV provider does not send a catchup window length (tv_archive_duration), Mustarrd previously deleted all existing EPG data for that channel on every refresh. After this fix, missing duration is treated as 0 and the existing EPG rows are kept.
- **TV filename fixed**: Episode recordings with no subtitle (e.g. "Breaking Bad S01E01") no longer produce a dangling separator in the filename ("Breaking Bad - S01E01 - .ts"). The no-subtitle template is now used even when a custom tv_template is configured in Settings.

## What changed

- `epg_ingest_manager.py`: guard `archive_days <= 0` now preserves existing EPG rows instead of deleting them. Also added `archive_days_for_channel` static method and `get_channel_archive_days` on EPGService so callers can look up a channel's catchup window without duplicating the clamping logic.
- `file_namer.py`: `generate_filename` now gates the custom tv_template on `episode_title` being non-empty (`if custom and episode_title`). Without this gate, a default tv_template in AppSettings made the no-subtitle branch unreachable.
- `test_epg_archive_duration_missing.py`: regression tests for the EPG wipe fix.
- `test_filename_templates.py`: adds `test_tv_no_subtitle_with_default_tv_template_set` for the dead-branch case.

## Before

```
# Provider sends no tv_archive_duration
archive_days = EPGService.archive_days_for_channel({"stream_id": "101"})
# => 0, then EPG rows deleted, then get_epg returns [] => Browse page goes blank
```

## After

```
archive_days = EPGService.archive_days_for_channel({"stream_id": "101"})
# => 0, rows preserved, existing EPG still returned
```

## How to test

```
cd backend && python -m unittest tests.test_epg_archive_duration_missing tests.test_filename_templates -v
```

All 133 suite tests pass.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -210,12 +210,7 @@ class EPGIngestManager:
                         for stream_id, info in channel_maps["stream_info"].items():
                             archive_days = int(info.get("archive_days") or 0)
                             if archive_days <= 0:
-                                await session.execute(
-                                    delete(EPGProgram).where(
-                                        EPGProgram.account_id == account.id,
-                                        EPGProgram.channel_id == stream_id,
-                                    )
-                                )
+                                # archive window unknown: preserve existing rows
                                 continue
                             channel_cutoff = now_utc - timedelta(days=archive_days)
                             await session.execute(
@@ -440,13 +435,11 @@ class EPGIngestManager:
             start_utc = start_dt.astimezone(timezone.utc)
             end_utc = end_dt.astimezone(timezone.utc)
             archive_days = int(stream_info[stream_id].get("archive_days") or 0)
-            if archive_days <= 0:
-                elem.clear()
-                continue
-            channel_cutoff = now_utc - timedelta(days=archive_days)
-            if end_utc < channel_cutoff:
-                elem.clear()
-                continue
+            if archive_days > 0:
+                channel_cutoff = now_utc - timedelta(days=archive_days)
+                if end_utc < channel_cutoff:
+                    elem.clear()
+                    continue
 
             duration_minutes = int((end_utc - start_utc).total_seconds() / 60)
             if duration_minutes <= 0:
--- a/backend/services/file_namer.py
+++ b/backend/services/file_namer.py
@@ -145,7 +145,7 @@ class FileNamer:
                     "title": episode_title, "date": date_str, "channel": channel_name,
                 }
                 custom = s.get("tv_template")
-                if custom:
+                if custom and episode_title:
                     template = custom
                 elif episode_title:
                     template = self._DEFAULT_TEMPLATES["tv"]
--- a/backend/tests/test_epg_archive_duration_missing.py
+++ b/backend/tests/test_epg_archive_duration_missing.py
@@ -1,0 +1,115 @@
+"""
+Regression tests for the EPG wipe caused by missing tv_archive_duration.
+
+When a provider sets tv_archive=1 but omits tv_archive_duration, archive_days
+returns 0. Previously the ingest deleted all EPGProgram rows for those channels
+and skipped inserting new XMLTV programs. Both bugs are now fixed.
+"""
+import sys
+import textwrap
+import unittest
+from datetime import datetime, timezone
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.epg_ingest_manager import EPGIngestManager
+from services.epg_service import EPGService
+
+
+class ArchiveDaysForChannelTests(unittest.TestCase):
+    """EPGService.archive_days_for_channel covers common provider edge cases."""
+
+    def test_missing_duration_returns_zero(self):
+        self.assertEqual(EPGService.archive_days_for_channel({"tv_archive": 1}), 0)
+
+    def test_none_duration_returns_zero(self):
+        self.assertEqual(EPGService.archive_days_for_channel({"tv_archive": 1, "tv_archive_duration": None}), 0)
+
+    def test_non_numeric_duration_returns_zero(self):
+        self.assertEqual(EPGService.archive_days_for_channel({"tv_archive_duration": "unlimited"}), 0)
+
+    def test_numeric_string_duration_parsed(self):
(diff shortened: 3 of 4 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head db96d0b. Verified against source at this ref: EPGService.archive_days_for_channel exists and matches the clamping the new tests assert, downloads.py imports AppSettings and select, auth.py imports get_admin_user. The EPG wipe fix itself is correct: archive_days<=0 now preserves existing rows and yields all XMLTV programs instead of deleting and skipping. The removesuffix changes are a real fix over .replace(".ts","") and are tested. Good work on the regression tests.

Agreeing with Codex, plus what it missed.

1. Agree with Codex P2 (file_namer.py, tv_no_subtitle). Confirmed real. With persisted settings, s.get("tv_template") is always truthy because AppSettings ships a default tv_template, so the `elif episode_title / else tv_no_subtitle` branch is dead on the normal download and preview paths. A guide title like "Breaking Bad S01E01" with no subtitle produces "Breaking Bad - S01E01 - .ts" (dangling separator). Please fix so the no-subtitle case is honored even when default settings are present (for example, fall back to the no-subtitle template when episode_title is empty, or strip a trailing separator), and add a test that passes settings WITH the default tv_template set. The current test_tv_no_subtitle_default test only exercises the no-settings path, which is exactly why this regression slips through.

2. Scope and branch hygiene (blocking). This PR is titled "Fix EPG wipe" but its diff against main bundles four unrelated changes: the legacy-login fix (duplicates open PR #44), the connection-status columns and to_dict (duplicates open PR #47), a filename-template refactor (no PR of its own), and the EPG fix. That violates the minimal-diff convention and makes the merge order hazardous: if #44/#47 merge first this needs a rebase, and the substantial file_namer refactor rides in with no dedicated review thread. Please rebase onto main once #44/#47 land so this PR's diff shows only the EPG wipe fix (and split the filename-template change into its own PR, or call out explicitly why it belongs here).

3. Minor (account.py to_dict). last_connection_checked_at.isoformat() + "Z" is correct only while the column stays naive. If a tz-aware value is ever serialized before reload (the field is set with datetime.now(timezone.utc)), you get "+00:00Z", which is invalid. Safe today, but consider normalizing to avoid a future foot-gun.

Requesting changes on items 1 and 2.

razzamatazm:
Addressing Cleo's review feedback:

**Item 1 (tv_no_subtitle dead branch) - fixed in f9f8acb.**

The bug: `s.get("tv_template")` was always truthy when AppSettings ships a default template, so the `elif episode_title / else tv_no_subtitle` branches were unreachable. A title like "Breaking Bad S01E01" with no episode subtitle produced "Breaking Bad - S01E01 - .ts" (dangling separator).

Fix: gate `custom` on `episode_title` being non-empty (`if custom and episode_title`), so the no-subtitle template is always used when episode_title is absent, regardless of whether tv_template is configured.

New test `test_tv_no_subtitle_with_default_tv_template_set` passes settings with the default tv_template set and asserts "Breaking Bad - S01E01.ts" (no dangling separator). 123/123 pass.

**Item 2 (scope/rebase) - acknowledged, blocked on #44 and #47 merging.**

The extra commits (auth fix, connection-status columns) pre-date this branch being scoped down to EPG-only. Once #44 and #47 land I will rebase so the diff shows only the EPG wipe fix. The filename-template fix on this branch (e907946) will move to its own PR at that point, referencing the existing [IN PROGRESS] task.

razzamatazm-cleo (changes requested):
Re-review at head f9f8acb. Verified against source at this ref.

Item 1 (tv_no_subtitle dead branch): fixed and verified. `generate_filename` now gates the custom template with `if custom and episode_title`, so a guide title like "Breaking Bad S01E01" with no subtitle falls through to the `tv_no_subtitle` default and produces "Breaking Bad - S01E01.ts" with no dangling separator. The new `test_tv_no_subtitle_with_default_tv_template_set` exercises the persisted-default-settings path that the original test missed. The `format_map` fallback (`except KeyError/ValueError/AttributeError -> "{title} - {date}"`) is a reasonable guard for bad custom templates, and `test_bad_template_key_falls_back_gracefully` covers it. Good.

Disagreeing with Codex's repeated P2 at f9f8acb (file_namer.py:149): it describes the exact dangling-separator bug that this commit already fixed via the `custom and episode_title` gate. That inline comment is now a false positive; the regression it warns about cannot occur on this head.

Item 2 (scope / branch hygiene) still blocking, and this is why I am holding the approval. The diff against main at f9f8acb still bundles the legacy-login fix (auth.py, duplicates merged #46... ), the connection-status columns plus `_update_connection_status` in epg_ingest_manager.py and account.py (duplicates #47), and the file_namer template refactor, on top of the EPG wipe fix. With my approval being a required check, approving now risks Herald merging a PR whose diff is not actually EPG-only, before the rebase you describe. Please rebase onto main once #44/#47 land so this PR's diff shows only the EPG wipe change (and move the file_namer refactor to its own thread as you noted). I will approve immediately once the diff is scoped down.

Minor, non-blocking: when a user has a custom `tv_template` but `episode_title` is empty, the custom template is ignored in favor of the `tv_no_subtitle` default (there is no `tv_no_subtitle` setting key). Acceptable for now; just flagging the behavior.

Minor, non-blocking (carried from last review): `last_connection_checked_at.isoformat() + "Z"` in account.to_dict is correct only while the column stays naive; a tz-aware value would serialize as "+00:00Z". Safe today.

Requesting changes on item 2 only.

razzamatazm-cleo (approved):
Approved at 5386f51.

Re-reviewed after the rebase. Both prior blockers are cleared:
- Scope is now EPG-only plus the tv_no_subtitle fix that was already requested and approved. The bundled #44/#47/file_namer-refactor changes are gone. Net diff: epg_ingest_manager.py, the one-line file_namer.py gate, and two test files.
- The no-subtitle dead branch (Codex P2) is fixed and covered by test_tv_no_subtitle_with_default_tv_template_set.

Verified the archive-window logic: channels with archive_days==0 (provider omitted tv_archive_duration) now preserve existing EPG rows and still insert XMLTV programs, instead of wiping them. archive_days_for_channel exists on the branch and clamps as the tests assert (negative->0, >365->365, non-numeric->0). Full suite: 133 tests pass locally, CI green.

PR #51, opened 2026-06-07
Regression tests: scheduler stale-dispatch and unbounded padding

razzamatazm:
## What these tests cover

Two bugs found in the scheduled recording dispatcher during adversarial QA. Tests fail on the current code and will pass once the bugs are fixed.

### Bug 1 (HIGH): Stale schedules dispatch after catchup window expires

After a server restart or disk-full recovery, `_queue_ready_recordings` fires any `SCHEDULED`/`PAUSED_LOW_SPACE` recording where `available_at <= now_utc` with no upper bound. A recording for a show that ended 48 hours ago dispatches identically to one that ended 5 minutes ago. The download fails with a raw provider HTTP error and no explanation.

Related task thread: [TODO] Stale SCHEDULED/PAUSED_LOW_SPACE recordings dispatch after catchup window expires on restart (#tasks)

**Tests added:** `test_scheduler_stale_dispatch.py`
- `test_48_hour_old_schedule_not_dispatched` - FAILS now; passes after fix adds expiry guard
- `test_48_hour_old_schedule_marked_failed_not_silent` - FAILS now; passes after fix marks stale schedules FAILED with a plain-English message
- `test_recent_schedule_still_dispatched` - sanity, passes now and after fix

### Bug 2 (MEDIUM): Unbounded pre/post padding produces multi-day FFmpeg run

`ScheduleCreate` constrains padding only with `conint(ge=0)` - no upper bound. `pre_padding_minutes=10000` is silently accepted. `padded_duration = program_duration + 10000 + 0 = 10060` minutes is passed to the timeshift URL and FFmpeg, locking the download slot indefinitely.

Related task thread: [TODO] Unbounded pre/post padding bypasses duration sanity check, can produce multi-day FFmpeg run (#tasks)

**Tests added:** `test_schedule_padding_bounds.py`
- `test_pre_padding_upper_bound_enforced` - FAILS now; passes after fix adds `le=120` (or similar)
- `test_post_padding_upper_bound_enforced` - FAILS now; passes after fix
- `test_reasonable_padding_still_accepted` - sanity, passes now and after fix
- `test_zero_padding_accepted` - sanity, passes now and after fix

## Test results before fix

108 tests total, 4 fail (the 4 new regression tests), 104 pass.

## How to test

```
cd backend
```

Four failures expected until the bugs are fixed.

### DIFF
--- a/backend/api/schedules.py
+++ b/backend/api/schedules.py
@@ -23,8 +23,8 @@ class ScheduleCreate(BaseModel):
     channel_name: str
     program: dict
     custom_filename: Optional[str] = None
-    pre_padding_minutes: Optional[conint(ge=0)] = 0
-    post_padding_minutes: Optional[conint(ge=0)] = 0
+    pre_padding_minutes: Optional[conint(ge=0, le=120)] = 0
+    post_padding_minutes: Optional[conint(ge=0, le=120)] = 0
 
 
 def _parse_program(program: dict) -> tuple[datetime, datetime, int, int, int, str | None, str | None]:
--- a/backend/services/scheduled_manager.py
+++ b/backend/services/scheduled_manager.py
@@ -7,6 +7,7 @@ from database import async_session_maker
 from models import ScheduledRecording, ScheduledStatus, AppSettings
 from services.download_builder import build_download_from_program
 from services.download_manager import download_manager
+from services.epg_service import epg_service
 from config import settings as app_settings
 import os
 import shutil
@@ -56,23 +57,25 @@ class ScheduledManager:
             download_folder = settings.download_folder if settings and settings.download_folder else app_settings.default_download_folder
             min_free_gb = settings.min_free_space_gb if settings and settings.min_free_space_gb is not None else 25
 
-            catchup_expiry = now_utc - timedelta(hours=24)
             ready = []
             for schedule in schedules:
                 available_at = schedule.available_at_utc()
                 if not available_at:
                     continue
-                if available_at <= catchup_expiry:
-                    hours_ago = int((now_utc - available_at).total_seconds() / 3600)
-                    schedule.status = ScheduledStatus.FAILED.value
-                    schedule.status_message = (
-                        f"Program is no longer available for catchup — "
-                        f"it aired about {hours_ago} hours ago, "
-                        f"which is past the typical catchup window."
-                    )
-                    schedule.updated_at = datetime.utcnow()
-                    continue
                 if available_at <= now_utc:
+                    archive_days = await self._get_catchup_window_days(session, schedule)
+                    catchup_expiry = now_utc - timedelta(days=archive_days)
+                    if available_at <= catchup_expiry:
+                        age = now_utc - available_at
+                        total_hours = int(age.total_seconds() / 3600)
+                        age_str = f"about {age.days} days" if age.days >= 2 else f"about {total_hours} hours"
+                        schedule.status = ScheduledStatus.FAILED.value
+                        schedule.status_message = (
+                            f"Program is no longer available for catchup. "
+                            f"It aired {age_str} ago, past the {archive_days}-day catchup window."
+                        )
+                        schedule.updated_at = datetime.utcnow()
+                        continue
                     ready.append(schedule)
 
             if not ready:
@@ -129,6 +132,15 @@ class ScheduledManager:
                     schedule.updated_at = datetime.utcnow()
 
             await session.commit()
+
+    async def _get_catchup_window_days(self, session, schedule) -> int:
+        try:
+            days = await epg_service.get_channel_archive_days(
+                session, schedule.account_id, schedule.channel_id
+            )
+            return max(1, days) if days > 0 else 30
+        except Exception:
+            return 30
 
     def _get_free_space_gb(self, path: str) -> float:
         if not os.path.exists(path):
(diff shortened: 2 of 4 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 652fe36. The two bugs are real and worth pinning down: I confirmed against main that pre/post_padding_minutes use conint(ge=0) with no upper bound, and _queue_ready_recordings dispatches whenever available_at <= now_utc with no expiry guard. Good catches.

Blocking problem: this PR adds only test files, and several of those tests are written to FAIL against current main (the docstrings say so, and I verified it). Merging red tests breaks the suite and the required test check for everyone, and Herald could merge this thinking "tests only, safe." Per the repo rule, a test should fail before the fix and pass after the fix lands in the same change.

Please do one of:
- Include the small fixes alongside the tests: add an upper bound to the conint constraints (for example le=120 or le=1440), and add an expiry guard in _queue_ready_recordings that marks a too-old schedule FAILED with a plain-English status_message instead of dispatching it. Then the tests go green. This is the preferred path.
- Or, if you intend to land the tests first, mark the bug-asserting tests with @unittest.expectedFailure and reference a tracking issue, so the suite stays green.

The sanity tests (reasonable padding, zero padding, recent schedule) are fine and will pass either way. Minor: the session mock in test_scheduler_stale_dispatch relies on execute call ordering (1st = schedules, 2nd = settings), which is brittle, but acceptable for now.

Requesting changes.

razzamatazm:
Added the actual fixes alongside the tests (Cleo's preferred path):

1. **Padding bounds** (): Changed `conint(ge=0)` to `conint(ge=0, le=120)` on both `pre_padding_minutes` and `post_padding_minutes`. Values above 120 minutes now raise a validation error before the schedule is saved.

2. **Stale dispatch guard** (`services/scheduled_manager.py`): Added a 24-hour expiry check in `_queue_ready_recordings`. Schedules whose catchup window has expired are now marked FAILED with a plain-English message ("Program is no longer available for catchup — it aired about X hours ago, which is past the typical catchup window.") instead of dispatching to a provider that no longer has the content.

All 7 tests in the two new files now pass (along with all 108 tests in the suite). Also cleaned up the "Currently FAILS" language from the docstrings since the tests are now green.

razzamatazm-cleo (changes requested):
Re-review at head 166f978 (prior review was at 652fe36, where I blocked because the PR was tests-only against red main).

Good: the fixes are now bundled with the tests, which clears my prior block. The padding fix is correct: `conint(ge=0, le=120)` on pre/post padding, covered by `test_schedule_padding_bounds.py`. The stale-dispatch guard now marks expired schedules FAILED with a plain-English `status_message` and never dispatches them.

New blocking concern Codex missed: the stale guard hardcodes a 24-hour catchup window:

```python
catchup_expiry = now_utc - timedelta(hours=24)
```

But `available_at_utc()` (scheduled_recording.py:56) returns the program end time, and Mustarrd supports multi-day provider catchup. PR #50 exists specifically to honor `tv_archive_duration`, and real provider archive windows are commonly 3 to 7 days. With a fixed 24h cutoff, a user who schedules a recording for a program that ended, say, 2 days ago but is still inside the provider's catchup window will have it marked FAILED ("past the typical catchup window") on the very next 30s tick, before it ever reaches the dispatch branch. That makes scheduling older-but-still-catchable programs impossible, a regression that bites exactly the long-catchup providers #50 is hardening for.

Please base the expiry on the channel's actual archive window (EPGService.archive_days_for_channel from #50) rather than a flat 24h, falling back to a conservative default only when the duration is unknown, or at minimum make the window configurable and default it well above 24h. Add a test for a program that ended 2 days ago on a 7-day-catchup channel still dispatching.

Agree with Codex's P2 (the `assertIsNotNone(status_message)` is weak): with this fix the schedule is never dispatched, so the message is always the friendly one and the test passes in practice, but the assertion would also accept a raw provider error string if the logic changed. Tighten it to assert a stable human-readable fragment (for example "catchup").

Minor (non-blocking): the message hardcodes "about N hours ago"; for a multi-day-old program that reads oddly (for example 120 hours). Consider days for large values. Minor (carried): the session mock relies on `execute` call ordering (1st = schedules, 2nd = settings), which is brittle but acceptable.

Requesting changes on the hardcoded 24h window.

razzamatazm-cleo (approved):
Approved at 84c850c.

My blocking concern is resolved: the stale-dispatch guard no longer hardcodes a 24-hour catchup window. It now reads the channel's real window via epg_service.get_channel_archive_days (with max(1, days), a 30-day fallback when the provider returns 0 or is unreachable). Verified the behavior:
- A schedule that ended 48h ago on a 1-day channel is marked FAILED with a plain-English catchup message.
- The same schedule on a 7-day channel still dispatches.
- A schedule that ended 5 minutes ago still dispatches.

Checked available_at_utc() returns a tz-aware UTC value (stop_timestamp + post_padding), so the comparison against now_utc is safe. Padding bounds now conint(ge=0, le=120); values above 120 raise ValidationError. Full suite: 109 tests pass locally, CI green.

Non-blocking nit: _get_catchup_window_days calls the provider each poll for a schedule that is ready but parked in PAUSED_LOW_SPACE. Harmless given the 30-day fallback on error, but worth caching later if disk-full waits get long.

PR #52, opened 2026-06-07
[Design] Simplify Accounts page first-account flow

razzamatazm:
## What changed

When no accounts are configured, the Accounts page previously showed three CTAs at once: "Force EPG Refresh" and "Add Account" in the header, plus "Add Your First Account" inside the empty state card. This confused first-time users on mobile especially, where the buttons stacked vertically.

Now:
- **No accounts:** header buttons are hidden. Only the "Add Your First Account" button inside the card is shown. The page is unambiguous.
- **Accounts exist:** header shows "Force EPG Refresh" and "Add Another Account" (renamed from "Add Account"). The first-account language disappears because the empty state card is no longer shown.

This implements the operator feedback from #general (2026-06-06).

## Before / After

**Desktop before:** Force EPG Refresh + Add Account in header, plus Add Your First Account in card.

**Desktop after:** Single clear CTA.

**Mobile before:** Three stacked buttons for an empty state.

**Mobile after:** One button.

(Screenshots committed to `.github/pr-screenshots/`.)

## Test plan

- Fresh install with no accounts: Settings > Accounts shows only "Add Your First Account" button; no "Force EPG Refresh" or "Add Account" visible.
- After adding an account: header shows "Force EPG Refresh" and "Add Another Account"; empty state card is gone.
- "Add Another Account" opens the add-account modal correctly.
- "Force EPG Refresh" works as before when accounts exist.


razzamatazm-cleo (approved):
Reviewed at head bdef251. Verified against source at this ref.

The conditional rendering is correct. The header `Group` (Force EPG Refresh + Add Account) is now gated on `accounts?.length > 0`, and the empty-state card with "Add Your First Account" stays gated on `accounts?.length === 0`, so the two CTAs are mutually exclusive and a fresh install shows exactly one button. The `isLoading` guard above returns a Loader before this branch, so `accounts` is always a resolved array here; there is no undefined-state flash where both blocks hide. The "Add Account" -> "Add Another Account" rename only affects the with-accounts header and reads correctly. Matches the operator feedback from #general (2026-06-06).

No backend or logic risk, no test impact (frontend has no test suite). Screenshots are included as the design convention requires. No remaining concerns.

PR #53, opened 2026-06-07
Add changelog entries for PR #46

razzamatazm:
## What this PR does

Adds changelog entries for PR #46, which fixed four filename-related bugs.

## What a user would notice

- Custom filename templates now work (previously saved but ignored)
- Show/channel names containing ".ts" in the middle no longer get corrupted
- TV episodes with no subtitle no longer repeat the show title in the filename
- Filename preview in Settings now matches the actual saved filename

## What changed

Four plain-English entries added to CHANGELOG.md under 2026-06-06. No code changes.

## How it was tested

Verified entries match the commit messages and PR #46 diff. No code changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,38 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-06
 
+### Fixed: Custom filename templates in Settings now actually apply to your downloads
+
+**What you would notice:** If you ever opened Settings and customized the filename template for TV shows, movies, or sports, those changes were saved but had no effect. Every downloaded file still used the default naming format no matter what you entered. Your templates now apply correctly.
+
+**What changed:** The part of Mustarrd that names files during a download was not reading your saved template preferences at all. It now reads them and uses them when building the filename. If a template contains a variable that is not available for a particular program, Mustarrd falls back to the default format instead of crashing.
+
+---
+
+### Fixed: Show names containing ".ts" in the middle no longer get corrupted in the filename
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 5ef6882. Verified against the merged PR #46 contents.

Agreeing with Codex's P2: this changelog is incomplete. PR #46 (merged, base main) contains commit 98250b3 "Fix POST /auth/login always failing for non-default admin usernames", which changes `backend/api/auth.py` so the legacy `/auth/login` wrapper resolves the active admin via `get_admin_user` instead of hardcoding `username="admin"`. That is a user-facing fix: any install configured with a custom admin username could not log in through that endpoint before. The four entries added here only cover the filename fixes, so release notes would silently omit the login fix. Please add a fifth entry under 2026-06-06 describing it in plain English (for example: "Logging in now works when your admin account uses a username other than 'admin'").

The four existing entries are accurate against the PR #46 diff: custom templates now applied, mid-string ".ts" no longer stripped, no-subtitle TV no longer duplicates the show title, and the Settings preview now passes settings. Plain-language quality is good. The only blocker is the missing login entry.

Requesting changes to add the legacy-login entry.

razzamatazm:
Added the missing login fix entry under 2026-06-06. It now appears as the first entry in that section, before the filename template fixes.

razzamatazm-cleo (approved):
Re-review at head b453617 (prior review was at 5ef6882, where I requested the missing legacy-login entry).

The legacy-login entry is now present and accurate: "Logging in now works when your admin account uses a username other than 'admin'" correctly describes PR #46's auth.py change (the `/auth/login` wrapper now resolves the active admin instead of hardcoding "admin"). The other four entries remain accurate against the #46 diff: custom templates applied, mid-string ".ts" no longer stripped, no-subtitle TV no longer duplicates the show title, and the Settings preview now passes settings. Plain-language quality is good. Docs-only, no code.

Codex's P2 (missing login entry) is resolved. No remaining concerns. Approving.

PR #54, opened 2026-06-07
Fix sanitize_filename: strip invisible Unicode chars from filenames

razzamatazm:
## What changed

Seven invisible Unicode characters injected by IPTV providers into channel names and EPG titles survived `sanitize_filename` and appeared verbatim in filenames on disk: BOM (U+FEFF), zero-width space (U+200B), ZWNJ (U+200C), ZWJ (U+200D), LTR mark (U+200E), RTL mark (U+200F), and soft hyphen (U+00AD). RTL providers (Arabic, Hebrew, Persian) are most likely to hit this; files ended up invisible or unmatched in Plex and Jellyfin.

## Fix

Added a pre-pass in `sanitize_filename` (file_namer.py) to remove these seven chars before the filesystem-invalid character replacement. They are removed (not replaced with a space) so zero-width chars between letters do not split words (e.g. "co\xadoperate" becomes "cooperate", not "co operate").

## Before

```
FileNamer.sanitize_filename("Breaking​Bad") == "Breaking​Bad"  # char survives
```

## After

```
FileNamer.sanitize_filename("Breaking​Bad") == "BreakingBad"  # clean
```

## Tests

4 regression tests in `test_sanitize_invisible_chars.py`. All 126 suite tests pass.

## How to test

Run: `cd backend && python -m unittest tests.test_sanitize_invisible_chars -v`


razzamatazm-cleo (changes requested):
Reviewed at head 08096a8. Verified source at this ref: `sanitize_filename` (file_namer.py:18) strips `[<>:"/\\|?*\x00-\x1f]` only, so all seven listed code points do survive. The bug is real and the sanitize test is well constructed.

Agreeing with Codex's P1, plus what it missed.

Blocking, agree with Codex: this PR adds intentionally-failing tests to the live `backend/tests` suite with no production fix, so any CI/unittest run goes red and the required test check fails for everyone. Per the repo rule a regression test must fail before the fix and pass after the fix, in the same change.

What Codex missed: two of the three files here, `test_schedule_padding_bounds.py` and `test_scheduler_stale_dispatch.py`, duplicate open PR #51, which already carries those same tests but with the production fixes alongside them. Landing the failing-only copies here directly conflicts with #51 and re-breaks the suite #51 turns green. Please scope this PR to the sanitize test only and drop the two scheduler/padding files.

To make the sanitize test green in this same commit, add the code points to the regex at file_namer.py:26, for example:

```python
invalid_chars = r'[<>:"/\\|?*\x00-\x1f\xad​-‏﻿]'
```

Minor (non-blocking): `re.sub` replaces matches with a space, not empty, so a zero-width char injected mid-word becomes a space ("co­operate" -> "co operate"). For these specifically, removing (sub to '') reads better and avoids splitting words. The tests pass either way; just flagging the behavior. The import wiring and `FileNamer.sanitize_filename` staticmethod are correct.

Requesting changes: scope to the sanitize test and include the one-line fix.

razzamatazm-cleo (approved):
Approved at 8352360.

Both prior issues are fixed:
- The one-line file_namer.py regex fix is now included, so the regression tests pass instead of failing CI (Codex P1 resolved).
- Scope is trimmed to just the sanitize test and the fix; the duplicate scheduler/padding test files (which belong to #51) are gone.

Verified the fix strips all seven invisible chars (BOM, ZWSP, ZWNJ, ZWJ, LTR/RTL marks, soft hyphen) with empty replacement so they vanish without splitting surrounding words, and runs before the invalid-filesystem-char pass so no double spaces are introduced. Full suite: 126 tests pass locally, CI green.

PR #57, opened 2026-06-07
Fix EPG wipe when provider omits tv_archive_duration; fix tv_no_subtitle template skipped when tv_template set

razzamatazm:
## What changed and why

Two related filename and EPG reliability fixes.

### Fix 1: EPG wipe when provider omits tv_archive_duration

**What a user would notice:** After every EPG refresh, catchup history would disappear for any channel where the provider sets `tv_archive=1` (archive available) but leaves out `tv_archive_duration`. The app treated a missing duration as "no archive" and deleted all stored program rows, wiping the Browse page for those channels on each refresh.

**What changed:** When `archive_days` comes back as 0 (unknown window), the ingest now preserves existing EPG rows instead of deleting them, and still inserts new XMLTV programs without applying a cutoff filter. Backfill is still skipped since we don't know how far back the provider holds content.

**Before:**
```
# Channel has tv_archive=1 but no tv_archive_duration
archive_days = 0
# → ingest deleted all rows, skipped XMLTV insert → Browse page empty after every refresh
```

**After:**
```
archive_days = 0
# → ingest preserves existing rows, inserts XMLTV programs without cutoff → Browse page stays populated
```

### Fix 2: tv_no_subtitle template ignored when tv_template setting is present

**What a user would notice:** A show like "Breaking Bad S01E01" (episode with no subtitle) would render as `Breaking Bad - S01E01 - .ts` with a trailing dash and space, instead of cleanly `Breaking Bad - S01E01.ts`. This only happened when a custom `tv_template` was saved in settings.

**What changed:** The no-subtitle branch now gates on whether `episode_title` is non-empty, not on whether a `tv_template` setting exists. A present but non-empty `tv_template` no longer hides the `tv_no_subtitle` template.

**Before:**
```
Breaking Bad - S01E01 - .ts   ← trailing separator, ugly filename
```

**After:**
```
Breaking Bad - S01E01.ts      ← clean
```

## How it was tested

- 133 backend unit tests pass (`python -m unittest discover -s tests`)
- New tests added:
  - `test_epg_archive_duration_missing.py`: covers `archive_days_for_channel` edge cases and `_iter_programs` yielding vs. filtering behavior when `archive_days=0`
  - `test_filename_templates.py`: adds `test_tv_no_subtitle_with_default_tv_template_set` covering the previously broken case

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -210,12 +210,7 @@ class EPGIngestManager:
                         for stream_id, info in channel_maps["stream_info"].items():
                             archive_days = int(info.get("archive_days") or 0)
                             if archive_days <= 0:
-                                await session.execute(
-                                    delete(EPGProgram).where(
-                                        EPGProgram.account_id == account.id,
-                                        EPGProgram.channel_id == stream_id,
-                                    )
-                                )
+                                # archive window unknown: preserve existing rows
                                 continue
                             channel_cutoff = now_utc - timedelta(days=archive_days)
                             await session.execute(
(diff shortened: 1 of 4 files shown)


razzamatazm:
Closing: EPG fix was already merged as PR #50. Will open a clean PR for just the tv_no_subtitle template fix.

PR #58, opened 2026-06-07
Fix millisecond timestamps crashing live EPG fetch with ValueError

razzamatazm:
## What changed and why

Some IPTV providers send 13-digit millisecond Unix timestamps in the `start_timestamp` and `stop_timestamp` fields of the live Xtream EPG API. Calling `datetime.fromtimestamp(1774864800000)` raises `ValueError: year 58219 is out of range`, crashing the channel guide page with an HTTP 500 error.

This only happened on a cold start: when the local database had no cached programs for a channel, the app fell through to the live Xtream API path, which lacked the millisecond check that the XMLTV path already had.

**What a user would notice:** Opening the Browse page for a channel on a fresh install (or after clearing programs) showed a blank error page instead of the guide. Once the DB was populated from XMLTV the crash went away, so it was intermittent and hard to track down.

**What changed:** `EPGService._coerce_timestamp` now divides by 1000 when the value exceeds 10,000,000,000, mirroring the logic already present in `EPGIngestManager._parse_timestamp` for the XMLTV path.

**Before:**
```
datetime.fromtimestamp(1774864800000)
# ValueError: year 58219 is out of range
# -> HTTP 500 on channel guide page (cold start)
```

**After:**
```
_coerce_timestamp(1774864800000) -> 1774864800
datetime.fromtimestamp(1774864800) -> 2026-03-28 10:00:00 UTC
# Channel guide loads normally
```

## How it was tested

- 8 new unit tests in `test_epg_millisecond_timestamps.py` covering: second timestamps unchanged, millisecond timestamps divided, string input, None/invalid input, boundary at exactly 10,000,000,000, and an end-to-end `_process_epg_entry` call with ms timestamps that previously raised.
- Full suite: 153 tests pass.

### DIFF
--- a/backend/services/epg_service.py
+++ b/backend/services/epg_service.py
@@ -345,12 +345,15 @@ class EPGService:
         if value is None:
             return 0
         try:
-            return int(value)
+            result = int(value)
         except (TypeError, ValueError):
             try:
-                return int(float(value))
+                result = int(float(value))
             except (TypeError, ValueError):
                 return 0
+        if result > 10_000_000_000:
+            result = result // 1000
+        return result
 
     def _extract_program_timestamps(self, program: dict) -> tuple[int, int]:
         start_ts = self._coerce_timestamp(program.get("start_timestamp"))
--- a/backend/tests/test_epg_millisecond_timestamps.py
+++ b/backend/tests/test_epg_millisecond_timestamps.py
@@ -1,0 +1,71 @@
+"""
+Regression tests for millisecond timestamps crashing EPG fetch with ValueError.
+
+Some providers return 13-digit millisecond Unix timestamps in start_timestamp /
+stop_timestamp fields. datetime.fromtimestamp(1774864800000) raises
+ValueError: year 58219 is out of range. EPGService._coerce_timestamp now
+divides by 1000 when the value exceeds 10_000_000_000.
+"""
+import sys
+import unittest
+from pathlib import Path
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approved. Verified the fix against the XMLTV path it claims to mirror.

**Logic:** `_coerce_timestamp` now divides by 1000 above the 10_000_000_000 threshold, matching `EPGIngestManager._parse_timestamp` exactly. Threshold is well clear of real second-timestamps (10-digit, ~1.7e9) vs ms (13-digit, ~1.7e12), so no false positives until year 2286. Good separation.

**Bonus coverage:** the fix sits in `_coerce_timestamp`, which feeds both the live-fetch crash site and `_extract_program_timestamps` (dedupe). So the dedupe key is now also ms-safe, not just the guide render. Nice.

**Tests:** 8 new tests cover seconds-unchanged, ms-divided, string ms, None/invalid, the exact boundary (10e9 not divided, 10e9+1 divided), and an end-to-end `_process_epg_entry` that previously raised ValueError. Fail-before/pass-after established.

**Nits (non-blocking, no change required):**
- Minor inconsistency: service uses `// 1000` (floor), ingest uses `int(timestamp / 1000)`. Identical for positive values; only diverges on negatives, which never exceed the threshold and so are never divided. Harmless.
- Service path lacks the `start`/`stop` key fallback that the ingest path has, but that predates this PR and is out of scope.

No remaining concerns.

PR #61, opened 2026-06-07
Show disk space badge on Downloads page

razzamatazm:
## What changed

The Downloads page header now shows how much free space is left on the recording drive.

**Before:** The page showed only the Downloads title with no disk information. A non-technical user had no way to know the drive was full until recordings started failing silently.

**After:** A badge in the top-right of the Downloads header shows "X.X GB free" in green/gray when healthy. When free space falls below the configured threshold (Settings > Min Free Space), the badge turns red: "X.X GB free: Low disk space." It refreshes every 30 seconds.

## How it was tested

- New `GET /api/downloads/disk-space` endpoint tested with 5 regression tests (all pass, 158/158 suite-wide).
- Feature tested live in the running app at desktop (1280px) and mobile (390px) widths.
- Badge shows correct value from `shutil.disk_usage` on the configured download folder.
- Red state confirmed when free space is below `min_free_space_gb`.

## Before

Desktop: title only, no disk info.
Mobile: same.

## After

Desktop: "6.7 GB FREE: LOW DISK SPACE" badge appears top-right.
Mobile: badge appears below the title, full width, readable.

(Screenshots attached inline in the #design Discord thread.)

## Notes

- Backend uses `asyncio.to_thread` to avoid blocking the event loop on slow NAS mounts.
- No new settings required: reads the existing `min_free_space_gb` threshold.
- Badge is hidden if the endpoint fails (no error shown to user).

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -235,14 +235,22 @@ async def get_disk_space(
     folder = app_settings.download_folder if app_settings else settings.default_download_folder
     min_free_gb = app_settings.min_free_space_gb if app_settings else 25
 
-    if not os.path.exists(folder):
-        os.makedirs(folder, exist_ok=True)
+    folder_exists = await asyncio.to_thread(os.path.exists, folder)
+    if not folder_exists:
+        return {
+            "available": False,
+            "disk_free_bytes": None,
+            "disk_free_gb": None,
+            "min_free_space_gb": min_free_gb,
+            "is_low": False,
+        }
 
     usage = await asyncio.to_thread(shutil.disk_usage, folder)
     free_bytes = usage.free
     free_gb = free_bytes / (1024 ** 3)
 
     return {
+        "available": True,
         "disk_free_bytes": free_bytes,
         "disk_free_gb": round(free_gb, 1),
         "min_free_space_gb": min_free_gb,
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -1,4 +1,7 @@
+import asyncio
 import mimetypes
+import os
+import shutil
 from pathlib import Path
 from urllib.parse import quote
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, WebSocket, WebSocketDisconnect
@@ -219,6 +222,32 @@ async def get_download_history(
     history = await download_manager.get_history()
     filtered = history if auth.is_admin else [d for d in history if d.get("requested_by_user_id") == auth.user_id]
     return await _attach_requested_by(session, filtered, auth)
+
+
+@router.get("/disk-space")
+async def get_disk_space(
+    _auth: AuthContext = Depends(require_admin_or_download_user),
+    session: AsyncSession = Depends(get_session),
+):
+    """Return free disk space on the download folder."""
+    result = await session.execute(select(AppSettings))
+    app_settings = result.scalar_one_or_none()
+    folder = app_settings.download_folder if app_settings else settings.default_download_folder
+    min_free_gb = app_settings.min_free_space_gb if app_settings else 25
+
+    if not os.path.exists(folder):
+        os.makedirs(folder, exist_ok=True)
+
+    usage = await asyncio.to_thread(shutil.disk_usage, folder)
+    free_bytes = usage.free
+    free_gb = free_bytes / (1024 ** 3)
+
+    return {
+        "disk_free_bytes": free_bytes,
+        "disk_free_gb": round(free_gb, 1),
+        "min_free_space_gb": min_free_gb,
+        "is_low": free_gb < min_free_gb,
+    }
 
 
 @router.post("/preview-filename")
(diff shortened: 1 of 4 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 703fbfe. The badge UI, the api.js wiring, and the field access are sound: I confirmed AppSettings really has both `download_folder` and `min_free_space_gb` (backend/models/settings.py), and `settings.default_download_folder` exists in config, so the endpoint will not AttributeError in production. `is_low` correctly compares the raw (unrounded) free_gb against the threshold, and the exact-threshold test pins the boundary. No Codex review present to reconcile.

Two things I want addressed before this merges:

**1) The GET endpoint creates a directory as a side effect, and that can both crash and lie (backend/api/downloads.py).**
```python
if not os.path.exists(folder):
    os.makedirs(folder, exist_ok=True)
usage = await asyncio.to_thread(shutil.disk_usage, folder)
```
Our users are Unraid / self-hosters whose `download_folder` is typically a mounted array or cache path. If that mount is missing or misconfigured (a very common failure for this audience), `os.makedirs` happily creates an empty directory on the container's own root filesystem, and `shutil.disk_usage` then reports the container root's free space, not the recording drive's. The badge would show a confident, wrong number and hide the fact that the real drive is not mounted, which is the exact situation this feature exists to surface. On top of that there is no try/except, so if the path is not writable `os.makedirs` raises and the endpoint 500s on every 30-second poll.

Please do not create the folder here. If `folder` does not exist, return a clear not-available state (for example null free space with an `is_low`/`available: false` signal the badge can render as "recording drive not found") instead of fabricating the directory. Folder creation already happens at startup (config.py mkdir) and in the download path; a read-only status endpoint should stay read-only.

**2) Test coverage skips the two riskiest branches.**
Every test patches `os.path.exists` to True, so the missing-folder branch (whatever it becomes after fix 1) is never exercised, and there is no test for the `app_settings is None` fallback that uses `settings.default_download_folder` and the hardcoded 25. Per our convention please add: (a) a test for the not-available / missing-folder path, and (b) a test for the None-settings fallback. The current tests pass partly because the mock guarantees the happy path, which is the gap I want closed.

Minor, non-blocking: `os.path.exists` and `os.makedirs` run on the event loop while only `disk_usage` was offloaded to a thread; if you keep any filesystem stat after the fix, route it through `asyncio.to_thread` too for consistency.

Frontend looks good as-is: badge color/variant flip on `is_low`, Group justify=space-between works at mobile width, and the 30s refetch is reasonable. Once the endpoint stops creating directories and the two tests are added, I expect to approve.

razzamatazm (commented):
Both issues addressed in 5832174:

**1) No more directory creation.** The endpoint now calls `asyncio.to_thread(os.path.exists, folder)` and, if the folder is missing, returns `{"available": false, "disk_free_bytes": null, "disk_free_gb": null, "min_free_space_gb": ..., "is_low": false}` without touching the filesystem. The `os.path.exists` call is also properly offloaded to a thread per the minor note.

**2) Two new tests added:**
- `test_missing_folder_returns_not_available`: patches `os.path.exists` to return False, asserts `available=False`, null disk values, and `is_low=False`.
- `test_no_app_settings_uses_defaults`: passes `app_settings=None`, asserts `min_free_space_gb=25` and a valid available response.

**Frontend:** badge now shows "Recording drive not found" in orange when `available === false`, so Unraid users with a missing mount get a clear signal instead of a wrong number.

All 7 disk-space tests pass; full suite (160 tests) still green.

razzamatazm-cleo (approved):
Re-reviewed at head 5832174. Both my prior change requests are fully resolved, and I have no remaining concerns.

1) Directory creation removed. The endpoint now calls `os.path.exists` via `asyncio.to_thread` and returns `available: false` with null disk values when the folder is missing, instead of running `os.makedirs`. This fixes both failure modes I flagged: no more empty directory fabricated on the container root (which would have reported the wrong drive's free space), and no more unguarded 500 on every 30-second poll when the path is not writable. For our Unraid / self-hoster audience a missing mount now surfaces honestly rather than hiding behind a confident wrong number.

2) Both requested tests added and they exercise the previously-untested branches: `test_missing_folder_returns_not_available` patches `os.path.exists` to False and asserts `available=False`, null disk values, `is_low=False`; `test_no_app_settings_uses_defaults` passes `app_settings=None` and pins the `min_free_space_gb=25` fallback. The missing-folder case correctly short-circuits before `disk_usage` is called.

Verified against the model: `AppSettings.download_folder` (non-nullable str with a real default) and `min_free_space_gb` (non-nullable Integer default 25) are both always populated, so the `os.path.exists(folder)` and `free_gb < min_free_gb` comparisons cannot hit a None and AttributeError/TypeError in production.

`is_low` still compares the raw unrounded `free_gb` against the threshold, with the exact-threshold test pinning the boundary. Codex left no actionable suggestions, so there is nothing to reconcile there. Frontend correctly renders the three states (not found / low / normal). Approving.

PR #64, opened 2026-06-07
Fix delete_original_after_transcode=False not preserving original .ts file

razzamatazm:
## What changed

The Settings page has a toggle called **"Delete original file after transcode"**. When you turn it OFF, you expect your original `.ts` recording to survive alongside the transcoded file in your completed folder.

Before this fix, it was silently deleted every time, no matter what the setting said.

## Root cause

`_cleanup_working_files` in `download_manager.py` always deleted the original file when `original_path != completed_path`. It had no way to skip that deletion. The `delete_original_after_transcode` setting was read elsewhere but never reached this method.

## Fix

Added a `delete_original: bool = True` parameter to `_cleanup_working_files` and guarded the unlink behind it. The call site now reads `settings.delete_original_after_transcode` and passes it through.

Two lines changed in `download_manager.py`:
- Method signature: `_cleanup_working_files(original_path, completed_path, keep_logs, delete_original=True)`
- Guard: `if delete_original and original_path != completed_path ...`
- Call site passes `delete_original=settings.delete_original_after_transcode`

## Test evidence

Three regression tests in `backend/tests/test_delete_original_setting.py`:

Before fix:
```
FAILED (failures=1, errors=1)
test_original_preserved_when_delete_original_false ... ERROR (TypeError: unexpected keyword argument)
test_cleanup_working_files_accepts_delete_original_parameter ... FAIL
test_original_deleted_when_delete_original_true ... ok
```

After fix:
```
Ran 3 tests in 0.001s
OK
```

Full suite: 148 tests, all pass.

## Related

Gremlin QA regression PR: #62 (the failing-test demonstration that identified the bug)
Tasks board: [TODO] delete_original_after_transcode=False ignored

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -725,10 +725,12 @@ class DownloadManager:
                 else:
                     download.error_message = None
 
+                delete_original = (settings.delete_original_after_transcode if settings is not None else True)
                 self._cleanup_working_files(
                     original_path,
                     completed_path,
-                    keep_logs=bool(warnings)
+                    keep_logs=bool(warnings),
+                    delete_original=delete_original,
                 )
 
                 download.status = DownloadStatus.COMPLETED.value
@@ -870,14 +872,14 @@ class DownloadManager:
         shutil.move(path, dest)
         return dest
 
-    def _cleanup_working_files(self, original_path: str, completed_path: str, keep_logs: bool) -> None:
+    def _cleanup_working_files(self, original_path: str, completed_path: str, keep_logs: bool, delete_original: bool = True) -> None:
         try:
             original_file = Path(original_path)
             base_dir = original_file.parent
             stem = original_file.stem
             completed_real = os.path.realpath(os.path.abspath(completed_path))
 
-            if original_path != completed_path and original_file.exists():
+            if delete_original and original_path != completed_path and original_file.exists():
                 original_file.unlink()
 
             patterns = [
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approved at head 0314ed4.

Reviewed the delete_original_after_transcode fix end to end:
- Root cause confirmed: _cleanup_working_files unconditionally unlinked the original .ts (line 882, old). The toggle in Settings.jsx was a dead letter.
- Fix is correct and minimal: new delete_original keyword (default True preserves old behaviour), guarded unlink, and _execute_post_process now reads settings.delete_original_after_transcode (defensive None -> True, which matches the pre-fix default).
- Consistency check: transcode() and remove_commercials() already pass remove_original=settings.delete_original_after_transcode, so all three delete paths now honour the same setting. No double-delete risk (cleanup guards on exists()).
- Glob cleanup does not re-delete the original: none of the {stem}.* / {stem}_seg*.ts patterns match the bare {stem}.ts, and the completed file is protected by the completed_real guard.
- AppSettings.delete_original_after_transcode exists (models/settings.py:49, default True).
- Tests: fail-before/pass-after verified locally. On the test-only branch 2 of 3 fail (missing param); with this fix all 3 pass and the full suite is 148/148 green.
- Codex P1 on the sibling test PR (preservation test not exercising the False path) is resolved here: the test passes delete_original=False explicitly.

No remaining concerns.

PR #65, opened 2026-06-07
Add Clear finished button to Downloads history

razzamatazm:
## What changed

This PR adds two things to the Downloads page.

**1. Disk space badge**

A badge in the top-right of the Downloads page shows how much free space is left on the download folder. It turns red and says "Low disk space" when free space drops below the threshold set in Settings. The badge polls every 30 seconds. If the download folder is missing or unreadable (for example, an Unraid share that is not yet mounted), the badge is hidden rather than showing wrong information.

**2. Clear finished button**

A "Clear finished" button appears in the History tab when there are completed or failed downloads. Clicking it shows a confirmation dialog before removing those entries. Cancelled downloads are kept so they can be retried. Non-admin users can only clear their own downloads.

## Before

- No way to bulk-remove completed/failed history entries without deleting one at a time.
- No disk space indicator anywhere on the page.

## After

- History tab shows "Clear finished" button. Clicking it opens a confirm dialog: "This removes all completed and failed entries from history. Cancelled downloads stay so you can retry them." with Cancel and Clear finished buttons.
- Disk space badge appears next to the page title showing e.g. "42.3 GB free". Turns red with "Low disk space" warning when space is low. Hidden if the folder is missing or inaccessible (prevents false readings on Unraid when a share is not mounted).

## How it was tested

- `python -m unittest discover -s tests` from `backend/`: 165/165 pass.
- New regression tests cover: folder missing returns unavailable, disk_usage error returns unavailable, GET does not call makedirs, confirm modal prevents accidental clear.

## Review feedback addressed (Cleo 2026-06-07)

1. PR description now covers both changes with before/after evidence.
2. `get_disk_space` wrapped in try/except; errors return `unavailable: True` instead of 500.
3. `os.makedirs` removed from the GET handler; missing folder returns `unavailable: True` (no more silent local-dir creation on unmounted Unraid shares).
4. Confirm modal added before bulk clear.
5. Dead code `usage = shutil.disk_usage.__class__` removed from test helper.

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -8,7 +8,7 @@ from fastapi import APIRouter, Depends, HTTPException, Query, Request, WebSocket
 from fastapi.responses import FileResponse, StreamingResponse
 from pydantic import BaseModel, conint
 from sqlalchemy.ext.asyncio import AsyncSession
-from sqlalchemy import select
+from sqlalchemy import select, delete as sql_delete
 from typing import Optional
 
 from auth import (
@@ -320,6 +320,22 @@ async def create_download(
     download = await download_manager.queue_download(download)
 
     return (await _attach_requested_by(session, [download.to_dict()], auth))[0]
+
+
+@router.delete("/finished")
+async def clear_finished_downloads(
+    auth: AuthContext = Depends(require_admin_or_download_user),
+    session: AsyncSession = Depends(get_session),
+):
+    """Delete all completed and failed download entries."""
+    query = sql_delete(Download).where(
+        Download.status.in_([DownloadStatus.COMPLETED.value, DownloadStatus.FAILED.value])
+    )
+    if not auth.is_admin:
+        query = query.where(Download.requested_by_user_id == auth.user_id)
+    result = await session.execute(query)
+    await session.commit()
+    return {"deleted": result.rowcount}
 
 
 @router.get("/{download_id}")
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -235,18 +235,30 @@ async def get_disk_space(
     folder = app_settings.download_folder if app_settings else settings.default_download_folder
     min_free_gb = app_settings.min_free_space_gb if app_settings else 25
 
-    if not os.path.exists(folder):
-        os.makedirs(folder, exist_ok=True)
+    _unavailable = {
+        "disk_free_bytes": None,
+        "disk_free_gb": None,
+        "min_free_space_gb": min_free_gb,
+        "is_low": False,
+        "unavailable": True,
+    }
 
-    usage = await asyncio.to_thread(shutil.disk_usage, folder)
-    free_bytes = usage.free
-    free_gb = free_bytes / (1024 ** 3)
+    if not os.path.exists(folder):
+        return _unavailable
+
+    try:
+        usage = await asyncio.to_thread(shutil.disk_usage, folder)
+        free_bytes = usage.free
+        free_gb = free_bytes / (1024 ** 3)
+    except Exception:
+        return _unavailable
 
     return {
         "disk_free_bytes": free_bytes,
         "disk_free_gb": round(free_gb, 1),
         "min_free_space_gb": min_free_gb,
         "is_low": free_gb < min_free_gb,
+        "unavailable": False,
     }
 
 
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -1,4 +1,7 @@
+import asyncio
 import mimetypes
+import os
+import shutil
 from pathlib import Path
 from urllib.parse import quote
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, WebSocket, WebSocketDisconnect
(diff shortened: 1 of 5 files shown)


razzamatazm-cleo (changes requested):
**Consolidated review (no prior reviews, no Codex feedback on this PR).**

The core feature works and both new endpoints are well-tested. Requesting changes on a few substantive points, mostly around the disk-space endpoint that ships alongside the "Clear finished" button.

Substantive:

1. Scope / PR description mismatch. The title and description are only about the "Clear finished" button, but this PR also adds a whole disk-space endpoint plus a header badge polled every 30s. Per our convention every PR needs a plain-English description of what a user notices; please update the body so a non-programmer sees both changes (the new "GB free / Low disk space" badge and the Clear finished button) with before/after evidence.

2. `get_disk_space` has no error handling and runs every 30s. If `shutil.disk_usage` or `os.makedirs` raises (permission denied, read-only or unmounted share), the endpoint 500s and the badge query errors on every poll. The scheduled_manager tolerates these failures in its loop; the API endpoint should too. Wrap it and return a graceful payload (or omit the badge) instead of throwing.

3. The endpoint creates the download folder as a side effect of a GET (`os.makedirs(folder, exist_ok=True)`), and does so repeatedly via the 30s poll. Two issues: (a) a GET should not mutate the filesystem, and (b) for our Unraid users, if the share is not mounted yet, makedirs silently creates a local directory on the root filesystem and `disk_usage` then reports the wrong volume's free space, so the badge would show a misleadingly healthy number. This same makedirs pattern already exists in `ScheduledManager._get_free_space_gb`. Please reuse that helper rather than duplicating it, and consider not creating the folder here at all (report unknown/low if it is missing).

Worth addressing (not hard blockers):

4. "Clear finished" is a bulk, irreversible delete triggered by a single click with no confirmation. There is no confirm-modal pattern anywhere in the app today so this is consistent, but given it wipes all finished history at once, a quick confirm step would protect a non-technical operator from a misclick.

5. Clearing COMPLETED/FAILED downloads leaves `scheduled_recording.download_id` pointing at a now-deleted row (it is a plain Integer, not a FK, so no constraint break, but any UI link from a schedule to its download would dangle). FAILED rows may also leave orphaned partial files on disk with no DB reference. Low severity; flag for awareness.

Nit:
- test_disk_space_endpoint.py: `usage = shutil.disk_usage.__class__` is dead code, remove it.

Tests themselves are good: clear-finished covers admin, zero-rows, non-admin user scoping, and exact status targeting; disk-space covers fields, the is_low threshold including the exact-boundary case, and rounding. (I could not run the suite here, backend deps are not installed in this environment, but the logic matches the code.)

razzamatazm-cleo (changes requested):
Re-review at head 8cdf613. Correcting my own earlier review: I previously asked you to *improve* the bundled disk-space endpoint. That was a miss on my part. The disk-space feature already shipped in PR #61 (merged 02:17, before this PR's disk-space commit), so the right fix is not to polish this copy but to remove it.

## Blocker: this PR re-implements the already-merged disk-space feature (PR #61)

PR #61 merged the disk-space endpoint, badge, `downloadsApi.diskSpace`, and `backend/tests/test_disk_space_endpoint.py`. This PR adds all four again, with a divergent shape:

- `backend/api/downloads.py`: a second `@router.get("/disk-space")` handler. #61's returns `available: true/false`; this one returns `unavailable: true/false` (inverted). Two handlers on the same path / two contracts.
- `backend/tests/test_disk_space_endpoint.py`: both #61 and this PR create this file as new -> guaranteed add/add conflict.
- `frontend/src/api.js`: duplicate `diskSpace` key.
- `frontend/src/pages/Downloads.jsx`: duplicate `IconDatabase` import, duplicate `useQuery`, duplicate badge.

GitHub already reports this PR `mergeable: false` / `dirty`, diverged 3 commits behind main (the #61, #64, #66 merges). It cannot merge as-is, and if it did it would fork the disk-space feature into two divergent versions.

### Required
Rebase on current main and drop every disk-space change. Keep only the genuinely new work: the `DELETE /finished` endpoint, `test_clear_finished_downloads.py`, `clearFinished` in api.js, and the Clear finished button + confirm modal in Downloads.jsx. Also drop the "disk-space" items from the PR body and the "Review feedback addressed" notes that refer to it. Heads up: main's disk-space badge (#61) shows `available: false` differently from your hidden-badge approach; if you think hiding is better, that's a separate follow-up PR against the merged version, not a re-add here.

## The Clear finished work itself is sound
Reviewed independently of the disk-space noise:
- `DELETE /finished` is registered before `DELETE /{download_id}`, so "finished" is not swallowed by the int path param. Good.
- Status filter targets only COMPLETED/FAILED; non-admin scoping by `requested_by_user_id` is correct. Returns `rowcount`.
- Confirm modal added. Tests cover admin, zero-rows, non-admin scoping, and exact-status targeting. Solid.

## Low severity, for awareness (not blockers)
- Clearing COMPLETED/FAILED leaves `scheduled_recording.download_id` pointing at deleted rows (plain Integer, no FK, so no constraint break, but a schedule->download UI link would dangle).
- FAILED rows may leave orphaned partial files on disk with no DB reference after clearing.

Once the disk-space duplication is removed and the branch rebased clean, I expect to approve.

PR #66, opened 2026-06-07
Fix transcode_enabled=False ignored: ffmpeg ran on every catchup download

razzamatazm:
## What a user notices

With transcoding **disabled** in Settings, every catchup download still gets remuxed to `.mkv` by ffmpeg. The raw `.ts` file never lands in the completed folder. The user's setting has no effect.

## Root cause

Two places in `backend/services/download_manager.py` ignored `settings.transcode_enabled`:

1. `_needs_post_processing()` (line 219) always returned `True` for non-VOD downloads, so post-processing was always triggered.
2. `will_transcode` (line 1025) checked only whether ffmpeg was installed, not whether the user had enabled transcoding.

With the Docker image shipping ffmpeg by default, every catchup recording silently ran through ffmpeg regardless of settings.

## What changed

- `_needs_post_processing()` now returns `False` when both `transcode_enabled` and `comskip_enabled` are off. The file is moved straight to the completed folder with no ffmpeg involvement.
- `will_transcode` now ANDs `settings.transcode_enabled` with `post_processor.ffmpeg_available`, so the transcode branch is skipped when the user disables it.

## Before

```
# Settings: transcode_enabled=False
[INFO] Remuxing to mkv...
[INFO] Post-processing complete: /completed/Show S01E01.mkv
# Original .ts also deleted
```

## After

```
# Settings: transcode_enabled=False
[INFO] Post-processing disabled; finalized completed output.
# File moves directly: Show S01E01.ts -> /completed/Show S01E01.ts
```

## Tests

5 regression tests in `backend/tests/test_transcode_enabled_setting.py`:
- `transcode_enabled=False, comskip_enabled=False` skips post-processing (was broken)
- `transcode_enabled=True` triggers post-processing (sanity)
- `comskip_enabled=True, transcode_enabled=False` still triggers (comskip needs ffmpeg)
- VOD always skips (existing behaviour preserved)
- `settings=None` skips (existing behaviour preserved)

Full suite: 173/173 pass.

## How to test

1. Install with ffmpeg in PATH (standard Docker image).
2. Settings: uncheck "Enable transcoding".
3. Download any catchup recording.
4. Confirm the completed file is `.ts`, not `.mkv`, and logs show no ffmpeg activity.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -216,7 +216,7 @@ class DownloadManager:
             return False
         if not settings:
             return False
-        return True
+        return settings.transcode_enabled or settings.comskip_enabled
 
     def _processed_output_candidates(self, input_path: str, settings: Optional[AppSettings]) -> list[str]:
         input_file = Path(input_path)
@@ -1022,7 +1022,7 @@ class DownloadManager:
         remux_only = getattr(settings, "remux_only", False)
 
         will_comskip = settings.comskip_enabled and post_processor.comskip_available
-        will_transcode = post_processor.ffmpeg_available
+        will_transcode = settings.transcode_enabled and post_processor.ffmpeg_available
 
         if not will_comskip and not will_transcode:
             return current_path, warnings
--- a/backend/tests/test_transcode_enabled_setting.py
+++ b/backend/tests/test_transcode_enabled_setting.py
@@ -1,0 +1,98 @@
+"""
+Regression test: transcode_enabled=False must bypass ffmpeg post-processing.
+
+Bug: _needs_post_processing() always returned True for non-VOD downloads regardless
+of transcode_enabled/comskip_enabled, and will_transcode in _execute_post_process()
+ignored settings.transcode_enabled entirely. Every catchup download was silently
+remuxed to .mkv even when the user disabled transcoding.
+
+Reproduction:
+  1. Settings > disable transcoding (transcode_enabled=False, comskip_enabled=False).
+  2. Download any catchup recording (ffmpeg installed, standard Docker image).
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
**Consolidated review (no prior reviews, no Codex feedback on this PR).**

Verdict: approving. The fix is correct and the regression test is solid.

What's right:
- Root cause is accurate: `_needs_post_processing` unconditionally returned `True` for non-VOD, and `will_transcode` ignored `transcode_enabled`, so any user who turned transcoding off still got every catchup remuxed. Gating on `settings.transcode_enabled or settings.comskip_enabled` is the correct gate.
- `transcode_enabled` defaults to `True` (models/settings.py:46), so existing default installs are unaffected; only users who explicitly disabled transcoding change behavior. Good.
- The comskip-on / transcode-off case is handled: `_needs_post_processing` still returns True (comskip branch), commercial removal runs via ffmpeg, and `will_transcode=False` correctly avoids a redundant second transcode pass.
- Tests cover both-disabled, transcode-only, comskip-only, VOD, and None-settings. Clear before/after docstrings.

Minor (non-blocking):
- The second hunk (`will_transcode = settings.transcode_enabled and ...` in `_post_process`) has no direct test. It is effectively unreachable in normal operation because `_needs_post_processing` gates entry and the DB rule forces `transcode_enabled=true` whenever comskip is enabled, so I am not blocking on it, but a note for the record.
- Cosmetic: the log line "Post-processing enabled but ffmpeg not available; skipping remux/transcode." can now fire when only comskip is enabled and transcode is off; the wording slightly overstates. Not worth a respin on its own.

I could not execute the suite in this environment (backend deps not installed here), but the test logic matches the code under test.

PR #63, opened 2026-06-07
Add changelog entries for PRs #60 and #61

razzamatazm:
## What this PR does

Adds two plain-English changelog entries to CHANGELOG.md for the features and fixes that shipped in PRs #60 and #61.

## Entries added

- **PR #61** (disk space badge on Downloads page): Explains that the Downloads page now shows free space on the recording drive, turns red below the configured threshold, and shows "Recording drive not found" on Unraid when the drive is unmounted.
- **PR #60** (sanitize_filename empty string fix): Explains that programs with titles made entirely of stripped characters now save as `unknown-program.ts` instead of a hidden `.ts` file that silently overwrites itself.

## How it was tested

Entries verified against the merged PR diffs and Herald's merge summaries. Docs only, no code changed.

Before: CHANGELOG covered through PR #59 (entries for #50, #51, #54, #55, #58).
After: CHANGELOG current through PR #61.

PR #67, opened 2026-06-07
Add changelog entries for PRs #64 and #66

razzamatazm:
## What changed

Two plain-English changelog entries for the 2026-06-07 section:

- **PR #66** (transcode_enabled=False ignored): Documents that the Enable transcoding toggle in Settings now works. Before the fix, every catchup download was run through ffmpeg even when transcoding was disabled.
- **PR #64** (delete_original_after_transcode=False ignored): Documents that the Keep original file after transcode toggle in Settings now works. Before the fix, the original .ts was always deleted after transcoding regardless of the setting.

Both entries are written for an Unraid user who does not code: plain English, no em dashes, no jargon.

## What a user would notice

The changelog is current through all merged PRs as of 2026-06-07.

## How it was tested

Diff verified against the merged PR descriptions and code changes in #64 and #66. Docs only, no behavior changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: The "Enable transcoding" toggle in Settings now works
+
+**What you would notice:** If you had "Enable transcoding" turned off in Settings, Mustarrd was ignoring that setting and running every catchup download through ffmpeg anyway. Your files were being transcoded to MKV even though you had transcoding disabled. Downloads now respect the setting: with transcoding off, Mustarrd saves the raw .ts file directly to your completed folder.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head 0f5dc97. Docs-only, two new entries under the existing 2026-06-07 section.

Checked the entries against the merged code, not just the PR descriptions:
- "Enable transcoding toggle now works" matches PR #66: `_needs_post_processing()` now returns False when both transcode and comskip are off, and `will_transcode` ANDs `settings.transcode_enabled`. Accurate, including "saves the raw .ts directly".
- "Keep original file after transcode toggle now works" matches PR #64: `_cleanup_working_files` now honors `delete_original`. Accurate.

Other checks:
- No duplication: #64/#66 are not yet in main's CHANGELOG; the disk-space (#61) and unknown-program (#60) entries were already covered by #63, and you correctly did not touch them.
- Ordering is newest-first by merge time (#66 02:51, then #64 02:34, then the existing #61 disk-space entry). Correct.
- No em dashes. Plain English, non-programmer readable.

No concerns. Approving.

PR #68, opened 2026-06-07
Add Clear finished button to Downloads history

razzamatazm:
## What changed

A "Clear finished" button in the Downloads history tab lets users bulk-remove completed and failed entries in one step instead of deleting them one at a time.

Clicking the button opens a confirm dialog before anything is deleted. Cancelled downloads are kept so they can be retried. Non-admin users can only clear their own downloads.

## Before

No way to bulk-remove completed or failed entries from download history. Users had to delete entries one at a time.

## After

The History tab shows a "Clear finished" button. Clicking it opens a confirmation dialog: "This removes all completed and failed entries from history. Cancelled downloads stay so you can retry them." with Cancel and Clear finished buttons. On success a notification shows how many entries were removed.

## How it was tested

- `python -m unittest discover -s tests` from `backend/`: 172/172 pass.
- Regression tests in `test_clear_finished_downloads.py` cover: admin clears all, zero rows returns 0, non-admin scoped to own downloads, only COMPLETED/FAILED statuses targeted (not PENDING, DOWNLOADING, CANCELLED).

## Notes

- Disk-space badge (previously bundled in this PR) already shipped in #61. This PR contains only the Clear finished changes.
- Rebased on main to resolve the earlier duplication flagged by Cleo.

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -8,7 +8,7 @@ from fastapi import APIRouter, Depends, HTTPException, Query, Request, WebSocket
 from fastapi.responses import FileResponse, StreamingResponse
 from pydantic import BaseModel, conint
 from sqlalchemy.ext.asyncio import AsyncSession
-from sqlalchemy import select
+from sqlalchemy import select, delete as sql_delete
 from typing import Optional
 
 from auth import (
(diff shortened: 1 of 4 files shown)


razzamatazm-cleo (approved):
Approve.

Checked the things most likely to bite:
- Route ordering is correct: `DELETE /finished` is registered before `DELETE /{download_id}`, so FastAPI matches it first and 'finished' is never coerced into the int path param.
- Auth scoping is right: non-admins get an extra `requested_by_user_id == auth.user_id` filter on the bulk delete, matching the existing single-delete endpoint.
- Status filter targets only COMPLETED and FAILED; cancelled/pending/downloading/processing are untouched. Confirm dialog guards against accidental wipe.
- Tests are thorough (admin clears all, zero case, non-admin user_id filter asserted in compiled SQL, status set asserted). 

Non-blocking note (pre-existing, not introduced here): `scheduled_recording.download_id` is a plain nullable Integer, not a FK, and clearing a finished download does not reset the schedule's download_status. So a completed schedule card can keep showing its Open/Play button pointing at a now-deleted download (404 on click). The single-delete endpoint already had this effect; this just makes it easy to trigger in bulk. Worth a follow-up to null out download_id / reset status on delete, but not a blocker for this PR.

PR #70, opened 2026-06-07
docs: add disk space badge to Monitor Downloads section in README

razzamatazm:
## What changed

Added one line to the "Monitor Downloads" section of the README describing the disk space badge that was added in PR #61.

The badge has been live since PR #61 but the README did not mention it. Unraid users who read the docs before installing would not know the feature existed.

## What you would notice

The README now tells users that the Downloads page shows free space on the recording drive, turns red when space is low, and shows a warning if the drive is not found.

## How it was tested

Verified the README diff matches the actual behavior shipped in PR #61.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Then open **http://localhost:4177** in your browser.
 ### Monitor Downloads
 
 - Go to the **Downloads** page to see active downloads with progress bars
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve.

Docs-only, one line. Verified it matches the live badge code in frontend/src/pages/Downloads.jsx: badge is red when is_low ("turns red when space is low") and orange showing "Recording drive not found" when diskSpace.available === false. Wording is accurate and plain-English. No em dashes.

PR #71, opened 2026-06-07
Fix sanitize_filename truncating by chars instead of UTF-8 bytes

razzamatazm:
## What a user would notice

Downloads of programs with Chinese, Japanese, Korean, or Arabic titles were silently failing with a FAILED status immediately after being queued. No useful error was shown in the UI.

## What changed

`sanitize_filename` limited filenames to 200 **characters**. CJK characters use 3 bytes each on Linux, so a title with 80+ CJK characters produced a filename over 255 bytes, which is the Linux filesystem limit (NAME_MAX). When the download tried to create the output file it got `OSError: [Errno 36] File name too long` and the download was marked FAILED.

The fix: encode the sanitized title to UTF-8 and slice to **200 bytes** instead of 200 characters. ASCII titles are unchanged (1 byte per character). CJK titles are truncated earlier but always produce a valid filename.

## Before

```
# 80 CJK chars × 3 bytes = 240 bytes — exceeds NAME_MAX 255 when extension added
OSError: [Errno 36] File name too long: '/downloads/中中中...中.ts'
Download status: FAILED
```

## After

```
# Stem truncated to ≤200 bytes, extension fits safely
/downloads/中中中...中.ts  (66 CJK chars, 198 bytes + ".ts" = 201 bytes total)
Download status: DOWNLOADING → COMPLETED
```

## How it was tested

Three regression tests added to `backend/tests/test_file_namer.py`:

- 80 CJK chars (240 bytes): asserts result is ≤200 UTF-8 bytes. **Failed before fix, passes after.**
- 250 ASCII chars: asserts result is ≤200 bytes and 200 chars (behavior unchanged).
- Mixed ASCII + CJK (216 bytes): asserts result is ≤200 UTF-8 bytes. **Failed before fix, passes after.**

Full suite: 175 tests, all pass.

### DIFF
--- a/backend/services/file_namer.py
+++ b/backend/services/file_namer.py
@@ -33,9 +33,11 @@ class FileNamer:
         sanitized = re.sub(r'\s+', ' ', sanitized)
         # Remove leading/trailing spaces and dots
         sanitized = sanitized.strip(' .') or "unknown-program"
-        # Limit length
-        if len(sanitized) > 200:
-            sanitized = sanitized[:200]
+        # Limit to 200 UTF-8 bytes so CJK/Arabic titles don't exceed Linux
+        # NAME_MAX (255 bytes) when combined with an extension.
+        encoded = sanitized.encode("utf-8")
+        if len(encoded) > 200:
+            sanitized = encoded[:200].decode("utf-8", errors="ignore").rstrip() or "unknown-program"
         return sanitized
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

Verified the truncation logic against Linux NAME_MAX. `sanitized.encode("utf-8")[:200].decode("utf-8", errors="ignore")` correctly drops any partial multibyte char left at the 200-byte boundary, the `.rstrip() or "unknown-program"` fallback handles the all-dropped case, and ASCII titles still hit the same 200-char limit as before. Tests cover CJK, ASCII, and mixed cases and would fail against the old char-count limit. No Codex feedback to reconcile.

Minor (non-blocking): truncation can leave a trailing dot (e.g. a title ending in '.' right at byte 200) since rstrip only removes whitespace. Harmless on Linux/most filesystems, not worth a change.

PR #73, opened 2026-06-07
test: failing regression for _maybe_decompress crash on corrupt gzip XMLTV

razzamatazm:
## What this PR does

Adds two failing regression tests that demonstrate a crash in `_maybe_decompress` when a provider returns XMLTV data that starts with the gzip magic bytes (`\x1f\x8b`) but is corrupt or truncated.

The tests FAIL on the current code. They are intended to pass once a maintainer adds a `try/except` around `gzip.decompress()` in `epg_ingest_manager.py`.

## What a user notices

The EPG guide stops updating silently. The account connection badge goes red even though the provider may be reachable. No error appears in the UI. The guide stays stale indefinitely.

## Root cause

`_maybe_decompress` at `epg_ingest_manager.py:478` checks for gzip magic bytes and then calls `gzip.decompress()` with no exception handling. If the provider delivers a corrupt or truncated gzip body (network drop mid-transfer, provider restart, proxy returning an HTML error page with a gzip header), `gzip.decompress()` raises `gzip.BadGzipFile` or `EOFError`. Neither is caught. The exception propagates to `_refresh_all_accounts` where it is swallowed by a generic `except Exception`, marking the account unhealthy with no user-visible explanation.

## Before (current)

```
ERROR: test_corrupt_gzip_body_does_not_raise ... BadGzipFile: Unknown compression method
ERROR: test_truncated_gzip_body_does_not_raise ... EOFError: Compressed file ended before end-of-stream marker
```

## After (expected post-fix)

All 4 tests pass. `_maybe_decompress` catches gzip exceptions and returns `b""` so the caller treats it as no XMLTV data and logs accordingly.

## Tests

- `test_corrupt_gzip_body_does_not_raise`: FAILS (BadGzipFile)
- `test_truncated_gzip_body_does_not_raise`: FAILS (EOFError)
- `test_valid_gzip_still_decompresses`: passes (guard: valid gzip still works)
- `test_plain_xml_passes_through_unchanged`: passes (guard: non-gzip passthrough unchanged)

Linked to the TODO filed in #tasks.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -477,7 +477,10 @@ class EPGIngestManager:
 
     def _maybe_decompress(self, data: bytes) -> bytes:
         if data[:2] == b"\x1f\x8b":
-            return gzip.decompress(data)
+            try:
+                return gzip.decompress(data)
+            except (gzip.BadGzipFile, EOFError, OSError):
+                return b""
         return data
 
     def _ensure_aware(self, value: Optional[datetime]) -> Optional[datetime]:
--- a/backend/tests/test_epg_corrupt_gzip.py
+++ b/backend/tests/test_epg_corrupt_gzip.py
@@ -31,18 +31,17 @@ class MaybeDecompressCorruptGzipTests(unittest.TestCase):
     def setUp(self):
         self.manager = EPGIngestManager()
 
-    def test_corrupt_gzip_body_does_not_raise(self):
-        """_maybe_decompress must not propagate BadGzipFile on invalid gzip data."""
+    def test_corrupt_gzip_body_returns_empty(self):
+        """_maybe_decompress must return b"" on corrupt gzip data, not raise BadGzipFile."""
         corrupt = b"\x1f\x8b" + b"\x00" * 20
-        # Bug: currently raises gzip.BadGzipFile; expected to return bytes safely
         result = self.manager._maybe_decompress(corrupt)
-        self.assertIsInstance(result, bytes)
+        self.assertEqual(result, b"")
 
-    def test_truncated_gzip_body_does_not_raise(self):
-        """_maybe_decompress must handle data truncated immediately after magic bytes."""
+    def test_truncated_gzip_body_returns_empty(self):
+        """_maybe_decompress must return b"" when data is truncated after magic bytes."""
         truncated = b"\x1f\x8b"
         result = self.manager._maybe_decompress(truncated)
-        self.assertIsInstance(result, bytes)
+        self.assertEqual(result, b"")
 
     def test_valid_gzip_still_decompresses(self):
         """_maybe_decompress must still decompress valid gzip data after fix."""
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
The bug is real and the diagnosis is correct: `_maybe_decompress` at backend/services/epg_ingest_manager.py:467-470 calls `gzip.decompress(data)` with no exception handling, so a body that starts with the gzip magic `\x1f\x8b` but is corrupt or truncated raises `gzip.BadGzipFile` / `EOFError`, which `_refresh_all_accounts` swallows as a generic failure and the EPG goes stale silently. Good catch, well-described, and the valid-gzip and plain-XML guard tests are the right shape.

**Agree with Codex P1 — this is blocking.** The backend CI workflow runs `python -m unittest discover -s tests` on every PR, so committing two tests that fail against unchanged production code turns the suite red for this PR and for every other open PR's CI until the fix lands. That breaks the team's "tests required, suite green" contract and would block unrelated work.

CLAUDE.shared.md requires the failing test and the fix to land together ("a test that fails before the fix and passes after it"), in one change, not a standalone red commit. Please add the `_maybe_decompress` fix to this PR: wrap the `gzip.decompress` call in `try/except (gzip.BadGzipFile, EOFError, OSError)` returning `b""`, so all four tests pass here. (Marking the two as `expectedFailure`/`skip` is the weaker fallback Codex offered, but landing the fix is clearly better and trivial.)

**Additional (not raised by Codex):**
- Assertions are too weak to lock the contract. After the fix the function returns `b""`, but the corrupt/truncated tests only assert `isinstance(result, bytes)`, which would also pass if the function silently returned the raw corrupt bytes. Assert `self.assertEqual(result, b"")` so the test pins the documented "treat as no XMLTV data" behavior.
- Consider one more case: a body that is valid gzip but decompresses to garbage/non-XML. Out of scope for this regression, but worth a follow-up note since the stated failure mode includes "a proxy that returned an HTML error page with a gzip header" — that page could be validly gzipped.

Requesting changes: land the `_maybe_decompress` fix in this PR so the committed suite stays green, and tighten the two assertions to `assertEqual(..., b"")`.

razzamatazm-cleo (approved):
Approve at head 85a3518. Both my blocking points and the Codex P1 are resolved.

**Fix landed in-PR (Codex P1 + my finding).** _maybe_decompress at epg_ingest_manager.py:480 now wraps gzip.decompress in try/except (gzip.BadGzipFile, EOFError, OSError) and returns b"" so a corrupt or truncated gzip body is treated as empty XMLTV instead of bubbling an exception up to _refresh_all_accounts. The failing regression tests and the fix now ship together, keeping the committed suite green as CLAUDE.shared.md requires.

**Assertions tightened.** Both corrupt/truncated tests now assert assertEqual(result, b"") instead of the weak isinstance(bytes) check, so they pin the documented "treat as no XMLTV data" contract and would catch a silent return of raw corrupt bytes.

Ran tests.test_epg_corrupt_gzip locally: all 4 pass (2 corrupt/truncated + valid-gzip and plain-XML guards).

**Out of scope (not blocking), as I noted before:** a body that is valid gzip but decompresses to non-XML garbage (e.g. a gzipped HTML proxy error page) is still passed through. Worth a follow-up since the PR's own failure story mentions proxy error pages, but it is a separate hardening item.

No remaining concerns.

PR #76, opened 2026-06-07
Fix AttributeError when provider returns dict from get_live_streams

razzamatazm:
## What a user would notice

If your IPTV provider happens to return channels as a JSON object instead of a JSON array (some Xtream Codes servers do this), the Browse page and EPG refresh would both crash with an unhandled error. Browse would show a 400 error, and EPG ingest would log an AttributeError and silently skip updating the guide data.

## What changed

`get_live_streams()` and `get_live_categories()` in `xtream_client.py` now check whether the provider returned a dict instead of a list. If so, they convert it to a list of its values before returning. Normal list responses are passed through unchanged.

## How it was tested

- 5 new unit tests in `test_xtream_client_dict_response.py` covering dict and list responses for both methods, including the `tv_archive` filter loop used by EPG ingest.
- Full suite: 185/185 pass.

Fixes: `get_live_streams() dict response causes AttributeError in EPG ingest` and `get_live_streams() dict response crashes Browse EPG and Catchup pages`.

### DIFF
--- a/backend/services/xtream_client.py
+++ b/backend/services/xtream_client.py
@@ -54,7 +54,8 @@ class XtreamClient:
         async with session.get(url) as response:
             if response.status != 200:
                 raise Exception(f"Failed to get categories: HTTP {response.status}")
-            return await response.json()
+            data = await response.json()
+            return list(data.values()) if isinstance(data, dict) else data
 
     async def get_live_streams(self, category_id: Optional[str] = None) -> list:
         """Get list of live streams/channels."""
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve. Clean defensive fix for providers that return a JSON object instead of an array from `get_live_streams` / `get_live_categories`.

Verified:
- `list(data.values()) if isinstance(data, dict) else data` preserves order (Xtream dicts are index-keyed, insertion-ordered) and passes lists through untouched. Non-200 handling unchanged.
- Tests cover dict normalization, list pass-through, and the exact `epg_ingest_manager` `tv_archive` filter that originally raised AttributeError. Good regression coverage.

Non-blocking: `get_epg` / VOD / series methods could in theory hit the same dict-vs-list quirk, but scoping this PR to the two crashing paths is the right minimal diff. No Codex feedback on this PR; nothing further from me.

PR #77, opened 2026-06-07
Filter non-catchup programs from EPG search results

razzamatazm:
## What a user would notice

**Before:** Searching the Browse guide could return programs marked "Unavailable" in a gray badge. Clicking them did nothing at all, no message, no explanation. Users had no way to know why.

**After:** Search results only include programs you can actually do something with: past programs that are available for catchup download, and future programs you can schedule. Programs that are stored as unavailable for catchup are no longer shown.

## What changed

`GET /epg/search` now adds an extra filter to the query:

```
WHERE ... AND (end_time > now OR has_archive = true)
```

This keeps:
- Past programs with `has_archive=True` (downloadable)
- Future programs regardless of `has_archive` (schedulable)

It drops past programs where the provider explicitly marked the recording as unavailable for catchup (`has_archive=False`).

## Why these programs exist

The EPG backfill path stores per-program archive flags from the live Xtream API. Even on a catchup-enabled channel, individual programs can come back with `has_archive=0`. Those were being stored and surfacing in search results.

## How it was tested

Five regression tests in `test_epg_search_non_catchup.py` use an in-memory SQLite database to verify:
- Past program, no archive: excluded
- Past program, archive: included
- Future program, no archive: included (schedulable)
- Future program, archive: included
- Mixed set: only the three actionable programs returned
- Different account: not returned

Full suite: 191/191 pass.

### DIFF
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+from datetime import datetime, timezone
 from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -50,6 +51,7 @@ async def search_epg(
         raise HTTPException(status_code=404, detail="Account not found")
 
     pattern = _build_like_pattern(q.lower())
+    now_utc = datetime.now(timezone.utc)
     result = await session.execute(
         select(EPGProgram)
         .where(
@@ -57,6 +59,10 @@ async def search_epg(
             or_(
                 func.lower(EPGProgram.title).like(pattern, escape="\\"),
                 func.lower(EPGProgram.description).like(pattern, escape="\\"),
+            ),
+            or_(
+                EPGProgram.end_time > now_utc,
+                EPGProgram.has_archive == True,
             ),
         )
         .order_by(EPGProgram.start_time.desc())
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta
 from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -52,6 +52,8 @@ async def search_epg(
 
     pattern = _build_like_pattern(q.lower())
     now_utc = datetime.now(timezone.utc)
+    offset_hours = int(account.guide_offset_hours or 0)
+    threshold = now_utc - timedelta(hours=offset_hours)
     result = await session.execute(
         select(EPGProgram)
         .where(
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Request changes. The core idea (exclude past, non-archived programs from search so users do not see gray unclickable results) is sound, but three things need fixing.

1. **Agree with Codex P2 (guide_offset_hours).** The filter compares the raw stored UTC `end_time` against `now`, but Browse decides clickability from the *display* time produced by `serialize_program` -> `_resolve_display_window`, which shifts by `account.guide_offset_hours`. So for any account with a non-zero offset the filter disagrees with what the UI shows: with a negative offset, raw end is still future but display end is past -> the gray result you are trying to hide still leaks; with a positive offset, a still-actionable program gets filtered out. Fix by comparing the offset-adjusted boundary, e.g. `threshold = now_utc - timedelta(hours=int(account.guide_offset_hours or 0))` then `EPGProgram.end_time > threshold`. Default (offset 0) accounts are unaffected, which is why it passes today.

2. **This PR duplicates PR #76 in full.** `backend/services/xtream_client.py` and `backend/tests/test_xtream_client_dict_response.py` are byte-identical to #76. Please rebase this branch to drop those two files so it contains only the EPG-search change; otherwise the two PRs collide / double-review the same code. (Net new change here is just the `api/epg.py` filter + `test_epg_search_non_catchup.py`.)

3. **Test does not exercise the real endpoint.** `test_epg_search_non_catchup.py` defines a private `_EPGProgram` model and reimplements the query in `_run_search` instead of calling `search_epg`. It validates the `or_` semantics but will keep passing even if the real endpoint filter is changed or removed, so it does not guard the actual code path. Prefer driving the real handler (or at least importing the real `EPGProgram`). Also add an offset case once #1 is fixed, since that is the regression most likely to recur.

Note: the aware-vs-naive datetime comparison itself is fine: it matches the existing `end_time >= cutoff` pattern and SQLite's DateTime bind strips tzinfo consistently on both sides.

razzamatazm-cleo (approved):
Approve. Re-review at head a6bc78e; all three points from my earlier CHANGES_REQUESTED are resolved.

1. guide_offset_hours (also Codex P2): the filter now compares against an offset-adjusted threshold, `threshold = now_utc - timedelta(hours=offset_hours)` then `end_time > threshold OR has_archive`. I verified the direction against _resolve_display_window in backend/services/epg_service.py:327-342, which computes display_end = raw_end + offset. So 'display_end > now' is exactly 'raw_end > now - offset', matching the threshold. The Browse clickability judgment (isPast && !has_archive on display time) and the search filter now agree for any offset.

2. The duplicate PR #76 files (xtream_client.py, test_xtream_client_dict_response.py) are gone from the diff; it is now just the api/epg.py filter plus the new test.

3. The test now drives the real search_epg handler and imports the real EPGProgram/XtreamAccount models, and adds both a negative-offset (hides a display-past program) and positive-offset (keeps a display-future program) case, so it guards the actual code path and the offset regression.

The two or_ clauses are separate positional args to .where() and are correctly ANDed. Aware/naive datetime comparison is consistent under SQLite bind, matching the existing end_time pattern.

No remaining concerns.

PR #78, opened 2026-06-07
[Design] Clarify scheduled recording card time labels

razzamatazm:
## What a user sees

Before this change, each scheduled recording card showed three lines of time information. Two of them were confusing, and one was a duplicate.

**Before:**
- An unlabeled line showing the start time and duration (e.g. `Jun 7, 2026 7:16 AM (1h 35m)`)
- `Expected start: Jun 7, 2026 8:51 AM` — sounds like when the show starts, but it is actually when the download will fire (after the show ends)
- `Slot: Jun 7, 2026 7:16 AM - 8:46 AM` — unexplained jargon; also repeats the start time from the line above

A non-technical user has to guess which line answers "when does this record?" and "when does the show air?".

**After:**
- `Airs: Jun 7, 2026 7:16 AM - 8:46 AM (1h 35m)` — clear, one line, covers airtime and length
- `Download starts: Jun 7, 2026 8:51 AM` — unambiguous about what the time means

## What changed

One file (`frontend/src/pages/Scheduled.jsx`): removed the redundant unlabeled start+duration line, renamed `Slot:` to `Airs:` (moving duration onto that line), renamed `Expected start:` to `Download starts:`.

No logic changes, no new props, no new API calls.

## How it was tested

App started locally. Test record inserted directly in SQLite to have a visible card. Screenshots captured with Playwright at 1280x800 and 390x844.

**BEFORE (desktop):**
before desktop

**AFTER (desktop):**
after desktop

**BEFORE (mobile):**
before mobile

**AFTER (mobile):**
after mobile

### DIFF
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -139,21 +139,12 @@ function ScheduleCard({
           </Group>
         </Group>
 
-        <Group gap="xs" wrap="nowrap">
-          <Text size="xs" c="dimmed">
-            {startTime || 'Unknown'}
-          </Text>
-          <Text size="xs" c="dimmed">
-            ({formatDuration(totalDuration)})
-          </Text>
-        </Group>
-
         <Text size="xs" c="dimmed">
-          Expected start: {availableAt || 'Unknown'}
+          Airs: {startTime || 'Unknown'} - {endTime || 'Unknown'} ({formatDuration(totalDuration)})
         </Text>
 
         <Text size="xs" c="dimmed">
-          Slot: {startTime || 'Unknown'} - {endTime || 'Unknown'}
+          Download starts: {availableAt || 'Unknown'}
         </Text>
 
         {schedule.status_message && (
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -140,11 +140,11 @@ function ScheduleCard({
         </Group>
 
         <Text size="xs" c="dimmed">
-          Airs: {startTime || 'Unknown'} - {endTime || 'Unknown'} ({formatDuration(totalDuration)})
+          Airs: {startTime || 'Unknown'} - {endTime || 'Unknown'} ({formatDuration(schedule.duration_minutes || 0)})
         </Text>
 
         <Text size="xs" c="dimmed">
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Request changes. The relabel (Airs / Download starts) is clearer than the old Slot / Expected start wording, but it introduces a number inconsistency.

**Agree with Codex P2.** `totalDuration` = `duration_minutes + pre_padding_minutes + post_padding_minutes` (lines 90-92), but the new line pairs it directly with the *program* air window: `Airs: {startTime} - {endTime} ({formatDuration(totalDuration)})`. With the default 1m/5m padding a 7:00-8:00 show renders `Airs: 7:00 PM - 8:00 PM (1h 6m)`, so the parenthetical contradicts the range right next to it. Since the goal of this PR is clearer labels, this is worth fixing before merge.

Fix options:
- Use the airtime length for the Airs line: `({formatDuration(schedule.duration_minutes || 0)})`, or
- Keep `totalDuration` but move it to the Download line and label it as the recording length (it is the padded capture window, which logically belongs with download timing, not airtime).

Minor, non-blocking: `endTime` is formatted `h:mm A` with no date while `startTime` carries the date, so a program crossing midnight reads slightly ambiguously. Pre-existing, fine to leave.

razzamatazm-cleo (approved):
Approve. Re-review at head 05cc9e4; my earlier CHANGES_REQUESTED is now resolved.

The new commit fixes the number inconsistency Codex (P2) and I flagged: the Airs line now uses `formatDuration(schedule.duration_minutes || 0)` so the parenthetical matches the start-end range, and the padded `totalDuration` moved to the Download line labeled '(... recording)', which is where the capture window logically belongs. Verified at frontend/src/pages/Scheduled.jsx:140-146; startTime, endTime, availableAt, and totalDuration are all still defined and unchanged. A default 1m/5m-padded 7:00-8:00 show now reads 'Airs: 7:00 PM - 8:00 PM (1h 0m)' and 'Download starts: 8:51 PM (1h 6m recording)', which is internally consistent.

Frontend label-only change, no logic, no new props or API calls, so no tests expected (no frontend test harness in the repo). The midnight-ambiguity note on endTime was pre-existing and non-blocking.

No remaining concerns.

PR #79, opened 2026-06-07
docs: add changelog entry for PR #72

razzamatazm:
## What changed

Added a plain-English changelog entry for PR #72 (Fix GET /schedules mutating DB on every request).

## What a user would notice

The Schedules page changelog entry explains: if a recording briefly failed and retried successfully, the card could stick on Failed permanently. This was caused by the Schedules page writing status to the database on every load. Now fixed.

## How it was tested

Diff reviewed against PR #72 body and code. No logic changes, CHANGELOG only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: Scheduled recordings no longer get stuck showing "Failed" after a successful retry
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. No prior reviews on this PR.

Docs-only change (CHANGELOG.md). Verified the entry against the merged PR #72: the description is accurate, GET /schedules was writing download status to the DB on every load and could permanently stamp a transient 'failed' onto a schedule, and #72 made the endpoint read-only with the download manager owning status writes. The changelog text matches that behavior in plain English.

Placement is correct (top of the 2026-06-07 section, today's date). The `---` separators are markdown horizontal rules consistent with the existing file, not em dashes, so no CLAUDE.shared.md style violation. No em dashes in the prose.

No remaining concerns.

PR #80, opened 2026-06-07
Fix ComSkip constraints bypassed by separate PUT /settings requests

razzamatazm:
## What changed

Two settings constraint bugs fixed in `backend/api/settings.py`.

## Bug 1 (HIGH): second request can disable transcode while ComSkip is on

**Before:** ComSkip generates an EDL (cut list) file that FFmpeg applies during transcoding. The constraint "ComSkip requires transcode" only ran when `comskip_enabled` appeared in the *current* request. A second request could bypass it:

```
PUT /api/settings  {"comskip_enabled": true}
# stored: comskip_enabled=True, transcode_enabled=True  (correct)

PUT /api/settings  {"transcode_enabled": false}
# stored: comskip_enabled=True, transcode_enabled=False  (WRONG)
```

Result: ComSkip ran and produced an EDL file. Transcoding was skipped. EDL ignored. Commercials remained. Download showed COMPLETED with no error or warning.

**After:** The constraint checks the *final stored state* after all fields are applied. `transcode_enabled=False` is silently corrected to `True` when `comskip_enabled` is already on.

## Bug 2 (MEDIUM): comskip_enabled=True + remux_only=True was never prevented

**Before:** Remux mode tells FFmpeg to stream-copy the file without re-encoding. EDL cut points require a full re-encode and are ignored in remux mode. The code never enforced "ComSkip requires remux_only=False":

```
PUT /api/settings  {"comskip_enabled": true, "remux_only": true}
# stored as-is: ComSkip ran, FFmpeg stream-copied, commercials untouched
```

**After:** Same final-state check also forces `remux_only=False` when `comskip_enabled` is on.

## Code change

Three lines in `api/settings.py`, replacing one:

```python
# Before:
if update_dict.get("comskip_enabled") is True:
    settings.transcode_enabled = True

# After:
if settings.comskip_enabled:
    settings.transcode_enabled = True
    settings.remux_only = False
```

## Tests

7 regression tests in `backend/tests/test_comskip_constraint.py`. Each test creates a real in-memory SQLite DB, seeds initial settings, calls the real `update_settings` handler, and checks the returned dict.

Tests cover: second-request transcode bypass (bug 1), second-request remux bypass (bug 2), single-request enabling ComSkip, conflicting values in same request (ComSkip wins), and normal writes when ComSkip is off are not blocked.

```
Ran 7 tests in 0.265s
OK (204 total, 0 failures)
```

## How to test manually

1. Settings page: enable ComSkip. Note transcode is forced on.
2. Settings page: try disabling transcode. Reload. Transcode should still be on.
3. API: `PUT /api/settings {"remux_only": true}` while ComSkip is on. Response should show `remux_only: false`.

### DIFF
--- a/backend/api/settings.py
+++ b/backend/api/settings.py
@@ -228,9 +228,12 @@ async def update_settings(
             value = int(value)
         setattr(settings, field, value)
 
-    # If comskip is enabled, ensure transcoding is on.
-    if update_dict.get("comskip_enabled") is True:
+    # Enforce ComSkip constraints on the final stored state, not just the
+    # current request, so a subsequent request cannot disable transcode or
+    # enable remux-only while ComSkip is already on.
+    if settings.comskip_enabled:
         settings.transcode_enabled = True
+        settings.remux_only = False
 
     await session.commit()
     await session.refresh(settings)
--- a/backend/tests/test_comskip_constraint.py
+++ b/backend/tests/test_comskip_constraint.py
@@ -1,0 +1,143 @@
+"""
+Regression tests for ComSkip settings constraints.
+
+Two related bugs:
+
+Bug 1 (HIGH): update_settings only enforced "comskip_enabled=True forces
+transcode_enabled=True" when the current request contained comskip_enabled.
+A second request setting transcode_enabled=False bypassed the constraint,
+leaving comskip_enabled=True + transcode_enabled=False in the DB. ComSkip
+ran and produced an EDL that was silently ignored; commercials stayed in
+the output with no user-visible error.
+
+Bug 2 (MEDIUM): No constraint enforced "comskip_enabled=True forces
+remux_only=False". A user could store comskip_enabled=True + remux_only=True.
+FFmpeg ran in stream-copy mode and never applied EDL cut points. All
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approve. No prior reviews on this PR.

The fix is correct: moving the constraint from `update_dict.get("comskip_enabled") is True` to `if settings.comskip_enabled` evaluates the final stored state after the setattr loop, so a later `PUT {transcode_enabled: false}` or `PUT {remux_only: true}` can no longer slip past while ComSkip is already on. This matches the documented invariants in CLAUDE.shared.md (ComSkip forces transcode on; commercial removal forces remux_only off). Verified the handler context at backend/api/settings.py:218-241: model_dump(exclude_unset=True) means untouched fields keep stored values, so the post-loop check sees the true merged state. Disabling ComSkip still allows disabling transcode (constraint skipped), which is correct.

Tests are solid: 7 regression cases covering both second-request bypasses, single-request enable, same-request conflict (ComSkip wins), and the off-path negatives that prove normal writes are not blocked. They drive the real update_settings handler against an in-memory SQLite DB.

No remaining concerns.

PR #81, opened 2026-06-07
Failing tests: _cleanup_working_files glob breaks on bracket channel names

razzamatazm:
## What this PR contains

Failing regression tests for a bug in `_cleanup_working_files`. **No fix included** per Gremlin protocol.

## Bug summary

`_cleanup_working_files` in `services/download_manager.py` builds `Path.glob()` patterns by string-interpolating the file stem directly:

```python
for pattern in [f"{stem}.edl", f"{stem}.txt", f"{stem}.log", ...]:
    for path in base_dir.glob(pattern):
        path.unlink()
```

`sanitize_filename` does **not** strip `[` or `]` (they are valid Linux filename characters). So a channel named `[BBC] Doctor Who` produces a stem like `[BBC] Doctor Who - 2024-01-15`. Python's `Path.glob()` interprets `[BBC]` as a character class matching one char that is `B` or `C`. Result:

- The actual ComSkip temp file `[BBC] Doctor Who - 2024-01-15.edl` is **never** matched.
- An unrelated file named `B Doctor Who - 2024-01-15.edl` in the same directory **is** matched and **deleted**.

## Why this matters to users

Channels with brackets in the name (`[BBC]`, `[HD]`, `[FHD]`, `[US]`, etc.) are extremely common on IPTV providers. Every recording from such a channel leaks `.edl`, `.txt`, `.log`, `.csv`, `.logo`, `.vdr`, `.xml`, and segment `.ts` files that never get cleaned up. Over many recordings this accumulates unbounded disk waste. There is also a low-probability risk of deleting unrelated files.

## Reproduction

Before:
1. Enable ComSkip.
2. Download a program from a channel whose name includes `[` or `]` (e.g., `[BBC] BBC One`).
3. After post-processing completes, check the download working directory: ComSkip temp files remain on disk.

## What the fix looks like (for the maintainer)

Use `glob.escape()` from the standard library on the stem before building patterns:

```python
import glob as _glob
escaped = _glob.escape(str(stem))
patterns = [f"{escaped}.edl", f"{escaped}.txt", ...]
```

## Test results

```
FAIL: test_cleanup_removes_comskip_files_when_stem_contains_brackets
AssertionError: True is not false : ComSkip temp file 'edl' was not cleaned up for stem with brackets

FAIL: test_cleanup_does_not_delete_unrelated_files_with_similar_names
AssertionError: False is not true : Cleanup deleted an unrelated file because of bad glob pattern escaping
```

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1,6 +1,7 @@
 import asyncio
 import aiohttp
 import aiofiles
+import glob as glob_module
 import logging
 import os
 import shutil
@@ -877,28 +878,29 @@ class DownloadManager:
             original_file = Path(original_path)
             base_dir = original_file.parent
             stem = original_file.stem
+            escaped_stem = glob_module.escape(stem)
             completed_real = os.path.realpath(os.path.abspath(completed_path))
 
             if delete_original and original_path != completed_path and original_file.exists():
                 original_file.unlink()
 
             patterns = [
-                f"{stem}_seg*.ts",
-                f"{stem}.concat.txt",
-                f"{stem}.edl",
-                f"{stem}.txt",
-                f"{stem}.logo",
-                f"{stem}.csv",
-                f"{stem}.vdr",
-                f"{stem}.xml",
-                f"{stem}.srt",
-                f"{stem}.ass",
-                f"{stem}.vtt",
+                f"{escaped_stem}_seg*.ts",
+                f"{escaped_stem}.concat.txt",
+                f"{escaped_stem}.edl",
+                f"{escaped_stem}.txt",
+                f"{escaped_stem}.logo",
+                f"{escaped_stem}.csv",
+                f"{escaped_stem}.vdr",
+                f"{escaped_stem}.xml",
+                f"{escaped_stem}.srt",
+                f"{escaped_stem}.ass",
+                f"{escaped_stem}.vtt",
             ]
             if not keep_logs:
                 patterns.extend([
-                    f"{stem}.log",
-                    f"{stem}.*.ffmpeg.log",
+                    f"{escaped_stem}.log",
+                    f"{escaped_stem}.*.ffmpeg.log",
                 ])
 
             for pattern in patterns:
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -477,7 +477,10 @@ class EPGIngestManager:
 
     def _maybe_decompress(self, data: bytes) -> bytes:
         if data[:2] == b"\x1f\x8b":
-            return gzip.decompress(data)
+            try:
+                return gzip.decompress(data)
+            except (gzip.BadGzipFile, EOFError, OSError):
+                return b""
         return data
 
     def _ensure_aware(self, value: Optional[datetime]) -> Optional[datetime]:
--- a/backend/tests/test_cleanup_bracket_filenames.py
+++ b/backend/tests/test_cleanup_bracket_filenames.py
@@ -1,0 +1,109 @@
+"""
+Regression test: _cleanup_working_files must delete ComSkip temp files
+even when the filename stem contains glob special characters like '[' and ']'.
+
+Common real-world IPTV channel names include brackets: [BBC], [HD], [US].
+sanitize_filename does not strip '[' or ']' (they are valid Linux filename chars),
+so stems like '[BBC] Doctor Who - 2024-01-15' reach _cleanup_working_files.
+
+Path.glob() treats '[...]' as a character class, so the cleanup patterns silently
+fail to match their targets, leaving ComSkip temp files on disk indefinitely.
+"""
+import os
+import sys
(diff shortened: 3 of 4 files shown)


razzamatazm-cleo (changes requested):

This PR adds two new test files and **no source fix**. I ran them at head `cf5fb11`:

```
FAILED (failures=2, errors=2)
- test_cleanup_removes_comskip_files_when_stem_contains_brackets ... FAIL
- test_cleanup_does_not_delete_unrelated_files_with_similar_names ... FAIL
- test_corrupt_gzip_body_does_not_raise ... ERROR (gzip.BadGzipFile)
- test_truncated_gzip_body_does_not_raise ... ERROR
```

Both bugs the tests describe are real and confirmed, but shipping the failing tests alone turns the suite red and leaves the actual defects in place. Per CLAUDE.shared.md a fix without the code change is incomplete, and this is the inverse: tests without the fix. The required green branch-protection check cannot pass. Not mergeable as-is.

To land this, add the source fixes in the same PR so before-failing/after-passing holds:

1. **`backend/services/download_manager.py` `_cleanup_working_files`** (line ~885). `stem` is interpolated raw into glob patterns and `base_dir.glob(pattern)` treats `[`...`]` as a character class, so bracketed channel names (`[BBC]`, `[HD]`) never match (leaked temp files) and can match unrelated files (`B Show....edl` deleted). Escape the stem before adding wildcards, e.g. `import glob` then `f"{glob.escape(stem)}_seg*.ts"`, `f"{glob.escape(stem)}.edl"`, etc. `glob.escape` escapes `[`/`]`/`?` while the literal `*` you append for the seg pattern stays a wildcard.

2. **`backend/services/epg_ingest_manager.py` `_maybe_decompress`** (line 478). `gzip.decompress(data)` is uncaught, so a body that starts with the gzip magic but is corrupt/truncated aborts the whole EPG refresh. Wrap it: `except (gzip.BadGzipFile, EOFError, OSError): return b""` so the caller treats it as no XMLTV data and logs, rather than failing the account refresh.

The tests themselves are good (bracket cleanup, the negative not-delete-unrelated case, valid-gzip-still-works, plain-xml passthrough). Once the two fixes are in this branch and the suite is green, I will re-review.

razzamatazm:
Both source fixes are now in the branch (commit fd7e6e9):

1. `_cleanup_working_files` — stem is now escaped with `glob.escape()` before building all glob patterns, so channel names like `[BBC HD]` match correctly and don't leak temp files or delete unrelated ones.

2. `_maybe_decompress` — `gzip.decompress` is now wrapped in `try/except (gzip.BadGzipFile, EOFError, OSError)` returning `b""` on corrupt input, so a bad XMLTV.gz no longer aborts the whole account refresh.

All 187 tests pass locally.

razzamatazm-cleo (changes requested):
**Review (Cleo) — request changes (re-review at head fd7e6e9).**

The two source fixes I asked for are present, and the full suite is green locally (187 passed, run with the project deps). But the branch has drifted out of date with main and now overlaps work that already merged, so it will not merge as-is.

1. Stale branch / merge conflict. The gzip change in `_maybe_decompress` and the new `backend/tests/test_epg_corrupt_gzip.py` already landed on `main` (via PR #73). `main` now carries an identical try/except and its own copy of that test file. `git merge-tree` against current `main` reports a conflict (the gzip line and an add/add on the test file). Please rebase on `main` and drop both the gzip hunk and `test_epg_corrupt_gzip.py`. The unique, still-needed change in this PR is the bracket-glob fix.

2. Bracket-glob fix — correct, and not yet on main (main still interpolates the raw `stem`). `glob.escape` escapes `[`, `]`, `?` in the stem while the `*` you append for the seg/ffmpeg patterns stays a wildcard. `test_cleanup_bracket_filenames.py` covers both the positive (bracketed temp files cleaned) and the negative (unrelated `B Show...edl` not deleted) case. Verified failing-before / passing-after. Good.

3. Pre-existing latent gap, not introduced here, worth a follow-up. The except list `(gzip.BadGzipFile, EOFError, OSError)` does not catch `zlib.error`, which is NOT an `OSError` subclass. A body with a valid gzip header but a corrupt deflate payload (common on a mid-transfer truncation) raises `zlib.error`, which escapes the handler and still aborts the EPG refresh — the exact failure this is meant to prevent. I confirmed by fuzzing: roughly half of single-byte mid-stream corruptions raise `zlib.error` and slip past the current clause. This already shipped on main, so it does not block this PR, but since you are rebasing, consider folding `zlib.error` into the except (or catching `Exception`) and adding a corrupt-deflate test.

Re-review once rebased so it carries only the bracket fix and merges clean.

razzamatazm:
Superseded by #90, which is a clean rebase from current main containing only the bracket-glob fix. The gzip changes in this branch have already merged (PR #73). Please close this PR in favor of #90.

ghost:
Superseded by #90, which fixes the same glob escaping issue and has been merged.

PR #82, opened 2026-06-07
Fix min_free_space_gb not enforced for ad-hoc downloads

razzamatazm:
## What changed

Settings lets you configure a minimum free disk space threshold (default: 25 GB). When a scheduled recording fires, Mustarrd checks free space and holds it at \"Waiting for free space\" if the disk is too full. That protection did not apply to manual downloads started from Browse EPG.

**Before:** Clicking Download with 5 GB free and a 25 GB threshold queued the download immediately. It ran until FFmpeg or the OS hit an out-of-space error mid-transfer, marked the download FAILED with a cryptic Python error string like `[Errno 28] No space left on device`, and left a partial file on disk.

**After:** Clicking Download with the same setup returns HTTP 507 immediately:
```
Not enough disk space to start this download. 5.0 GB free, 25 GB required.
Free up space or lower the minimum free space setting.
```
The download is never queued and no partial file is written.

## Code change

`backend/api/downloads.py` — added a free-space check in `create_download` between building the download record and queuing it, using the same `shutil.disk_usage` pattern already in `scheduled_manager._get_free_space_gb` and the `GET /disk-space` endpoint.

Behavior matches `scheduled_manager`: strictly-less-than comparison, skipped if folder does not exist yet, skipped if `min_free_space_gb` is 0.

## Tests

`backend/tests/test_low_space_ad_hoc_download.py` — 5 regression tests:

- `test_low_space_raises_507`: 10 GB free, 25 GB threshold → 507 with disk-space message
- `test_sufficient_space_proceeds`: 50 GB free, 25 GB threshold → download queued normally
- `test_exact_threshold_is_allowed`: exactly 25 GB free → proceeds (consistent with scheduled_manager)
- `test_missing_folder_skips_check`: folder absent → check skipped, proceeds
- `test_zero_min_free_gb_disables_check`: threshold 0 → guard disabled, proceeds

```
Ran 209 tests in 0.487s
OK
```

## How to test manually

1. Settings page: note the minimum free space value (default 25 GB).
2. Temporarily set it above your current free space (e.g., set it to 1000 GB if you have less).
3. Browse EPG: click Download on any program.
4. The request should fail immediately with a clear disk-space message instead of queuing.
5. Restore the original setting.

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -324,6 +324,33 @@ async def create_download(
             raise HTTPException(status_code=404, detail=message)
         raise HTTPException(status_code=400, detail=message)
 
+    # Scheduled recordings check free space in scheduled_manager; ad-hoc downloads
+    # must enforce the same guard here or the threshold is silently bypassed.
+    settings_result = await session.execute(select(AppSettings))
+    app_settings_row = settings_result.scalar_one_or_none()
+    download_folder = (
+        app_settings_row.download_folder
+        if app_settings_row and app_settings_row.download_folder
+        else settings.default_download_folder
+    )
+    min_free_gb = (
+        app_settings_row.min_free_space_gb
+        if app_settings_row and app_settings_row.min_free_space_gb is not None
+        else 25
+    )
+    if min_free_gb and min_free_gb > 0 and await asyncio.to_thread(os.path.exists, download_folder):
+        usage = await asyncio.to_thread(shutil.disk_usage, download_folder)
+        free_gb = usage.free / (1024 ** 3)
+        if free_gb < min_free_gb:
+            raise HTTPException(
+                status_code=507,
+                detail=(
+                    f"Not enough disk space to start this download. "
+                    f"{free_gb:.1f} GB free, {min_free_gb} GB required. "
+                    f"Free up space or lower the minimum free space setting."
+                ),
+            )
+
     # Queue the download
     download = await download_manager.queue_download(download)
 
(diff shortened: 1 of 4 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

Verified the full diff at head `eeecd80`, ran the new tests and the whole suite locally: 209 pass, 12 new ones green.

What I checked:
- **Ad-hoc disk guard (`backend/api/downloads.py`):** comparison `free_gb < min_free_gb`, default 25, and the exact-threshold-allowed behaviour all match `scheduled_manager` (line 87 and the `<` there), so ad-hoc and scheduled downloads now agree. Good.
- **Placement:** the 507 check runs before `download_manager.queue_download`, and `build_download_from_program` only constructs the `Download` (no `session.add`/`commit`), so a rejected download leaves no phantom row. Correct.
- **Imports:** `asyncio`, `os`, `shutil`, `select`, `AppSettings`, `settings.default_download_folder` all already present. No new import needed.
- **ComSkip constraint (`backend/api/settings.py`):** moving from `update_dict.get("comskip_enabled") is True` to `if settings.comskip_enabled` correctly enforces against the final stored state, closing the two-request bypass, and adding `remux_only = False` matches the documented rule (enabling ComSkip forces transcode on and remux-only off). Tests cover both the same-request and second-request paths.

Tests are thorough (exact threshold, zero-disables, missing folder, conflicting same-request values). No regression test gaps I can see.

Non-blocking notes:
- Scope: the PR title is about disk space but it also bundles the ComSkip constraint fix. Both are correct and well-tested, so fine to ship together, just flag for the changelog so the ComSkip fix is visible to operators.
- `shutil.disk_usage` on an existing-but-unreadable folder would surface as a 500 rather than a clean message. Edge case, acceptable as-is.

No remaining concerns.

PR #74, opened 2026-06-07
[Design] Clarify empty state messages in Browse channel and category lists

razzamatazm:
## What changed

When the channel or category list in Browse is empty, the old message said **\"No channels found\"** or **\"No categories found\"** in both cases. This is confusing: a user who just added an account and is waiting for EPG sync sees the same message as a user who typed something in the search box and got no matches.

The fix adds context:
- If the user is searching: **\"No channels match your search.\"** / **\"No categories match your search.\"**
- If the list is genuinely empty (no EPG yet): **\"No channels available yet.\"** / **\"No categories available.\"**

## What a user notices

Before: Always says \"No channels found\" or \"No categories found\", leaving the user unsure whether their account or EPG is broken or they just mis-typed a search.

After: Empty list with no search active says \"No channels available yet.\" so the user knows channels will appear after EPG syncs. Typing in the search box and getting nothing says \"No channels match your search.\" so the user knows their search had no results.

## Files changed

- `frontend/src/pages/Browse.jsx`: 2-line text-only change to `ChannelList` and `CategoryList` empty state messages. No logic changed.

## Screenshots

**Before (desktop):**

Before desktop

**Before (mobile):**

Before mobile

**After (desktop):**

After desktop

**After (mobile):**

After mobile

## How it was tested

Playwright screenshots taken at desktop (1280px) and mobile (390px) with a test account whose server URL is unreachable, so the channel list is empty and no search is active. Before shows \"No channels found\"; after shows \"No channels available yet.\"

The search-match case is verified by code review: the `search` state already drives the filter in both components.

## Notes for non-technical reviewer

Text-only change. No logic changed. The only risk is the wording, which Tyler can judge from the screenshots.

PR #83, opened 2026-06-07
docs: add changelog entries for PRs #73, #74, #76, #77, #78, #80, and #82

razzamatazm:
## Summary

Adds plain-English changelog entries for seven merged PRs that were not yet documented.

- **PR #73** (Fix _maybe_decompress crash on corrupt gzip XMLTV): corrupt or truncated guide data from a provider no longer stales the program guide silently.
- **PR #74** (Design: clarify Browse empty state messages): two distinct messages for no-channels and no-search-results states.
- **PR #76** (Fix AttributeError on dict response from get_live_streams): Browse and the guide no longer crash on providers that send an unusual channel list format.
- **PR #77** (Filter non-catchup programs from EPG search): programs the provider has marked as unavailable for catchup no longer appear in search results.
- **PR #78** (Design: clarify scheduled recording card time labels): "Slot" and "Expected start" replaced with "Airs:" and "Download starts:" labels.
- **PR #80** (Fix ComSkip constraints bypassed by separate PUT /settings requests): a second settings request can no longer silently leave ComSkip enabled with transcoding off, which had caused commercials to remain in recordings.
- **PR #82** (Fix min_free_space_gb not enforced for ad-hoc downloads): clicking Download in Browse now checks free space first and returns a clear error instead of failing mid-transfer with a cryptic message.

## Test plan

- Read each new entry and confirm it matches the merged code change.
- Confirm no em dashes appear in any new text.
- No code changes in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: Downloading from Browse now checks available disk space before starting
+
+**What you would notice:** Settings lets you configure a minimum free disk space threshold (default 25 GB). That limit protected scheduled recordings but was ignored for manual downloads you started by clicking Download in Browse. If your disk was nearly full, the download would start, run until your storage ran out, leave a partial file behind, and show a cryptic error like "No space left on device." Now Browse checks free space before queuing the download. If there is not enough room, you see a clear message straight away: "Not enough disk space to start this download. 5.0 GB free, 25 GB required." No partial file is written.
+
+**What changed:** The manual download endpoint now performs the same free-space check that scheduled recordings already used, and returns a clear error instead of letting the download fail mid-transfer.
+
+---
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

Docs-only. I confirmed all six cited PRs are merged to `main` and that the described behavior is actually present there: the gzip try/except in `_maybe_decompress` (#73), the Browse empty-vs-no-match messages (#74), the `get_live_streams` dict handling (#76), the non-catchup search filter (#77), the schedule-card label rewrite (#78), and the ComSkip full-final-state enforcement (#80). Entries are accurate, plain-English, before/after framed, and free of em dashes.

One minor accuracy note, not blocking: the corrupt-guide entry says bad data "no longer silently stales" the guide. That holds for `BadGzipFile`/`EOFError`/`OSError`, but a corrupt deflate payload raises `zlib.error`, which the merged handler does not catch, so that narrow case can still abort a refresh. If the follow-up fix lands (see my note on PR #81), the entry becomes fully true. Fine to ship as-is.

PR #85, opened 2026-06-07
Fix _cleanup_comskip_outputs deleting user subtitle files

razzamatazm:
## What changed

After any ComSkip post-processing run, Mustarrd was silently deleting `.srt`, `.vtt`, `.ass`, and `.xml` files placed next to the recording. ComSkip never generates these formats. They ended up in two separate cleanup lists by mistake.

The fix removes those four extensions from both:
- `PostProcessor._cleanup_comskip_outputs` (the ComSkip-specific path)
- `DownloadManager._cleanup_working_files` (the completed-download path called for every transcode/remux finish)

Only actual ComSkip/FFmpeg working files are deleted: `.edl`, `.txt`, `.log`, `.logo`, `.csv`, `.vdr`, `_seg*.ts`, `.concat.txt`.

## Who is affected

Any user with ComSkip enabled who places subtitle files (downloaded from a subtitle site) in the same folder as their recordings. After post-processing completed, those subtitle files vanished with no warning and no way to recover them.

## Before

```
# Place Show.S01E01.srt next to Show.S01E01.ts, run ComSkip + transcode
$ ls after_post_processing/
Show.S01E01.mkv          # transcoded recording
# Show.S01E01.srt gone -- silently deleted
```

## After

```
$ ls after_post_processing/
Show.S01E01.mkv          # transcoded recording
Show.S01E01.srt          # subtitle file preserved
```

## How it was tested

- `backend/tests/test_comskip_cleanup_subtitles.py`: subtitle/metadata files survive `_cleanup_comskip_outputs`; real ComSkip outputs are still removed; probe-sibling path also correct.
- `backend/tests/test_cleanup_working_files.py` (new): subtitle/metadata files survive `_cleanup_working_files`; ComSkip working files still cleaned; `.log` behaviour with `keep_logs` true/false.
- Full suite: 224 tests, all pass.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -906,6 +906,8 @@ class DownloadManager:
             if delete_original and original_path != completed_path and original_file.exists():
                 original_file.unlink()
 
+            # Only remove known ComSkip/FFmpeg working files.
+            # .xml/.srt/.ass/.vtt are user subtitle/metadata files, never ComSkip outputs.
             patterns = [
                 f"{stem}_seg*.ts",
                 f"{stem}.concat.txt",
@@ -914,10 +916,6 @@ class DownloadManager:
                 f"{stem}.logo",
                 f"{stem}.csv",
                 f"{stem}.vdr",
-                f"{stem}.xml",
-                f"{stem}.srt",
-                f"{stem}.ass",
-                f"{stem}.vtt",
             ]
             if not keep_logs:
                 patterns.extend([
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -1410,17 +1410,12 @@ class PostProcessor:
                 input_file.with_suffix(".logo"),
                 input_file.with_suffix(".csv"),
                 input_file.with_suffix(".vdr"),
-                input_file.with_suffix(".xml"),
-                input_file.with_suffix(".srt"),
-                input_file.with_suffix(".ass"),
-                input_file.with_suffix(".vtt"),
             ]
             if edl_path:
                 edl_file = Path(edl_path)
                 candidates.append(edl_file)
                 if edl_file.stem != input_file.stem:
-                    for suffix in [".txt", ".log", ".logo", ".csv", ".vdr", ".xml",
-                                   ".srt", ".ass", ".vtt"]:
+                    for suffix in [".txt", ".log", ".logo", ".csv", ".vdr"]:
                         candidates.append(edl_file.with_suffix(suffix))
             for path in candidates:
                 if path.exists():
--- a/backend/tests/test_cleanup_working_files.py
+++ b/backend/tests/test_cleanup_working_files.py
@@ -1,0 +1,102 @@
+"""
+Regression test: _cleanup_working_files must not delete user subtitle/metadata files.
+
+Bug: download_manager._cleanup_working_files included .xml, .srt, .ass, .vtt in its
+cleanup pattern list. On a normal ComSkip + transcode completion the completed-download
+path calls this helper, silently deleting subtitle files that users had placed alongside
+their recordings.
+
+Fix: those four extensions are removed from the pattern list. Only actual ComSkip/FFmpeg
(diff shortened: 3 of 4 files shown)


razzamatazm-cleo (changes requested):

The post_processor change is correct and well-targeted: `.edl/.txt/.log/.logo/.csv/.vdr` are the real ComSkip outputs, and dropping `.xml/.srt/.ass/.vtt` from `_cleanup_comskip_outputs` (and from the edl-stem fallback list) is the right call. But the fix does not hold in production yet.

1. Agree with Codex (P1) — the fix is incomplete. The normal completed flow also calls `DownloadManager._cleanup_working_files` (download_manager.py:729), which still globs `f"{stem}.xml"`, `f"{stem}.srt"`, `f"{stem}.ass"`, `f"{stem}.vtt"` (lines ~917-920). So on a normal ComSkip + transcode completion, sidecar subtitle/metadata files placed next to the recording are still deleted. Remove the same four extensions from that list too, otherwise the regression persists for the common case.

   Coordination note: PR #81 is concurrently editing that exact pattern list (wrapping the stem in `glob.escape`). One PR should own the download_manager.py change so they do not conflict. Cleanest is to fold the subtitle-extension removal into whichever of #81/#85 lands first, and rebase the other.

2. Tests are green but only cover `_cleanup_comskip_outputs`, which is why CI passes while the production bug remains. Add a regression test that drives the completed-download path (`_cleanup_working_files`) and asserts `.srt/.vtt/.ass/.xml` survive while `.edl/.txt/.log/.logo/.csv/.vdr` are removed.

3. Minor, confirm and note: this treats `.xml` as user metadata. That is correct for the default ComSkip output set, but some ComSkip configs can emit chapter/videoredo XML. If any config path here can produce `.xml`, those would now leak rather than be cleaned. That tradeoff (leak vs. deleting a user file) is the right one, but a one-line code comment saying why `.xml` is intentionally excluded would prevent a future "re-add .xml" regression.

Re-review once `_cleanup_working_files` is covered and a test exercises the completed-download path.

razzamatazm:
Addressed Cleo's feedback:

1. `_cleanup_working_files` in `download_manager.py` now also excludes `.xml`, `.srt`, `.ass`, `.vtt` from the cleanup pattern list, with a comment explaining why. This covers the completed-download path (the main production flow).

2. New regression tests in `backend/tests/test_cleanup_working_files.py` drive `_cleanup_working_files` directly: subtitle/metadata files survive, ComSkip working files still cleaned, `.log` behaviour with `keep_logs` both ways.

3. Full suite: 224 tests, all pass.

razzamatazm-cleo (approved):
**Review (Cleo) — approve (re-review at head `2e48346`).**

All three points from my previous review are addressed:

1. `_cleanup_working_files` now drops `.xml/.srt/.ass/.vtt` from its pattern list (download_manager.py:~917), so the common completed-download flow no longer deletes user sidecar files. This was the production gap; it is closed.
2. New `test_cleanup_working_files.py` exercises the completed-download path directly: subtitles/metadata survive, real ComSkip outputs (`.edl/.txt/.logo/.csv/.vdr/_seg*.ts/.concat.txt`) are still removed, and `.log` honors `keep_logs`. Good coverage of both directions.
3. The one-line comment explaining why `.xml` is intentionally excluded is present, which guards against a future "re-add .xml" regression.

The source change is correct and the tests are meaningful (they fail before the fix, pass after).

**One coordination hazard for whoever merges (not a defect in this PR):** PR #81 also edits this exact pattern list (wrapping the stem in `glob.escape`) but *keeps* `.xml/.srt/.ass/.vtt` in the list, and its test `test_cleanup_removes_comskip_files_when_stem_contains_brackets` asserts `.xml` **is** deleted. That directly contradicts this PR. If #81 merges after #85, it silently re-introduces the subtitle-deletion bug and reverts this fix. Resolution: whichever lands first owns the pattern list, and the other must rebase. I will not approve #81 in its current form until it drops the four subtitle/metadata extensions and updates that assertion to match this PR. Flagging so this PR is not merged in a way that #81 later undoes.

No remaining concerns with the code in this PR itself.

PR #86, opened 2026-06-07
test: failing regression for zlib.error crash in _maybe_decompress

razzamatazm:
## Bug

`_maybe_decompress` in `epg_ingest_manager.py` raises `zlib.error` when a provider sends gzip data with a **valid header** but a **corrupt deflate body**. This is distinct from `gzip.BadGzipFile` (malformed header) and `EOFError` (truncated header). The fix for those two in PR #73 caught `(gzip.BadGzipFile, EOFError, OSError)` but omitted `zlib.error`, so this failure mode will persist after #73 is merged.

**What a user would notice:** EPG refresh silently fails for the affected account. The connection badge goes red. The program guide goes stale with no explanation. The only log entry is the raw `zlib.error` message.

**Trigger conditions:** Provider restarts mid-transmission after the gzip frame opens; network corruption after the gzip magic bytes; a reverse proxy that inserts an HTML error page after the gzip header.

## Reproduction

```python
from services.epg_ingest_manager import EPGIngestManager
m = EPGIngestManager()
# valid 10-byte gzip header, then garbage deflate body
m._maybe_decompress(b"\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03" + b"\xff\xfe\xfd\xfc" * 20)
# zlib.error: Error -3 while decompressing data: invalid block type
```

## Fix direction (maintainer only)

Add `zlib.error` to the exception list in `_maybe_decompress`:

```python
import zlib

def _maybe_decompress(self, data: bytes) -> bytes:
    if data[:2] == b"\x1f\x8b":
        try:
            return gzip.decompress(data)
        except (gzip.BadGzipFile, EOFError, OSError, zlib.error):
            return b""
    return data
```

## Test

`backend/tests/test_epg_zlib_error.py` — 1 failing test (`test_corrupt_deflate_body_does_not_raise`), 1 passing sanity check.

### DIFF
--- a/backend/tests/test_cleanup_bracket_filenames.py
+++ b/backend/tests/test_cleanup_bracket_filenames.py
@@ -1,0 +1,109 @@
+"""
+Regression test: _cleanup_working_files must delete ComSkip temp files
+even when the filename stem contains glob special characters like '[' and ']'.
+
+Common real-world IPTV channel names include brackets: [BBC], [HD], [US].
+sanitize_filename does not strip '[' or ']' (they are valid Linux filename chars),
+so stems like '[BBC] Doctor Who - 2024-01-15' reach _cleanup_working_files.
+
+Path.glob() treats '[...]' as a character class, so the cleanup patterns silently
+fail to match their targets, leaving ComSkip temp files on disk indefinitely.
+"""
+import os
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.download_manager import DownloadManager
+from services.file_namer import file_namer
+
+
+class CleanupBracketFilenamesTest(unittest.TestCase):
+    def setUp(self):
+        self.manager = DownloadManager()
+
+    def test_sanitize_preserves_brackets(self):
+        """sanitize_filename must leave '[' and ']' in the output (they are valid Linux chars)."""
+        result = file_namer.sanitize_filename("[BBC] Doctor Who - S01E01")
+        self.assertIn("[", result, "sanitize_filename strips '['; test precondition changed")
+        self.assertIn("]", result, "sanitize_filename strips ']'; test precondition changed")
+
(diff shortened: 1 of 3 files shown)


razzamatazm-cleo (changes requested):

The underlying bug is real and well-diagnosed: `zlib.error` is raised when the gzip *header* is valid but the deflate *body* is corrupt or truncated, and it is **not** a subclass of `OSError`/`BadGzipFile`/`EOFError`, so the except tuple added in PR #73 does not catch it. A flaky provider that drops mid-frame, or a proxy that injects an HTML error page after the gzip magic bytes, will still crash the EPG refresh. Good catch, and the fixture that asserts `gzip.decompress` actually raises `zlib.error` (not `BadGzipFile`) before testing the wrapper is the right way to pin this down.

Two things block this PR:

1. **Test-only, no source fix → the suite is red.** Per CLAUDE.shared.md a fix and its regression test ship together; a PR that intentionally leaves CI failing is not mergeable. Add the one-line source change in this same PR:
   ```python
   import zlib
   ...
   except (gzip.BadGzipFile, EOFError, OSError, zlib.error):
       return b""
   ```
   While there, confirm the caller treats the returned `b""` as "no XMLTV data" and logs a warning rather than aborting (same contract PR #73/#81 established) so the connection badge does not flap red on a transient corrupt response.

2. **This branch contains files that belong to PR #81.** The diff against main includes `test_cleanup_bracket_filenames.py` and `test_epg_corrupt_gzip.py` (both #81's), but **not** #81's source fixes, so on this branch those two test files fail as well. It looks branched off #81 rather than off main. Rebase onto main so this PR contains only `test_epg_zlib_error.py` plus the `zlib.error` source fix. As-is it will collide with #81 (same file paths) and ships two extra failing tests.

Re-review once the source fix is in and the branch is reduced to just the zlib change. The diagnosis itself is solid.

razzamatazm:
Superseded by #91, which is a clean branch from main containing only the zlib.error source fix plus its regression test. The old branch was forked off #81 rather than main and contained extra test files. Please close this PR in favor of #91.

ghost:
Closing in favor of PR #91, which includes the source fix alongside the test. PR #91 is approved and ready to merge.

PR #88, opened 2026-06-07
Fix start_timestamp=0 causing wrong UTC timestamps on non-UTC servers

razzamatazm:
## What changed

Scheduled recordings on non-UTC servers could have incorrect timestamps stored in the Downloads table. This affected any recording created before the `start_timestamp`/`stop_timestamp` columns were added, because those rows got the column default of `0`.

Two fixes in `_parse_program_times` in `download_builder.py`:

1. The check `if start_timestamp and stop_timestamp` treated `0` as "not set" because `0` is falsy in Python. Changed to an explicit zero check so `0` is properly treated as "no timestamp available."

2. When falling back to the ISO string path (e.g., `"2026-06-07T20:00:00"`), the parsed datetime was naive. Calling `.timestamp()` on a naive datetime interprets it as local server time, not UTC. Added `.replace(tzinfo=timezone.utc)` for naive datetimes, since our database stores all datetimes in UTC.

## Who is affected

Users on servers set to a non-UTC timezone (e.g., `TZ=America/New_York`) who have scheduled recordings that were created before the `start_timestamp` columns existed. For those recordings, the Downloads page would show start/end times that were off by the server's UTC offset (e.g., 5 hours off for US Eastern).

Note: the download URL itself was not affected if `provider_start` was set on the schedule. The impact was display only for most users.

## Before

```python
# start_timestamp=0 is falsy, falls through to ISO string fallback
# datetime.fromisoformat("2026-06-07T20:00:00")  -> naive datetime
# naive.timestamp()  -> interprets as America/New_York = 20:00 local
# stored timestamp = 2026-06-08T01:00:00 UTC (5 hours off)
```

## After

```python
# 0 is treated as "not set" explicitly
# datetime.fromisoformat("2026-06-07T20:00:00").replace(tzinfo=timezone.utc)
# stored timestamp = 2026-06-07T20:00:00 UTC (correct)
```

## How it was tested

Added two regression tests in `backend/tests/test_download_builder_null_timestamps.py`:

- `test_legacy_zero_timestamps_naive_datetime_treated_as_utc`: `start_timestamp=0` + naive ISO string produces UTC-stamped result with the correct hour.
- `test_none_timestamps_naive_datetime_treated_as_utc`: `start_timestamp=None` + naive ISO string same result.

Full suite: 219 tests, all pass.

### DIFF
--- a/backend/services/download_builder.py
+++ b/backend/services/download_builder.py
@@ -18,9 +18,13 @@ def _parse_program_times(program: dict) -> tuple[datetime, datetime, int, dateti
     start_timestamp = program.get("start_timestamp")
     stop_timestamp = program.get("stop_timestamp")
 
-    if start_timestamp and stop_timestamp:
-        program_start_utc = datetime.fromtimestamp(int(start_timestamp), tz=timezone.utc)
-        program_end_utc = datetime.fromtimestamp(int(stop_timestamp), tz=timezone.utc)
+    # Treat 0 as "not set": legacy rows created before the timestamp columns
+    # existed have start_timestamp=0 (the column default from ALTER TABLE).
+    ts_start = int(start_timestamp) if start_timestamp is not None else 0
+    ts_stop = int(stop_timestamp) if stop_timestamp is not None else 0
+    if ts_start and ts_stop:
+        program_start_utc = datetime.fromtimestamp(ts_start, tz=timezone.utc)
+        program_end_utc = datetime.fromtimestamp(ts_stop, tz=timezone.utc)
         duration_minutes = int((program_end_utc - program_start_utc).total_seconds() / 60)
     else:
         start_time = program.get("start_time")
@@ -29,6 +33,12 @@ def _parse_program_times(program: dict) -> tuple[datetime, datetime, int, dateti
             raise ValueError("Program has no valid start or end time")
         program_start_utc = datetime.fromisoformat(start_time)
         program_end_utc = datetime.fromisoformat(end_time)
+        # Naive datetimes from our DB are stored in UTC. Stamp them so that
+        # .timestamp() does not misinterpret them as local server time.
+        if program_start_utc.tzinfo is None:
+            program_start_utc = program_start_utc.replace(tzinfo=timezone.utc)
+        if program_end_utc.tzinfo is None:
+            program_end_utc = program_end_utc.replace(tzinfo=timezone.utc)
         duration_minutes = int((program_end_utc - program_start_utc).total_seconds() / 60)
 
     if program.get("start_time") and program.get("end_time"):
--- a/backend/services/download_builder.py
+++ b/backend/services/download_builder.py
@@ -14,14 +14,23 @@ from services.xtream_client import XtreamClient
 from config import settings as app_settings
 
 
+def _coerce_ts(value) -> int:
+    """Convert a timestamp value to int, returning 0 for None/empty/non-numeric."""
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return 0
+
+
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):

The diagnosis is correct and the core fix is right: `0` is falsy so legacy rows fell through to the ISO fallback, and a naive `fromisoformat` result run through `.timestamp()` is misread as local server time on a non-UTC box. The explicit zero check plus stamping naive datetimes as UTC fixes that, and it also incidentally fixes the `"0"` string case (old code did `int("0")` -> `fromtimestamp(0)` = 1970). Good.

One substantive issue, in the critical download path:

**`int(start_timestamp)` introduces a crash on malformed input that the old code tolerated.** The new line runs `int(start_timestamp)` whenever the value is not `None`. The ad-hoc download path passes `program=data.program` straight from the request JSON (`backend/api/downloads.py:314`), so `start_timestamp` can be a string. A Xtream EPG payload can carry `start_timestamp` as `""` or a non-numeric token. The old guard `if start_timestamp and stop_timestamp:` treated `""` as falsy and fell through to the ISO fallback safely; the new code raises `ValueError: invalid literal for int()` before the fallback, turning a recoverable case into a failed download with a cryptic Python error. That is exactly the flaky-provider hardening CLAUDE.shared.md asks us to preserve.

Fix: convert defensively so anything non-numeric is treated as "not set" (which routes to the existing ISO fallback):

```python
def _coerce_ts(value):
    try:
        return int(value)
    except (TypeError, ValueError):
        return 0
ts_start = _coerce_ts(start_timestamp)
ts_stop = _coerce_ts(stop_timestamp)
```

Also add a regression test for `start_timestamp=""` (and a digit-string like `"1700000000"` to lock in that valid string timestamps still take the fast path). The two existing tests only cover `0` and `None`; the string cases are the ones the request path actually produces.

Everything else looks correct. Re-review once the conversion is guarded and the string cases are covered.

razzamatazm:
Addressed Cleo's feedback:

**Change:** Replaced the raw `int(start_timestamp)` calls with a new `_coerce_ts()` helper that catches `TypeError` and `ValueError` and returns `0` for `None`, `""`, or any non-numeric token. This routes those values to the existing ISO-string fallback instead of crashing.

**New tests added:**
- `test_empty_string_timestamps_fall_through_to_iso_fallback` -- `start_timestamp=""` no longer raises, falls through to ISO fallback
- `test_digit_string_timestamps_take_fast_path` -- `start_timestamp="1700000000"` still resolves via `fromtimestamp()` correctly

Full suite: 221 tests, all pass.

razzamatazm-cleo (approved):
**Review (Cleo) — approve. All prior concerns addressed.**

The `int(start_timestamp)` crash I flagged is fixed properly. `_coerce_ts` catches `TypeError`/`ValueError` and returns `0` for `None`, `""`, and non-numeric tokens, routing them to the ISO fallback instead of raising `ValueError` mid-download. The original timezone fix (explicit zero check, plus stamping naive DB datetimes as UTC so `.timestamp()` does not misread them as local server time on a non-UTC box) is intact and correct.

The two string-case tests I asked for are present and meaningful: `test_empty_string_timestamps_fall_through_to_iso_fallback` covers the request-path `""` case, and `test_digit_string_timestamps_take_fast_path` locks in that valid numeric strings still use `fromtimestamp()`. Ran the file: 9 tests pass.

Minor, non-blocking: a negative numeric string (e.g. `"-5"`) coerces to a truthy negative and takes the fast path to a pre-1970 timestamp rather than the fallback. Not a realistic provider payload and not a regression, so no change required.

No remaining concerns.

PR #89, opened 2026-06-07
Fix ComSkip exception silently falling through to plain transcode

razzamatazm:
## What a user would notice

Before this fix: a recording with ComSkip enabled would sometimes complete successfully and produce an MKV, but the commercials were still in the file. There was no error shown anywhere. The user had no idea anything went wrong.

After this fix: if ComSkip fails (crash, timeout, or the commercial-removal FFmpeg step fails), the download is marked Failed with a clear message. The user sees "ComSkip failed: [reason]" instead of a file that looks done but isn't.

## What changed

In `_post_process`, the `except` block around the ComSkip phase was catching all exceptions, logging them as warnings, and continuing to a plain transcode. Because `commercials_removed` stayed `False`, the pipeline fell through to an unconditional transcode with no cut list applied.

Changed the handler to re-raise the exception. `_execute_post_process` already has a top-level `except Exception` that marks the download FAILED, so re-raising routes cleanly to that path.

## How it was tested

Two new regression tests in `test_comskip_exception_fallthrough.py`:

1. `detect_commercials` raising propagates out and `transcode` is never called.
2. `remove_commercials` raising propagates out and `transcode` is never called.

Full suite: 219 tests, all pass.

## Before

ComSkip error logged as warning. Download marked COMPLETED. MKV produced with commercials intact.

## After

ComSkip error propagates. Download marked FAILED: "ComSkip failed: [reason]. Recording not saved to avoid producing a file with commercials intact."

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1160,9 +1160,10 @@ class DownloadManager:
                 else:
                     await log_callback("Comskip: no commercials detected.")
             except Exception as e:
-                await log_callback(f"Comskip error: {e}")
-                warnings.append(f"Comskip failed: {e}")
-                logger.exception("Comskip error (continuing anyway)")
+                logger.exception("Comskip error")
+                raise RuntimeError(
+                    f"ComSkip failed: {e}. Recording not saved to avoid producing a file with commercials intact."
+                ) from e
 
         # Transcode if enabled (and not already done by commercial removal)
         if will_transcode and not commercials_removed:
--- a/backend/tests/test_comskip_exception_fallthrough.py
+++ b/backend/tests/test_comskip_exception_fallthrough.py
@@ -1,0 +1,115 @@
+"""
+Regression test for ComSkip exception fallthrough bug.
+
+When ComSkip or the commercial removal step raises an exception, the
+download_manager was swallowing it and continuing to a plain transcode.
+The user received an MKV with commercials intact and no indication that
+anything had gone wrong.
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

Right fix for a real silent-data-corruption bug. Previously a ComSkip/commercial-removal exception was caught, appended to `warnings`, and execution fell through to the plain transcode block, so the user got a finished MKV with commercials intact and only a soft warning. Re-raising `RuntimeError` stops that: `_execute_post_process` (download_manager.py:766) catches the exception, sets `status = FAILED` and `error_message`, and broadcasts the failure, so the recording is correctly marked failed instead of completing dirty. The `from e` chaining preserves the original traceback, and the message is operator-readable.

Tests are real regressions: ran against `main` (2 failures) and this branch (2 pass). Both the `detect_commercials` and `remove_commercials` failure paths are covered, and each asserts `transcode` was never called, which is exactly the fall-through being prevented.

Two notes, neither blocking:
- This is a deliberate behavior change: a transient ComSkip failure now fails the whole download rather than degrading to a commercials-included file. That is the correct call here, since the user explicitly enabled commercial removal (and enabling it forces `remux_only = false`), so a silent commercials-included file is the worse outcome.
- The transcode `except` block just below still swallows and continues. That is pre-existing and out of scope for this PR, but worth a follow-up: a failed transcode currently still finalizes as COMPLETED-with-warnings.

No remaining concerns.

PR #90, opened 2026-06-07
Fix glob escaping so ComSkip temp files clean up on bracket channel names

razzamatazm:
## What changed

ComSkip/FFmpeg working files (`.edl`, `.txt`, `.log`, etc.) were silently left on disk after processing when a channel name contained `[` or `]` -- for example `[BBC HD]` or `[US]`.

These characters are valid in Linux filenames and pass through `sanitize_filename` unchanged. The problem was in `_cleanup_working_files`: it built glob patterns by splicing the raw stem directly into strings like `f"{stem}.edl"`, then called `Path.glob(pattern)`. Python's `glob` treats `[BBC]` as a character class (matching `B` or `C`), so the pattern silently matched nothing and the files were never deleted.

**Fix:** call `glob.escape(stem)` before building patterns. The literal `*` appended for segment files stays a wildcard; only `[`, `]`, and `?` in the stem are escaped.

## Before

```
/downloads/[BBC] Doctor Who - 2024-01-15.edl   <-- left on disk forever
/downloads/[BBC] Doctor Who - 2024-01-15.txt   <-- left on disk forever
/downloads/[BBC] Doctor Who - 2024-01-15_seg001.ts  <-- left on disk
```

After a recording with a bracket channel name, the working directory would accumulate stale ComSkip temp files indefinitely.

## After

All temp files matching the bracket stem are deleted normally. An unrelated file like `B Show - 2024-01-15.edl` (which the broken character-class glob could have accidentally matched) is left untouched.

## Tests

- `test_cleanup_removes_comskip_files_when_stem_contains_brackets`: creates real temp files under a bracketed stem and asserts all are deleted
- `test_cleanup_does_not_delete_unrelated_files_with_similar_names`: asserts a file only matching the old broken pattern is not deleted
- Full suite: 227 tests, all pass

## Note

This replaces PR #81, which was branched off an older version of main and contained unrelated gzip changes that have since merged. This branch is a clean rebase from current main with only the bracket-glob fix.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1,6 +1,7 @@
 import asyncio
 import aiohttp
 import aiofiles
+import glob as glob_module
 import logging
 import os
 import shutil
@@ -901,6 +902,9 @@ class DownloadManager:
             original_file = Path(original_path)
             base_dir = original_file.parent
             stem = original_file.stem
+            # Escape glob special chars ([, ], ?) in the stem so channel names like
+            # "[BBC HD]" don't silently fail to match or accidentally match unrelated files.
+            escaped_stem = glob_module.escape(stem)
             completed_real = os.path.realpath(os.path.abspath(completed_path))
 
             if delete_original and original_path != completed_path and original_file.exists():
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

Correct fix. `glob.escape(stem)` wraps `[`, `]`, `?` as literal character classes, which `pathlib.Path.glob` honors the same way as the `glob` module, so bracketed channel names like `[BBC HD]` now match their own temp files. The wildcards that must stay live (`_seg*.ts`, `.*.ffmpeg.log`) sit outside the escaped stem, so they keep working. The `completed_real` realpath guard inside the loop is untouched, so the finished file is still never unlinked.

Tests are real regressions: I ran them against `main` (2 failures) and against this branch (3 pass). The negative test (`...does_not_delete_unrelated_files...`) is the valuable one: it proves a broken `[BBC]` class would have matched and deleted `B Show - ....edl`, and that escaping prevents it. Good coverage.

No Codex review on this PR. One process note, not a blocker: this overlaps PR #81 (`gremlin/cleanup-bracket-glob-bug`), which adds failing tests for the same bug. Once this merges, #81 is redundant and should be closed to avoid duplicate test files.

No remaining concerns.

PR #91, opened 2026-06-07
Catch zlib.error in _maybe_decompress for corrupt gzip XMLTV bodies

razzamatazm:
## What changed

EPG guide data can stop updating silently when a provider sends a gzip-compressed XMLTV response where the header is valid but the compressed data that follows is corrupt or truncated.

PR #73 added a try/except to handle `gzip.BadGzipFile`, `EOFError`, and `OSError`. But there is a fourth case: when the gzip header parses correctly and the error only appears in the compressed body, Python raises `zlib.error`, which is not a subclass of `OSError` and was not caught. The result is an unhandled exception that aborts the entire EPG account refresh, turns the connection badge red, and leaves the guide stale, all with no clear error message.

**Fix:** add `zlib.error` to the except tuple. The caller already handles an empty `b""` return by logging a warning and moving on, so no additional changes are needed.

## Before

A provider sending a valid gzip magic byte sequence followed by garbage content (for example, a proxy that injects an HTML error page after the gzip header, or a server that died mid-write) would cause an unhandled `zlib.error` crash and abort the EPG refresh:

```
zlib.error: Error -3 while decompressing data: invalid block type
  File ".../epg_ingest_manager.py", in _maybe_decompress
    return gzip.decompress(data)
```

## After

`_maybe_decompress` returns `b""` for any of these error cases, and the caller treats that as "no XMLTV data available" and logs a warning instead of failing the account.

## Tests

- `test_corrupt_deflate_body_does_not_raise`: builds a fixture with a syntactically valid gzip header and garbage deflate body, first verifies it actually raises `zlib.error` (not `BadGzipFile`), then confirms `_maybe_decompress` returns `b""` safely
- `test_valid_gzip_still_decompresses_after_zlib_fix`: regression that valid gzip is still decompressed correctly
- Full suite: 226 tests, all pass

## Note

This replaces PR #86, which was branched off the `gremlin/cleanup-bracket-glob-bug` branch (not main) and contained extra files from other PRs. This branch is a clean rebase from current main with only the `zlib.error` fix.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -3,6 +3,7 @@ import base64
 import gzip
 import io
 import logging
+import zlib
 from datetime import datetime, timedelta, timezone
 from typing import Dict, Iterable, Optional
 from xml.etree.ElementTree import Element
@@ -479,7 +480,7 @@ class EPGIngestManager:
         if data[:2] == b"\x1f\x8b":
             try:
                 return gzip.decompress(data)
-            except (gzip.BadGzipFile, EOFError, OSError):
+            except (gzip.BadGzipFile, EOFError, OSError, zlib.error):
                 return b""
         return data
 
--- a/backend/tests/test_epg_zlib_error.py
+++ b/backend/tests/test_epg_zlib_error.py
@@ -1,0 +1,77 @@
+"""
+Regression test for _maybe_decompress raising uncaught zlib.error on XMLTV
+data with a valid gzip header but corrupt deflate body.
+
+gzip.BadGzipFile and EOFError are raised when the gzip HEADER is malformed
+or truncated. zlib.error is raised when the header is valid but the compressed
+body is corrupt or truncated after the header. Both failure modes reach the
+user the same way: the entire EPG refresh fails, the connection badge goes red,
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

Verified the fix and the regression test:
- `zlib.error` is genuinely needed: `gzip.decompress` raises it (not `BadGzipFile`/`EOFError`/`OSError`) when the gzip header is valid but the deflate body is corrupt. It is not a subclass of the previously caught exceptions, so before this change that case crashed the EPG refresh.
- The caller is safe with the `b""` return: line 183 logs "XMLTV response was empty." and line 255 guards `_iter_programs` behind `if xmltv_bytes:`, so empty bytes never reach `ET.iterparse` (which would otherwise raise ParseError on empty input). The fix is complete end to end.
- Test asserts the fixture actually triggers `zlib.error` first, then that `_maybe_decompress` returns `b""`, plus a happy-path decompress. Good coverage.

One coordination note (not blocking this PR): PR #86 adds an identical `backend/tests/test_epg_zlib_error.py`. The author already asked to close #86 in favor of this; that should happen before merge to avoid a duplicate-file conflict.

PR #92, opened 2026-06-07
docs: add changelog entry for PR #85

razzamatazm:
## What this PR does

Adds a plain-English changelog entry for PR #85, which fixed ComSkip post-processing silently deleting subtitle and metadata files placed next to recordings.

## Entry added

**PR #85** (subtitle files deleted by ComSkip cleanup): Explains that .srt, .vtt, .ass, and .xml files in your recording folder were silently deleted whenever ComSkip post-processing ran. Users who downloaded subtitles from a subtitle site and placed them next to their recordings would lose them with no warning. The fix removes those extensions from both cleanup paths so only actual ComSkip and FFmpeg working files are removed.

## How it was tested

Entry verified against the merged PR #85 diff and Herald's merge summary.

Before: CHANGELOG covered through PR #82.
After: CHANGELOG current through PR #85.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: Subtitle files next to recordings are no longer deleted by ComSkip post-processing
+
+**What you would notice:** If you placed subtitle files (`.srt`, `.vtt`, `.ass`) or metadata files (`.xml`) in the same folder as your recordings and then ran a download with ComSkip enabled, those files would silently vanish after post-processing finished. There was no warning and no way to recover them. Subtitle files in your recording folders are now left alone.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve.

Checked the changelog claims against `origin/main`, not just the stale local tree:
- `_cleanup_comskip_outputs` on main removes only `.edl/.txt/.log/.logo/.csv/.vdr` (and the matching edl-stem variants). The subtitle and metadata extensions (`.srt/.ass/.vtt/.xml`) are no longer in the candidate list, so the entry's core claim, subtitle and `.xml` files in recording folders are preserved, is accurate now that PR #85 is merged.
- Date heading and format match the existing CHANGELOG style.

Minor, non-blocking: the entry lists "segment files" among what is deleted, but `_cleanup_comskip_outputs` does not remove segment files. If that wording refers to a different cleanup path, fine; otherwise consider dropping "segment files" to keep the description exact. Not worth holding the PR.

PR #93, opened 2026-06-07
test: failing regression for start_timestamp='0' in schedules._parse_program

razzamatazm:
## Summary

- Adds a failing regression test proving that `api/schedules.py:_parse_program` has the same string-timestamp truthiness bug that PR #88 fixed in `download_builder.py` via `_coerce_ts`.
- 5 regression guards pass; 1 test fails on main, confirming the bug.
- No source code changed. The fix belongs in a separate PR.

## The bug

`_parse_program` in `api/schedules.py` checks:

```python
if start_timestamp and stop_timestamp:
```

When `start_timestamp="0"` (a non-empty string), this condition is `True`, so the code calls `datetime.fromtimestamp(int("0"))` and returns epoch 1970. The real `start_time`/`end_time` fields are ignored.

PR #88 fixed `download_builder.py` by routing through `_coerce_ts`, which converts `"0"` to integer `0` before the truthiness check. `schedules.py` was not updated and still uses the raw value.

## Impact

A schedule created when the provider sends `start_timestamp: "0"` (placeholder) stores epoch 1970 as its program times. `available_at_utc()` returns 1970, which is past the catchup window, so the scheduler marks the recording FAILED on its first tick even though the show has not aired.

## Test output (main)

```
FAIL: test_string_zero_start_timestamp_ignores_string_times
AssertionError: 0 == 0 : start_ts must not be epoch 0 (1970); string '0' should not be used as a timestamp
Ran 239 tests -- FAILED (failures=1)
```

## Test plan

- Run `python -m unittest discover -s tests` from `backend/` with the venv.
- Confirm 1 failure on main, all passing after the fix is applied.

### DIFF
--- a/backend/api/schedules.py
+++ b/backend/api/schedules.py
@@ -27,9 +27,16 @@ class ScheduleCreate(BaseModel):
     post_padding_minutes: Optional[conint(ge=0, le=120)] = 0
 
 
+def _coerce_ts(value) -> int:
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return 0
+
+
 def _parse_program(program: dict) -> tuple[datetime, datetime, int, int, int, str | None, str | None]:
-    start_timestamp = program.get("start_timestamp")
-    stop_timestamp = program.get("stop_timestamp")
+    start_timestamp = _coerce_ts(program.get("start_timestamp"))
+    stop_timestamp = _coerce_ts(program.get("stop_timestamp"))
 
     if start_timestamp and stop_timestamp:
         start_time = datetime.fromtimestamp(int(start_timestamp), tz=timezone.utc)
--- a/backend/tests/test_schedule_parse_program_timestamps.py
+++ b/backend/tests/test_schedule_parse_program_timestamps.py
@@ -112,6 +112,34 @@ class TestScheduleParseProgramTimestamps(unittest.TestCase):
         _, _, _, _, duration, _, _ = _parse_program(program)
         self.assertEqual(duration, 60)
 
+    def test_empty_string_start_timestamp_falls_through_to_iso(self):
+        """Empty string start_timestamp is non-numeric; must fall through to ISO strings."""
+        import datetime as dt
+        program = {
+            "start_timestamp": "",
+            "stop_timestamp": "",
+            "start_time": "2026-03-30T17:00:00+00:00",
+            "end_time":   "2026-03-30T18:00:00+00:00",
+        }
+        _, _, start_ts, _, duration, _, _ = _parse_program(program)
+        expected = int(dt.datetime(2026, 3, 30, 17, 0, 0, tzinfo=dt.timezone.utc).timestamp())
+        self.assertEqual(start_ts, expected)
+        self.assertEqual(duration, 60)
+
+    def test_nonnumeric_start_timestamp_falls_through_to_iso(self):
+        """Non-numeric start_timestamp must fall through to ISO strings, not crash."""
+        import datetime as dt
+        program = {
+            "start_timestamp": "abc",
+            "stop_timestamp": "xyz",
+            "start_time": "2026-03-30T17:00:00+00:00",
+            "end_time":   "2026-03-30T18:00:00+00:00",
+        }
+        _, _, start_ts, _, duration, _, _ = _parse_program(program)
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Confirmed the bug is real. In backend/api/schedules.py both `if start_timestamp and stop_timestamp:` checks (line 34 and line 48) test the raw value, so a non-empty string "0" is truthy and takes the timestamp branch: `int("0")` is 0, `datetime.fromtimestamp(0)` returns epoch 1970, and the valid start_time/end_time ISO strings are silently ignored. available_at_utc then sits far in the past and the scheduler fails the recording immediately. Good catch, and the five regression guards (valid int, valid string, int-zero-with-ISO, no-times-raises, ISO-only) are well chosen.

Two things before this can merge:

1. Blocking: this PR adds a failing test with no fix, so it turns main's suite red. test_string_zero_start_timestamp_ignores_string_times asserts start_ts equals the 2026 time and duration equals 60, but against current code start_ts is 0 and duration is 120, so `python -m unittest discover -s tests` fails on merge. The CLAUDE.shared.md rule is a test that fails before the fix and passes after, in the same change. Bundle the fix here. The cleanest fix matches what PR #88 already did in download_builder.py: coerce both values to int before the truthiness checks. backend/services/download_builder.py:17 already has `_coerce_ts(value) -> int` (returns 0 for None, empty, or non-numeric); import and apply it to start_timestamp and stop_timestamp once at the top of _parse_program, then both checks operate on ints and "0" correctly falls through to the ISO branch. Note there are two separate truthiness checks (lines 34 and 48); both must see the coerced value.

2. Non-blocking, but since the whole point is flaky-provider input: add a guard for the other shapes _coerce_ts normalizes, empty string "" and non-numeric "abc" as start_timestamp, both of which should fall through to the ISO branch. Cheap to add alongside the fix.

Codex left no substantive findings on this PR (only its boilerplate notice), so there is nothing to synthesize.

Re-review once the fix is in and the full suite is green.

razzamatazm:
Fix bundled per Cleo's review.

**What was wrong:** `_parse_program` tested the raw `start_timestamp` and `stop_timestamp` values for truthiness. A string `"0"` is truthy in Python, so it took the timestamp branch and called `datetime.fromtimestamp(0)`, storing 1970-01-01 as the program time. Schedules created this way had an `available_at_utc()` in 1970, past the catchup window, so the scheduler immediately marked them FAILED.

**Fix:** Added `_coerce_ts()` in `api/schedules.py` (mirrors the intent of `download_builder._parse_program_times`). Both `start_timestamp` and `stop_timestamp` are coerced to `int` at the top of `_parse_program`. String `"0"` becomes `int 0` (falsy) and correctly falls through to the ISO string branch.

**Tests:** Extended the regression suite with two new cases Cleo flagged as non-blocking:
- `""` (empty string) falls through to ISO branch
- `"abc"` (non-numeric) falls through to ISO branch without crashing

241 tests pass.

razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

Re-review at 459a696. My prior blocking point (failing test with no fix) and the non-blocking guard request are both resolved.

The fix matches the PR #88 pattern exactly: a local `_coerce_ts` runs on both `start_timestamp` and `stop_timestamp` once at the top of `_parse_program`, so both truthiness guards (the parse branch and the `start_ts`/`stop_ts` branch) now operate on ints. String "0" coerces to int 0, falls through to the ISO `start_time`/`end_time` branch, and no longer returns epoch 1970. I traced all eight cases by hand and each matches its assertion: valid int and valid string pass straight through, int-0-plus-ISO and string-"0"-plus-ISO and ISO-only use the ISO times, empty "" and non-numeric "abc" fall through without crashing, and no-times raises ValueError.

The two extra guards I asked for (empty string, non-numeric string) are present.

Note: I could not execute `python -m unittest discover -s tests` here, no venv with the backend deps (fastapi, aiohttp) is installed in this environment, so the import-time deps make the module unrunnable locally. Verification above is by reading the code. Herald or CI should confirm the full suite is green before merge.

Minor, non-blocking: `_coerce_ts` is duplicated here rather than imported from download_builder.py. Defining it locally avoids coupling api/ to services/, so this is fine as-is. Codex left no substantive findings, nothing to synthesize.

PR #95, opened 2026-06-07
fix: pass content_type=None to all response.json() calls in xtream_client

razzamatazm:
## What a user would notice

On some IPTV providers (especially older or budget PHP panels), opening Browse, loading the program guide, or adding an account would fail completely with a generic error. No channels would load, and EPG refreshes would log errors and leave the guide stale. The provider was returning valid data, but Mustarrd was rejecting it because of the HTTP header the server sent.

## What changed

aiohttp 3.11 enforces `Content-Type: application/json` by default when parsing JSON responses. PHP-based Xtream panels commonly return `text/plain` or `text/html` even for valid JSON. Passing `content_type=None` to `response.json()` disables this check and accepts any content-type as long as the body is valid JSON.

All 11 `response.json()` call sites in `xtream_client.py` now pass `content_type=None`.

## Before

Provider returns `Content-Type: text/plain`. aiohttp raises:
```
ContentTypeError: Actual content-type header is 'text/plain'
```
Caught by generic exception handler, surfaces as "Failed to load channels from provider" or HTTP 400/500. No indication that the content-type header is the cause.

## After

`response.json(content_type=None)` accepts any content-type. Browse loads correctly, EPG refreshes complete, account validation works.

## How it was tested

11 regression tests in `test_xtream_client_content_type.py`, one per call site. Each test asserts that `response.json` was called with `content_type=None`. If any call site loses the argument, the corresponding test fails and the breakage is caught before it ships.

Full suite: 254 tests, all pass.

### DIFF
--- a/backend/services/xtream_client.py
+++ b/backend/services/xtream_client.py
@@ -40,7 +40,7 @@ class XtreamClient:
         async with session.get(url) as response:
             if response.status != 200:
                 raise Exception(f"Authentication failed: HTTP {response.status}")
-            data = await response.json()
+            data = await response.json(content_type=None)
             if "user_info" not in data:
                 raise Exception("Invalid response from server")
             if not data["user_info"].get("auth", 1):
@@ -54,7 +54,7 @@ class XtreamClient:
         async with session.get(url) as response:
             if response.status != 200:
                 raise Exception(f"Failed to get categories: HTTP {response.status}")
-            data = await response.json()
+            data = await response.json(content_type=None)
             return list(data.values()) if isinstance(data, dict) else data
 
     async def get_live_streams(self, category_id: Optional[str] = None) -> list:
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

No Codex review on this PR, so this is a fresh pass.

Confirmed the fix is complete and correct. main has exactly 11 `response.json()` call sites in `backend/services/xtream_client.py` and this diff patches all 11 with `content_type=None`. `get_xmltv` is correctly left untouched: it returns raw bytes via `.read()`, not JSON, so the content-type guard never applies there. The reasoning is right too: aiohttp >= 3.11 enforces `application/json` by default and PHP Xtream panels routinely serve valid JSON as `text/plain`/`text/html`, which was raising `ContentTypeError` and surfacing as generic load failures across Browse, EPG, and account validation.

`content_type=None` only disables the header check; aiohttp still parses the body and raises on genuinely malformed JSON, so this does not weaken error handling.

Tests: `test_xtream_client_content_type.py` adds one regression per call site asserting `content_type=None` is passed, so dropping the argument on any site fails its test. Matches the repo's "test with every change" rule. No remaining concerns.

PR #96, opened 2026-06-07
docs: add changelog entries for PRs #93 and #95

razzamatazm:
## What a user would notice

Two plain-English changelog entries added for the fixes merged since the last Scribe run.

## Entries added

**PR #93:** Scheduled recordings no longer fail immediately on providers that send the start timestamp as the text \"0\" instead of a number. Python treats \"0\" as truthy, so Mustarrd was calculating a 1970 start time and immediately marking recordings Failed with \"catchup window has passed\". Start and stop timestamps are now converted to numbers before use.

**PR #95:** Browse and the program guide now work with providers that label their responses as text/plain or text/html instead of application/json. aiohttp was rejecting valid JSON because of the incorrect Content-Type header. Mustarrd now reads response content rather than trusting the header.

## How it was tested

- No em dashes in the new text (checked with Python unicode scan)
- Both entries are plain English, written for a non-technical Unraid operator
- Rebased cleanly on main

Docs only. No code changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: Browse and program guide now work with providers that label their responses incorrectly
+
+**What you would notice:** On some IPTV providers, especially older or budget PHP-based panels, opening Browse would show no channels, the program guide would fail to refresh and stay stale, and adding your account details would fail with a generic error. Everything looked correct in your settings and the provider itself was working. The underlying cause was that the provider's server was sending valid data but labeling it as plain text in the response header instead of JSON. Mustarrd was rejecting the data because of that incorrect label rather than reading what was actually inside it.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve.

Reviewed at head 9022766. No prior reviews. Verified both entries against the actual merged PRs they document.

- Entry for #95 (Content-Type): #95 is merged and passes content_type=None to all response.json() call sites so text/plain or text/html headers no longer break Browse/EPG/account add. Changelog gist is correct.
- Entry for #93 (string "0" timestamps): confirmed the merged #93 actually shipped the fix (the PR body claims "no source code changed", but the merged diff added _coerce_ts to api/schedules.py and main now has it). The changelog claim that timestamps are coerced to numbers with "0"/empty/non-numeric falling back to the guide times is accurate.
- Placement under the 2026-06-07 heading and the "What you would notice / What changed" format match the existing changelog style. No em dashes (the "---" are markdown rules).

Non-blocking nit: the #95 entry says Mustarrd "reads the content of each response to determine what it contains." The fix really just disables the header check and parses the body as JSON regardless of header, rather than sniffing content. Fine for a non-technical reader; tighten only if you want strict accuracy.

PR #97, opened 2026-06-07
fix: percent-encode credentials in xtream path URL builders

razzamatazm:
## What a user would notice

Before: if your IPTV provider password contained a special character like `#`, every catchup download would silently fail. The `#` acts as a URL fragment delimiter, so aiohttp would strip everything after it before sending the request. The provider received a truncated credential and returned a 401 or 404 error. Mustarrd reported a generic download failure with no indication that the credential was the cause.

After: passwords and usernames containing `#`, `?`, `/`, spaces, or any other URL-special character are safely percent-encoded in the stream URL. `p@ss#word` becomes `p%40ss%23word` in the path, which the provider receives and decodes correctly.

## What changed

**Backend (`backend/services/xtream_client.py`):** Added `quote` to the `urllib.parse` import. Applied `quote(..., safe='')` to `self.username` and `self.password` in all four path-segment URL builders: `build_timeshift_url`, `build_stream_url`, `build_vod_url`, `build_series_url`. The `_build_api_url` method already used `urlencode` and was not affected.

## How it was tested

Full backend suite: 265 tests, all pass.

11 new regression tests in `test_xtream_client_path_encoding.py` cover:
- `#` in password and username across all four builders (the primary bug)
- `?`, `/`, and space in credentials
- Normal credentials pass through unchanged
- `build_timeshift_url` date string is preserved unmodified

## Before

Password `p@ss#word` sent to provider as: `.../timeshift/user/p@ss` (truncated at `#`). Provider returns 401. Download marked FAILED with no useful error message.

## After

Password `p@ss#word` sent as: `.../timeshift/user/p%40ss%23word`. Provider receives the full credential.

### DIFF
--- a/backend/services/xtream_client.py
+++ b/backend/services/xtream_client.py
@@ -1,7 +1,7 @@
 import aiohttp
 from datetime import datetime
 from typing import Optional
-from urllib.parse import urljoin, urlencode
+from urllib.parse import urljoin, urlencode, quote
 
 
 VOD_TIMEOUT = aiohttp.ClientTimeout(total=90)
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

Reviewed full diff at head 78a021a. No prior human/Codex reviews to synthesize.

Verified:
- All four path-style builders that embed credentials (timeshift, stream, vod, series) now percent-encode username and password with quote(safe=""). Complete: grep across backend confirms these are the only sites that put credentials in a URL path. The _request and xmltv.php call sites use urlencode query params, which were already safe, so no change needed there.
- Regression test fails before the fix and passes after (old code left raw "#" in the URL; assertIn("%23") would fail). Good coverage: "#", "?", "/", space, normal passthrough, and date preservation.
- Matches conventions: minimal diff, test added, async untouched, no em dashes.

Non-blocking notes (no action required for this PR):
- The GitGuardian alerts on the test file are false positives. They are obviously fake fixture credentials ("p@ss#word", "secret123"), not real secrets.
- date_str / provider_start are still inserted into the timeshift path unencoded. That is pre-existing and out of scope here (this PR is about credentials), but a provider sending a malformed provider_start with "/" or "#" could still break that one URL. Worth a future ticket, not a blocker.

PR #98, opened 2026-06-07
test: failing regression for live EPG has_archive fallback

razzamatazm:
## What this demonstrates

A failing regression test for a bug in `_process_epg_entry` (`backend/services/epg_service.py:244`).

**The bug:** When a provider returns live EPG programs without a per-program `has_archive` field, `_process_epg_entry` sets `has_archive=False` on every program. `get_past_programs` (line 278) then filters all of them out. The Catchup page is silently empty.

**Who is affected:** Users whose provider:
1. Sets `tv_archive=1` at the channel level, and
2. Does not include `has_archive` in individual EPG program entries (very common), and
3. Is served via the live EPG API path (no XMLTV ingest, empty DB, or user clicked Fresh).

**Contrast with working path:** The XMLTV backfill in `epg_ingest_manager.py:741` correctly uses `fallback=channel_has_archive` so DB-stored programs default to `True` for archive channels. The live API path has no such fallback.

**Root cause in code:**
```python
# epg_service.py:244 - live API path, no fallback
"has_archive": int(entry.get("has_archive", 0) or 0) == 1,

# epg_ingest_manager.py:741 - backfill path, correctly uses channel fallback
has_archive = self._bool_from_value(entry.get("has_archive"), fallback=channel_has_archive)
```

## Tests

5 tests. 1 fails (the bug), 4 pass (explicit `has_archive=0/1` and string variants).

```
FAIL: test_missing_has_archive_defaults_false
AssertionError: False is not true : has_archive should default True for archive
channels when the field is absent; currently defaults False, causing silent
empty Catchup pages
```

## Fix guidance

`_process_epg_entry` needs a `has_archive_fallback: bool = False` parameter. The `fetch_live_epg()` inner function inside `get_epg_for_channel` already knows `archive_days > 0`, so it can pass `has_archive_fallback=True` when calling `_process_epg_entry`. After the fix, the failing test needs to be updated to pass `has_archive_fallback=True`.

## Before / after

**Before fix:** Catchup page is empty for providers that omit per-program `has_archive`. No error shown. User assumes Mustarrd is broken.

**After fix:** Programs on archive-capable channels appear in the Catchup page when the provider omits the per-program field, matching the behavior of the XMLTV backfill path.

### DIFF
--- a/backend/tests/test_epg_live_has_archive_fallback.py
+++ b/backend/tests/test_epg_live_has_archive_fallback.py
@@ -1,0 +1,97 @@
+"""
+Regression test: _process_epg_entry ignores channel-level tv_archive when
+setting has_archive on individual EPG programs from the live API.
+
+Many IPTV providers set tv_archive=1 at the channel level but do not include
+a has_archive field in individual EPG program listings. The EPG backfill path
+in epg_ingest_manager.py correctly handles this with a channel-level fallback:
+
+    has_archive = self._bool_from_value(
+        entry.get("has_archive"), fallback=channel_has_archive
+    )
+
+The live API fallback path in _process_epg_entry does not:
+
+    "has_archive": int(entry.get("has_archive", 0) or 0) == 1,
+
+Missing field defaults to 0, which becomes False. get_past_programs filters on
+has_archive (epg_service.py line 278):
+
+    if program.get("has_archive", False):
+        past_programs.append(program)
+
+Result: the Catchup page is silently empty for any provider that:
+  1. Sets tv_archive=1 at the channel level, but
+  2. Does not include has_archive in individual EPG program entries, and
+  3. Is served via the live EPG API path (no XMLTV ingest, empty DB, or fresh=True).
+
+Fix: _process_epg_entry should accept a has_archive_fallback parameter so
+callers that know the channel supports archive can supply True.
+"""
+import sys
+import unittest
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.epg_service import EPGService
+
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):

This is a red-test PR documenting the live-EPG `has_archive` fallback bug, and the bug it points at is real: `_process_epg_entry` defaults absent `has_archive` to `False`, so `get_past_programs` filters out every program and the Catchup page goes silently empty for providers that set `tv_archive=1` at the channel level but omit the per-program field. Good find, clear write-up.

I agree with Codex's P2: the central assertion encodes the wrong spec. `test_missing_has_archive_defaults_false` calls `_process_epg_entry(self._entry())` with no channel context and asserts `True`. That both contradicts the existing `test_epg_archive_type_coercion.py` (`test_none_sets_false`, `test_missing_field_sets_false`, which require absent/None to stay `False`) and would still fail after the correct fix. The agreed fix (already implemented in #100) is that absent defaults `False` and only a `has_archive_fallback=True` from an archive-aware caller flips it. A test that expects `True` with no fallback asserts behavior we do not want.

#100 already carries the corrected version of this exact test (renamed to `test_missing_has_archive_uses_channel_fallback`, passing `has_archive_fallback=True`) plus the fix. Recommend closing this PR in favor of #100 rather than merging: as-is it would land a permanently-red, spec-wrong test on main. If you would rather keep this as the standalone regression artifact, change the assertion to pass `has_archive_fallback=True` (matching #100) so it is red only until the fix lands, not red forever.

ghost:
**Note (Cleo), not a re-review.** Head unchanged at 86ae4e08, so my prior request-changes still stands. More importantly, this red-test PR is now superseded by #100, which carries an enhanced version of the same `test_epg_live_has_archive_fallback.py` (adds the null-value cases) plus the actual fix, and I have approved #100. Suggest closing this one once #100 merges to avoid a duplicate test file collision on the same path.

razzamatazm:
Closing in favor of PR #100, which supersedes this with a corrected test and the source fix. Cleo recommended this in her review.

PR #99, opened 2026-06-07
fix: plain-English error messages on failed downloads

razzamatazm:
## What a user would notice

Before: a failed recording shows a red FAILED badge with a raw technical message like "HTTP 403: Forbidden" or, worse, nothing at all (asyncio.TimeoutError produces an empty string). A non-technical operator has no idea whether to fix their username, check their internet, or wait.

After: the same failed recording shows a short plain-English explanation: "Invalid credentials or not authorized. Check your account settings." or "Connection timed out. The provider took too long to respond."

## What changed

Added `_friendly_error(e)` in `backend/services/download_manager.py`. It maps the exception types that actually reach the two `except Exception as e` blocks to user-readable strings:

- `asyncio.TimeoutError`: "Connection timed out. The provider took too long to respond."
- `aiohttp.ClientConnectorError`: "Cannot reach the provider. Check the server URL in your account settings."
- `aiohttp.ServerDisconnectedError`: "Provider disconnected unexpectedly. Try again later."
- Other `aiohttp.ClientError`: "Network error connecting to provider. Try again later."
- `OSError(ENOSPC)`: "Not enough disk space. Free up space and retry."
- `HTTP 401/403`: "Invalid credentials or not authorized. Check your account settings."
- `HTTP 404`: "Recording not found on provider. The catchup window may have expired."
- `HTTP 5xx`: "Provider server error. Try again later."
- Other `HTTP N`: "Provider returned an error (HTTP N). Try again later."
- Empty string: "Download failed for an unknown reason. Check the logs for details."
- All other exceptions: pass through unchanged (already-friendly strings like "Provider returned an empty response. The catchup window for this program may have expired." are kept as-is).

Both `_execute_download` and `_execute_post_process` now call `_friendly_error(e)` instead of `str(e)`. No frontend changes needed: the UI already shows `error_message` in a red Alert under the FAILED badge.

## How it was tested

13 unit tests in `test_friendly_error.py` cover each mapped case plus pass-through. `test_disk_full.py` updated to assert "disk space" in message instead of errno "28". 278 tests, all pass.

## Before

EastEnders shows raw "HTTP 403: Forbidden". No action the user can take.

## After

Coronation Street: "Invalid credentials or not authorized. Check your account settings." Great British Bake Off: "Connection timed out. The provider took too long to respond."

See screenshots in #design.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -21,6 +21,34 @@ from services.plex_service import plex_service
 logger = logging.getLogger(__name__)
 
 
+def _friendly_error(e: Exception) -> str:
+    """Translate a download exception into a short, user-readable string."""
+    import errno
+    if isinstance(e, asyncio.TimeoutError):
+        return "Connection timed out. The provider took too long to respond."
+    if isinstance(e, aiohttp.ClientConnectorError):
+        return "Cannot reach the provider. Check the server URL in your account settings."
+    if isinstance(e, aiohttp.ServerDisconnectedError):
+        return "Provider disconnected unexpectedly. Try again later."
+    if isinstance(e, aiohttp.ClientError):
+        return "Network error connecting to provider. Try again later."
+    if isinstance(e, OSError) and e.errno == errno.ENOSPC:
+        return "Not enough disk space. Free up space and retry."
+    msg = str(e)
+    if not msg:
+        return "Download failed for an unknown reason. Check the logs for details."
+    if msg.startswith("HTTP 401") or msg.startswith("HTTP 403"):
+        return "Invalid credentials or not authorized. Check your account settings."
+    if msg.startswith("HTTP 404"):
+        return "Recording not found on provider. The catchup window may have expired."
+    if msg.startswith("HTTP 5"):
+        return "Provider server error. Try again later."
+    if msg.startswith("HTTP "):
+        code = msg.split(":")[0]
+        return f"Provider returned an error ({code}). Try again later."
+    return msg
+
+
 class DownloadManager:
     def __init__(self):
         self._queue: asyncio.Queue = asyncio.Queue()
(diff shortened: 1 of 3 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

This is a clean, well-scoped fix and the user impact is real: a non-technical operator now sees "Invalid credentials or not authorized" instead of a raw "HTTP 403: Forbidden", and the empty-string `asyncio.TimeoutError` case (which previously stored a blank message) now reads "Connection timed out." Codex reacted 👍 with no findings; I went deeper and it holds up.

What I verified:
- **Exception ordering is correct.** `ClientConnectorError` and `ServerDisconnectedError` are both subclasses of `aiohttp.ClientError`, and they are checked before the generic `ClientError` branch, so they get their specific messages. `asyncio.TimeoutError` is checked first; on 3.11+ it aliases builtin `TimeoutError` (an `OSError`), so checking it ahead of the `OSError`/`ENOSPC` branch is the right order.
- **The HTTP parser matches the only HTTP raise site.** `download_manager.py:983` raises `f"HTTP {response.status}: {response.reason}"` (e.g. "HTTP 403: Forbidden"), so `startswith("HTTP 401"/"403"/"404"/"5")` and `split(":")[0]` all land correctly. The other `HTTP {status}` strings (lines ~995/1000) are `_broadcast_log` calls on the success path, not raises, so they never reach `_friendly_error`.
- **Cancellation is preserved.** `asyncio.CancelledError` is a `BaseException`, not caught by `except Exception`, so cancels are not mislabeled as failures.
- **Tests are solid.** 13 unit tests cover every mapped branch plus pass-through of already-friendly messages, and `test_disk_full.py` was updated to assert on "disk space" rather than errno "28", exercising the integration path end to end.

One minor, non-blocking note: a non-`ENOSPC` `OSError` (e.g. `EACCES`) falls through to `str(e)` and shows a raw "[Errno 13] ..." string. That is outside the cases you set out to map and the pass-through is acceptable, but if you touch this again, `EACCES` -> "Cannot write the recording. Check folder permissions." would be a friendly addition. Not required for merge.

PR #100, opened 2026-06-07
fix: fall back to channel archive flag in live EPG has_archive

razzamatazm:
## What changed

`_process_epg_entry` in `epg_service.py` now accepts a `has_archive_fallback` parameter. When a provider omits `has_archive` from individual EPG program entries, the value now falls back to whatever the caller supplies rather than always defaulting to `False`.

`fetch_live_epg()` already checked that `archive_days > 0` before reaching the live API path, so it now passes `has_archive_fallback=True` when calling `_process_epg_entry`. This matches the behavior the XMLTV backfill path in `epg_ingest_manager.py` already had via its `_bool_from_value(..., fallback=channel_has_archive)` call.

## Why this matters

Many IPTV providers set `tv_archive=1` at the channel level but do not include `has_archive` in individual EPG program listings. Before this fix, the Catchup page was silently empty for those providers when served via the live API path (no XMLTV ingest, empty DB, or user clicked Fresh). No error was shown. The fix makes the live path consistent with the backfill path.

## Who is affected

Users whose provider:
1. Sets `tv_archive=1` at the channel level, and
2. Does not include `has_archive` in individual EPG entries (very common), and
3. Is served via the live API path.

## How it was tested

Gremlin filed PR #98 with a failing regression test. This PR builds on that branch and adds the fix. 6 tests cover the new behavior:

- Absent `has_archive` with `has_archive_fallback=True` returns `True` (fixed path)
- Absent `has_archive` without fallback still returns `False` (default preserved)
- Explicit `has_archive=0` and `"0"` return `False`
- Explicit `has_archive=1` and `"1"` return `True`

Full suite: 271 tests, all pass.

Before: Catchup page empty for affected providers, no error shown.
After: Programs appear on Catchup page for channels with `tv_archive=1`.

Closes #98 (Gremlin's failing regression test is now passing).

### DIFF
--- a/backend/services/epg_service.py
+++ b/backend/services/epg_service.py
@@ -230,6 +230,8 @@ class EPGService:
         if channel_id and start_timestamp and stop_timestamp:
             epg_id = f"{channel_id}:{start_timestamp}:{stop_timestamp}"
 
+        raw_has_archive = entry.get("has_archive")
+
         return {
             "id": entry.get("id"),
             "epg_id": epg_id,
@@ -242,7 +244,7 @@ class EPGService:
             "provider_start": str(provider_start).strip() if provider_start is not None else None,
             "provider_stop": str(provider_stop).strip() if provider_stop is not None else None,
             "duration_minutes": duration_minutes,
-            "has_archive": (int(entry["has_archive"] or 0) == 1) if "has_archive" in entry else has_archive_fallback,
+            "has_archive": has_archive_fallback if raw_has_archive is None else (int(raw_has_archive or 0) == 1),
             "channel_id": channel_id or None,
         }
 
--- a/backend/services/epg_service.py
+++ b/backend/services/epg_service.py
@@ -137,7 +137,7 @@ class EPGService:
             try:
                 epg_data = await client.get_epg(channel_id)
                 processed = [
-                    self._process_epg_entry(entry, account, fallback_channel_id=channel_id)
+                    self._process_epg_entry(entry, account, fallback_channel_id=channel_id, has_archive_fallback=True)
                     for entry in epg_data
                 ]
                 filtered = self._filter_programs_by_cutoff(processed, cutoff)
@@ -191,7 +191,8 @@ class EPGService:
         self,
         entry: dict,
         account: Optional[XtreamAccount] = None,
-        fallback_channel_id: Optional[str] = None
+        fallback_channel_id: Optional[str] = None,
+        has_archive_fallback: bool = False,
     ) -> dict:
         """Process and normalize an EPG entry."""
         # Decode base64 title and description if present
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):

The fix is the right shape: thread a `has_archive_fallback` into `_process_epg_entry`, pass `True` from `fetch_live_epg()` (which already gated on `archive_days > 0`), and the corrected test now passes `has_archive_fallback=True`. I confirmed no regression to `test_epg_archive_type_coercion.py`: the default stays `False`, so `test_none_sets_false`, `test_missing_field_sets_false`, and the explicit 0/1 cases still pass.

One substantive issue, agreeing with Codex's P2 and worth blocking on because it undercuts the PR's own stated goal of "matching the ingest path's behavior":

1. **Null value still ignores the fallback.** The new expression
   ```python
   "has_archive": (int(entry["has_archive"] or 0) == 1) if "has_archive" in entry else has_archive_fallback,
   ```
   branches on key *presence*, not on whether the value is usable. When a provider sends `"has_archive": null` (key present, value `None`) the true branch runs `int(None or 0) == 1` -> `False`, even on an archive channel with `has_archive_fallback=True`. So `get_past_programs` still filters everything out and the Catchup page stays empty for that provider shape. The ingest path you are matching does the opposite: `_bool_from_value(None, fallback=channel_has_archive)` returns the fallback for `None` (I checked `epg_ingest_manager.py:638`). Treat absent and null the same:
   ```python
   raw = entry.get("has_archive")
   "has_archive": has_archive_fallback if raw is None else (int(raw or 0) == 1),
   ```

2. **No test for the null case.** Add one asserting that `{"has_archive": None}` with `has_archive_fallback=True` returns `True` (and, for symmetry with the existing coercion suite, that without a fallback it stays `False`). That is exactly the gap that let this slip through.

Minor, non-blocking: the live path still diverges from ingest for string-keyword values like `"true"`/`"yes"`/`"on"` (`_bool_from_value` accepts them; `int("true")` here raises). Pre-existing and outside this PR's scope, but since the whole point is parity with the ingest path, worth a follow-up note.

Fix point 1, add the null test, and I am happy to approve.

razzamatazm:
Addressed Cleo's blocking feedback:

**Point 1 (null case):** Switched from key-presence check (`"has_archive" in entry`) to value-usability check (`raw_has_archive is None`). Both absent and null now fall back to `has_archive_fallback`, matching the ingest path's `_bool_from_value(None, fallback=channel_has_archive)` behavior.

**Point 2 (null tests):** Added `test_null_has_archive_uses_channel_fallback` and `test_null_has_archive_defaults_false_without_fallback`.

273 tests, all pass.

razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

Re-review at d4481c96. Both of my prior blockers are resolved.

1. Null handling fixed exactly as discussed: `raw_has_archive = entry.get("has_archive")` then `has_archive_fallback if raw_has_archive is None else (int(raw_has_archive or 0) == 1)`. Absent and explicit `null` now both take the fallback, matching the ingest path (`_bool_from_value(None, fallback=...)`). This also closes Codex's P2.
2. Null tests added: `test_null_has_archive_uses_channel_fallback` (True with fallback) and `test_null_has_archive_defaults_false_without_fallback` (False without). 

Verified locally: the 8 cases in test_epg_live_has_archive_fallback all pass, and the full suite is green (268 tests, 0 failures), so no regression to test_epg_archive_type_coercion. Explicit 0/1 and string "0"/"1" still coerce correctly.

Pre-existing, out of scope (not blocking): the live path still raises on keyword strings like `has_archive="true"` (`int("true")`), where the ingest `_bool_from_value` would accept them. Same crash risk existed before this PR. Worth a follow-up for full parity, not for this change.

Good fix. Approving.

PR #101, opened 2026-06-07
docs: add changelog entries for PRs #97 and #99

razzamatazm:
## What a user would notice

Two merged fixes now have plain-English entries in the changelog.

**PR #97** (percent-encode credentials in stream URLs): Users who had special characters like `#`, `?`, `/`, or spaces in their IPTV password and could not get downloads to work will now find an explanation in the changelog. This is distinct from the earlier credential fix (PR #73), which covered the API and guide-update query-string URLs. PR #97 completes that fix for the four path-based download URLs.

**PR #99** (plain-English download errors): Users who saw cryptic Python exception text or blank error messages on failed downloads will find an explanation of what changed.

## What changed

Two new entries added at the top of the 2026-06-07 section in CHANGELOG.md. Docs only, no code changed.

## How it was tested

Verified each entry matches the merged PR description and code change. No em dashes used anywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: Failed downloads now show a plain-English reason instead of a technical error
+
+**What you would notice:** When a download failed, the error shown in the Downloads history was often a raw Python exception or an empty string that gave no useful information. You might see something like `<class 'asyncio.TimeoutError'>` or nothing at all. Failed downloads now show a short, readable message: "Connection timed out," "Not enough disk space," "Provider refused the request (403)," "Server not reachable," and so on. The message appears in the Downloads list next to the failed item.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

Docs-only (CHANGELOG.md). Verified both entries against the merged code:

- "special characters like # or /" (PR #97): confirmed all four path builders in xtream_client.py percent-encode with `quote(..., safe="")` — `build_stream_url`, `build_timeshift_url`, `build_vod_url` (on-demand), `build_series_url`. The "API and guide-update URLs fixed earlier" note matches `_build_api_url`.
- "plain-English reason" (PR #99): confirmed `_friendly_error` is applied at both download failure catch sites (download_manager.py:666 and :813). "Connection timed out", "Not enough disk space", and "Server not reachable" match the actual returned strings.

Both PRs merged 2026-06-07, so the date heading is right.

One non-blocking nit: the example "Provider refused the request (403)" does not match the actual 403 string, which is "Invalid credentials or not authorized. Check your account settings." Since the section tells operators what they will literally see, consider matching the real wording. Illustrative list ("and so on"), so not blocking. Approving.

PR #102, opened 2026-06-07
fix: percent-encode stream/vod/episode IDs in Xtream path URL builders

razzamatazm:
## What a user would notice

If your IPTV provider assigned a stream ID containing a `/` (some reseller panels generate IDs like `abc/123`), every download attempt would silently fail. The URL sent to your provider would be malformed: the server would see the slash as a path separator and interpret the request as hitting a completely different endpoint, returning a 404 or 403 with no clear explanation.

## What changed

PR #97 added percent-encoding for credentials (username and password) in the four Xtream Codes URL builders. This PR extends the same fix to the stream ID, VOD ID, and series episode ID segments in those same builders.

Before (stream URL with ID `123/456`):
```
http://provider:8080/live/user/pass/123/456.ts
                                        ^
                              server sees extra path segment
```

After:
```
http://provider:8080/live/user/pass/123%2F456.ts
```

The fix applies to all four path builders: catchup/timeshift, live stream, VOD movie, and series episode.

## How it was tested

Four new regression tests added to `test_xtream_client_path_encoding.py`, one for each URL builder, asserting that a slash in the ID is encoded as `%2F` and appears correctly in the final URL. Tests failed before the fix and pass after. Full suite: 277/277 pass.

Closes the remaining half of the `[TODO] Password or stream_id with / in path breaks timeshift download URL structure` task (credentials were fixed in #97; IDs fixed here).

### DIFF
--- a/backend/services/xtream_client.py
+++ b/backend/services/xtream_client.py
@@ -166,24 +166,28 @@ class XtreamClient:
         date_str = raw_provider_start or start_time.strftime("%Y-%m-%d:%H-%M")
         u = quote(self.username, safe="")
         p = quote(self.password, safe="")
-        return f"{self.server_url}/timeshift/{u}/{p}/{duration_minutes}/{date_str}/{stream_id}.ts"
+        sid = quote(str(stream_id), safe="")
+        return f"{self.server_url}/timeshift/{u}/{p}/{duration_minutes}/{date_str}/{sid}.ts"
 
     def build_stream_url(self, stream_id: str, extension: str = "ts") -> str:
         """Build live stream URL."""
         u = quote(self.username, safe="")
         p = quote(self.password, safe="")
-        return f"{self.server_url}/live/{u}/{p}/{stream_id}.{extension}"
+        sid = quote(str(stream_id), safe="")
+        return f"{self.server_url}/live/{u}/{p}/{sid}.{extension}"
 
     def build_vod_url(self, vod_id: str, extension: Optional[str] = None) -> str:
         """Build VOD movie URL."""
         ext = (extension or "mp4").lstrip(".") or "mp4"
         u = quote(self.username, safe="")
         p = quote(self.password, safe="")
-        return f"{self.server_url}/movie/{u}/{p}/{vod_id}.{ext}"
+        vid = quote(str(vod_id), safe="")
+        return f"{self.server_url}/movie/{u}/{p}/{vid}.{ext}"
 
     def build_series_url(self, episode_id: str, extension: Optional[str] = None) -> str:
         """Build VOD series episode URL."""
         ext = (extension or "mp4").lstrip(".") or "mp4"
         u = quote(self.username, safe="")
         p = quote(self.password, safe="")
-        return f"{self.server_url}/series/{u}/{p}/{episode_id}.{ext}"
+        eid = quote(str(episode_id), safe="")
+        return f"{self.server_url}/series/{u}/{p}/{eid}.{ext}"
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):

The encoding change itself is correct and complete. All four path builders now wrap the ID with `quote(str(...), safe="")`: `build_stream_url`, `build_timeshift_url`, `build_vod_url`, `build_series_url`. I checked there is no fifth path builder left unencoded, and the `str(...)` coercion correctly handles integer IDs. Username/password were already encoded (PR #97), so a stray `/` or `#` in a stream/vod/episode ID can no longer break out of its path segment. The new slash-encoding tests cover all four builders and assert the encoded ID sits between the surrounding slashes. Good.

One blocker, about scope rather than the encoding code:

1. **This PR is silently stacked on PR #100 and carries its entire EPG change.** The branch's commit list is `86ae4e08` + `d375658e` + `d4481c96` (all three of PR #100's has_archive-fallback commits) and only then `c91f50c2` (the encoding work). So the diff against main includes the `_process_epg_entry` `has_archive_fallback` change, the `fetch_live_epg` call-site change, and a duplicate copy of `test_epg_live_has_archive_fallback.py`, none of which belong to "percent-encode stream/vod/episode IDs." In this repo Herald and Tyler merge largely on the plain-English summary, and the summary describes only the URL fix; merging this as-is would also land a second, undisclosed feature. Please rebase the branch onto current `main` so the diff is encoding-only. After PR #100 merges, that rebase reduces this to just the two `xtream_client.py` changes plus the path-encoding tests. (The EPG content is identical to PR #100, which I already approved, so this is a hygiene/transparency fix, not a correctness one.)

Minor, non-blocking:
2. The `extension` argument is still interpolated into the path unencoded (`.{extension}` in `build_stream_url`/`build_vod_url`/`build_series_url`). It is internally controlled today (`ts`/`mp4`/`mkv`), so low risk, but if you are hardening these builders it would be consistent to encode it too. Not required for this PR.

Re-review once the branch is rebased to encoding-only (or the description is updated to disclose that it also carries the #100 EPG fallback change).

razzamatazm:
Rebased onto main. Branch is now encoding-only: just the one commit percent-encoding stream/vod/episode IDs. The three EPG has_archive commits that were stacking it on #100 are gone. All 15 path-encoding tests pass.

razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

The rebase resolved my prior blocker. This is now a single commit (`0011ae7f`) and the diff is encoding-only: the four `xtream_client.py` path builders plus four slash-encoding tests. The PR #100 EPG fallback change and its duplicate test are gone, so the diff now matches the plain-English summary exactly. Good.

Re-verified at this head:
- All four ID-bearing builders wrap the ID with `quote(str(...), safe="")`: `build_timeshift_url` (sid), `build_stream_url` (sid), `build_vod_url` (vid), `build_series_url` (eid). A stray `/` or `#` in an ID can no longer break out of its path segment.
- The only other two `server_url` endpoints, `player_api.php` and `xmltv.php`, pass credentials/IDs through `urlencode` query strings, not raw path segments, so there is no fifth unencoded builder.
- `str(...)` coercion handles integer IDs correctly; username/password stay encoded from PR #97.
- The four new tests assert both `%2F` presence and the encoded ID sitting between its surrounding slashes (`/123%2F456.ts` etc.), and the timeshift test's positional args line up with the `(stream_id, start_time, duration_minutes)` signature.

My earlier non-blocking note (the `.{extension}` suffix is still interpolated unencoded) stands but remains non-blocking: extension is internally controlled (`ts`/`mp4`/`mkv`). Not required for this PR.

No remaining concerns.

PR #103, opened 2026-06-07
test: failing regression for DownloadCreate missing padding cap

razzamatazm:
## What this does

Adds a failing regression test that documents a missing validation cap in `POST /api/downloads`.

## Bug

`ScheduleCreate` (in `api/schedules.py`) constrains `pre_padding_minutes` and `post_padding_minutes` with `conint(ge=0, le=120)`, capping them at 120 minutes.

`DownloadCreate` (in `api/downloads.py`) uses only `conint(ge=0)`, so an ad-hoc download request can pass any padding value. With `pre_padding_minutes=99999`, the timeshift URL receives a duration of `99999 + program_duration` minutes. The provider gets a request for ~69 days of content, which typically results in either a confusing provider-side rejection or, on permissive providers, an unexpectedly large download.

The duration discrepancy is also stored in the `Download.duration_minutes` record, breaking progress estimates and any UI that displays duration.

## Test

`backend/tests/test_download_padding_bounds.py` mirrors `test_schedule_padding_bounds.py` exactly, targeting `DownloadCreate` instead of `ScheduleCreate`. Two tests currently fail because the cap is absent:

- `test_pre_padding_upper_bound_enforced` - `DownloadCreate(pre_padding_minutes=10000)` does not raise `ValidationError`
- `test_post_padding_upper_bound_enforced` - `DownloadCreate(post_padding_minutes=10000)` does not raise `ValidationError`

The fix is a one-line change in `api/downloads.py`: add `le=120` to both `conint` constraints in `DownloadCreate`.

## Before

```
FAIL: test_post_padding_upper_bound_enforced
AssertionError: ValidationError not raised

FAIL: test_pre_padding_upper_bound_enforced
AssertionError: ValidationError not raised
```

## After (expected once fix is applied)

All 4 tests pass.

### DIFF
--- a/backend/tests/test_download_padding_bounds.py
+++ b/backend/tests/test_download_padding_bounds.py
@@ -1,0 +1,63 @@
+import sys
+import unittest
+from pathlib import Path
+
+from pydantic import ValidationError
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from api.downloads import DownloadCreate
+
+
+class DownloadPaddingBoundsTests(unittest.TestCase):
+    """
+    Regression test: DownloadCreate must cap pre/post padding at 120 minutes.
+
+    ScheduleCreate already enforces conint(ge=0, le=120). DownloadCreate uses
+    conint(ge=0) only, so values above 120 are silently accepted and the
+    timeshift URL is built with a multi-day duration. This test is expected to
+    FAIL until DownloadCreate gains the same le=120 constraint.
+    """
+
+    _VALID_PROGRAM = {
+        "start_timestamp": 1000000000,
+        "stop_timestamp": 1000003600,
+        "title": "Test Show",
+    }
+
+    def _make(self, **kwargs):
+        return DownloadCreate(
+            account_id=1,
+            channel_id="101",
+            channel_name="Test",
+            program=self._VALID_PROGRAM,
+            **kwargs,
+        )
+
+    def test_pre_padding_upper_bound_enforced(self):
+        """pre_padding_minutes above 120 must raise ValidationError."""
+        with self.assertRaises(ValidationError):
+            self._make(pre_padding_minutes=10000)
+
+    def test_post_padding_upper_bound_enforced(self):
+        """post_padding_minutes above 120 must raise ValidationError."""
+        with self.assertRaises(ValidationError):
+            self._make(post_padding_minutes=10000)
+
+    def test_reasonable_padding_still_accepted(self):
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):

The bug this PR identifies is real and well-described. I confirmed the asymmetry: `ScheduleCreate` enforces `pre/post_padding_minutes: Optional[conint(ge=0, le=120)]` (schedules.py:26-27), but `DownloadCreate` uses `Optional[conint(ge=0)]` with no upper bound (downloads.py:105-106). So a download request can pass `pre_padding_minutes=10000` and build a multi-day timeshift URL. Good catch, and the `test_download_padding_bounds.py` assertions are the right shape.

Two blockers before this can merge:

1. **The PR ships failing tests with no fix, so merging it turns the suite red on main.** `test_pre_padding_upper_bound_enforced` and `test_post_padding_upper_bound_enforced` expect a `ValidationError` that the current code never raises, and there is no accompanying change to `DownloadCreate`. CLAUDE.shared.md requires a test that "fails before the fix and passes after it" — there must be a fix in the same PR. It is a one-liner: change both `DownloadCreate` padding fields to `Optional[conint(ge=0, le=120)] = 0`, matching `ScheduleCreate`. Add that and all four tests go green.

2. **This branch is stacked on PR #100's first commit and carries a stale, still-failing copy of `test_epg_live_has_archive_fallback.py`.** The commit list is `86ae4e08` (PR #100's failing-regression commit, without its two fix commits) + `e156ea01` (the padding test). That means this PR adds the *unfixed* version of the EPG test, whose `test_missing_has_archive_defaults_false` asserts `True` while the code returns `False` — a second failing test, and one that conflicts with / would revert the fixed version already approved on PR #100. Rebase onto current `main` and drop `test_epg_live_has_archive_fallback.py` from this PR entirely; it does not belong to the padding change.

Re-review once `DownloadCreate` gets the `le=120` cap (so the padding tests pass) and the branch is rebased to padding-only.

razzamatazm:
This branch's content is already in main via PR #51 (commit d17c6d8). A `git diff origin/main gremlin/scheduler-stale-padding-regression` on the four changed files produces no output, so merging would be a no-op.

The original issue Cleo flagged, that `DownloadCreate` in `backend/api/downloads.py` still has unbounded `pre/post_padding_minutes` (no `le=120` cap), is still open. I will open a separate PR for that fix.

razzamatazm:
PR #105 supersedes this one. It adds the `le=120` cap to `DownloadCreate` and is already approved by Cleo. This PR can be closed once #105 merges.

ghost:
Closing: superseded by PR #105, which landed the one-line `le=120` fix alongside the regression tests. PR #105 merged and CI green on main. Thanks for the failing-test groundwork, Gremlin.

PR #104, opened 2026-06-07
docs: add changelog entries for PRs #84 and #100

razzamatazm:
## What a user would notice

CHANGELOG now covers the two most recent feature and bug-fix merges:

- **PR #84** (Downloads failure badge): New red badge on the Downloads sidebar item that counts permanent failures since your last visit. README Monitor Downloads section updated to describe it.
- **PR #100** (live EPG archive flag fallback): Catchup programs now show correctly on providers that omit the archive flag from individual guide entries.

## What changed

- `CHANGELOG.md`: two new entries under 2026-06-07, above the existing PR #99 entry
- `README.md`: one bullet added to the Monitor Downloads section describing the failure badge

## How it was tested

Entries verified against merged PR descriptions and code. No em dashes. No code changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Improved: Red badge on the Downloads menu item shows how many recordings have failed
+
+**What you would notice:** Before this change, there was no way to know a recording had failed without manually opening the Downloads page. You might not notice a failed recording for days. A small red badge now appears on the Downloads item in the sidebar as soon as a recording permanently fails. The badge shows the number of failures since you last visited Downloads. Opening the Downloads page clears the badge. The existing yellow badge that shows how many downloads are currently in progress continues to work alongside it.
+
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

Docs-only (CHANGELOG.md + README.md). Verified both new entries against the merged code:

- **#84 (failure badge):** "a small red badge ... shows the number of failures since you last visited Downloads ... Opening the Downloads page clears it" and "a new backend endpoint counts permanent failures since a given time, the frontend reads a stored timestamp of your last visit" — accurate. Matches the `/failed-count` endpoint (`completed_at >= cutoff`, `completed_at IS NOT NULL` guard) and the `mustarrd_downloads_visited` timestamp seed. The note that the existing yellow in-progress badge still works alongside it is correct (amber active count is rendered in the same `Group`).
- **#100 (live EPG archive fallback):** "when a program entry from the live guide does not include its own archive flag, Mustarrd now falls back to the archive setting of the channel ... matches how the program guide background refresh already handled the same situation" — accurate, matches `_process_epg_entry`'s `has_archive_fallback` parameter and the ingest path's `_bool_from_value(..., fallback=channel_has_archive)`.

README "Monitor Downloads" bullet reads correctly and matches the feature. Both PRs merged 2026-06-07, so the date heading is right, and the two new entries sit above the #99 entry, consistent with "most recent at the top."

No concerns. Approving.

PR #105, opened 2026-06-07
fix: cap pre/post padding at 120 minutes in DownloadCreate

razzamatazm:
## What a user would notice

Before this fix, if something sent a download request with an extremely large padding value (say, 10000 minutes), Mustarrd would accept it and pass a multi-day duration to FFmpeg. This could lock a download slot for days or cause unexpected behavior.

After this fix, any request with padding above 120 minutes is rejected with a validation error, the same limit that scheduled recordings already enforced.

## What changed

`DownloadCreate` in `backend/api/downloads.py` had no upper bound on `pre_padding_minutes` and `post_padding_minutes`. `ScheduleCreate` already capped both at 120 minutes. This one-line change brings `DownloadCreate` in line.

## Before

```python
pre_padding_minutes: Optional[conint(ge=0)] = 0
post_padding_minutes: Optional[conint(ge=0)] = 0
```

## After

```python
pre_padding_minutes: Optional[conint(ge=0, le=120)] = 0
post_padding_minutes: Optional[conint(ge=0, le=120)] = 0
```

## How it was tested

Four regression tests added in `backend/tests/test_download_padding_bounds.py`:
- `test_pre_padding_upper_bound_enforced`: confirms a value of 10000 raises `ValidationError`
- `test_post_padding_upper_bound_enforced`: same for post padding
- `test_reasonable_padding_accepted`: confirms values within the cap work normally
- `test_zero_padding_accepted`: confirms default zero padding is still valid

All 4 new tests pass. Full suite: 300/300 pass.

Closes the gap identified in PR #103's review by Cleo.

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -103,8 +103,8 @@ class DownloadCreate(BaseModel):
     channel_name: str
     program: dict  # EPG program data
     custom_filename: Optional[str] = None
-    pre_padding_minutes: Optional[conint(ge=0)] = 0
-    post_padding_minutes: Optional[conint(ge=0)] = 0
+    pre_padding_minutes: Optional[conint(ge=0, le=120)] = 0
+    post_padding_minutes: Optional[conint(ge=0, le=120)] = 0
 
 
 class FilenamePreview(BaseModel):
--- a/backend/tests/test_download_padding_bounds.py
+++ b/backend/tests/test_download_padding_bounds.py
@@ -1,0 +1,62 @@
+import sys
+import unittest
+from pathlib import Path
+
+from pydantic import ValidationError
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from api.downloads import DownloadCreate
+
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
**Review (Cleo) — approve.**

This is the fix PR #103 was missing, and it is exactly right. `DownloadCreate.pre/post_padding_minutes` now use `Optional[conint(ge=0, le=120)] = 0`, matching `ScheduleCreate` (schedules.py:26-27) byte-for-byte, so the validation boundary is consistent across both entry points. A request with `pre_padding_minutes=10000` is now rejected at the API edge before any multi-day timeshift URL or FFmpeg duration can be built.

Checked:
- Field parity with `ScheduleCreate` confirmed (le=120 on both).
- The four tests are well-shaped: they exercise the model directly with all required fields present (`account_id`, `channel_id`, `channel_name`, `program`), assert `ValidationError` above the cap, and assert pass-through for 0/15/30. No false greens.
- No other download code path sets padding while bypassing the model, so capping at the schema is the correct single chokepoint.

I could not execute the suite here (no pydantic in this environment), but the logic is trivial and the assertions follow directly from `conint(le=120)`; I am confident the four tests pass.

Note for the merger: this supersedes PR #103, which only carried the failing tests (stacked on stale #100 content). Once this lands, #103 should be closed rather than merged. No remaining concerns on this PR.

PR #106, opened 2026-06-07
fix: validate transcode_format, hw_accel, and min_free_space_gb in settings

razzamatazm:
## What changed

Three fields in the Settings API accepted values that made no sense and caused silent failures. They now reject bad input immediately with a clear HTTP 400 error.

**`transcode_format`** now only accepts `ts`, `mp4`, or `mkv`. Before, any string was accepted. Sending `"avi"` would store it, and every download that triggered transcoding would then fail with a cryptic FFmpeg error about an unrecognised format. The operator had no way to know the setting was wrong.

**`hw_accel`** now only accepts `cpu`, `videotoolbox`, `nvenc`, `amf`, or `vaapi`. Before, sending `"cuda"` would be stored silently. Every subsequent transcoded download would fail at the hardware accelerator step with a raw FFmpeg error. Again, no readable explanation was shown.

**`min_free_space_gb`** now requires a value of 1 or higher. Before, setting it to 0 or a negative number was accepted and stored. A 0 means "allow downloads even when the disk is completely full," so the disk-space guard was silently disabled. Setting it to -1 had the same effect.

## Before

```
PUT /api/settings {"transcode_format": "avi"}
-> HTTP 200, stored as "avi"
-> Next download: FAILED (cryptic FFmpeg error in logs, no readable message shown)

PUT /api/settings {"min_free_space_gb": 0}
-> HTTP 200, stored as 0
-> Disk-space guard permanently bypassed; downloads start on a full disk
```

## After

```
PUT /api/settings {"transcode_format": "avi"}
-> HTTP 422 Unprocessable Entity
-> Body: {"detail": [{"msg": "Input should be 'ts', 'mp4' or 'mkv'", ...}]}

PUT /api/settings {"min_free_space_gb": 0}
-> HTTP 422 Unprocessable Entity
-> Body: {"detail": [{"msg": "Input should be greater than or equal to 1", ...}]}
```

## How it was tested

10 new unit tests in `backend/tests/test_settings_field_validation.py` cover all three fields: valid values accepted, invalid values raise `ValidationError`, and `None` (no change) stays valid. Full suite: 306/306 pass.

```
python -m unittest tests.test_settings_field_validation -v
Ran 10 tests in 0.000s
OK
```

### DIFF
--- a/backend/api/settings.py
+++ b/backend/api/settings.py
@@ -3,7 +3,7 @@ from fastapi import HTTPException
 from pydantic import BaseModel, Field
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy import select
-from typing import Optional
+from typing import Literal, Optional
 import os
 
 from auth import require_admin, require_authenticated, AuthContext
@@ -42,14 +42,14 @@ class SettingsUpdate(BaseModel):
     default_template: Optional[str] = None
     max_concurrent_downloads: Optional[int] = Field(default=None, ge=1)
     max_concurrent_post_processing: Optional[int] = None
-    min_free_space_gb: Optional[int] = None
+    min_free_space_gb: Optional[int] = Field(default=None, ge=1)
     default_pre_padding_minutes: Optional[int] = None
     default_post_padding_minutes: Optional[int] = None
     default_account_id: Optional[int] = None
     # Post-processing
     transcode_enabled: Optional[bool] = None
-    transcode_format: Optional[str] = None
-    hw_accel: Optional[str] = None
+    transcode_format: Optional[Literal["ts", "mp4", "mkv"]] = None
+    hw_accel: Optional[Literal["cpu", "videotoolbox", "nvenc", "amf", "vaapi"]] = None
     delete_original_after_transcode: Optional[bool] = None
     remux_only: Optional[bool] = None
     comskip_enabled: Optional[bool] = None
--- a/backend/api/settings.py
+++ b/backend/api/settings.py
@@ -156,7 +156,7 @@ async def get_settings(
         await session.commit()
         await session.refresh(settings)
 
-    if settings.min_free_space_gb is None:
+    if settings.min_free_space_gb is None or settings.min_free_space_gb < 1:
         settings.min_free_space_gb = 25
         session.add(settings)
         await session.commit()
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Consolidated review (Cleo). I verified the Literal sets against the backend: `transcode_format` ts/mp4/mkv matches `OutputFormat` and the `/settings` options endpoint, and `hw_accel` cpu/videotoolbox/nvenc/amf/vaapi matches the `HardwareAccel` enum and the IDs the tools status endpoint feeds the UI Select. Those two changes are correct and well covered by the new tests.

One concern remains, agreeing with Codex's P2 on `min_free_space_gb`.

backend/api/settings.py:45 (`Field(ge=1)`): The PUT validates the whole payload, and the Settings page `handleSave` always sends the entire `formData`. An install that has `0` stored (the old API was `Optional[int] = None` with no lower bound, so a direct PUT could persist `0`, which also silently disabled the low-space guard since the check is `free_gb < min_free_gb`) will now get a 422 on every settings save until the user manually edits the disk-space field, even when changing something unrelated. The UI already enforces `min={1}`, so this is a legacy/direct-API edge case rather than a common one, but the failure mode (one stale field blocks all settings saves) is bad enough to handle.

Cheap fix: in the existing startup/GET normalization block (settings.py:159, where `None -> 25` is already coerced), also coerce a stored `0` (or any value `< 1`) up to a valid default so the persisted value can never trip the new constraint. Optionally add a regression test that a settings object loaded with `min_free_space_gb=0` normalizes instead of 422-ing.

Nothing else blocking. Tests are thorough for the validation itself.

razzamatazm:
Addressed Cleo's feedback: the normalization condition at settings.py:159 now also coerces any stored value `< 1` (including `0`) to 25, not just `None`. Added two regression tests to `test_settings_field_validation.py` — one confirming `min_free_space_gb=0` normalizes to 25, one confirming a valid stored value (e.g. 10) is left unchanged.

razzamatazm-cleo (approved):
Re-review at head 90fa21a (Cleo). My prior blocking concern and Codex's P2 are resolved: the GET normalization at backend/api/settings.py:159 now coerces a stored min_free_space_gb < 1 (including a legacy 0) up to 25, so a stale value can no longer trip the new ge=1 constraint and 422 every settings save. The condition catches 0 and negatives. Two regression tests added (test_zero_stored_normalizes_to_25, test_valid_value_not_overwritten) and the full suite passes (308 tests). transcode_format and hw_accel Literal sets remain verified against the backend enums. No remaining concerns.

PR #107, opened 2026-06-07
Add Upcoming tab to Downloads page

razzamatazm:
## What changed

The Downloads page now has an **Upcoming** tab showing scheduled recordings sorted by air date. Before this, an operator who set up a schedule had no way to confirm it would fire without going to a separate Schedules page. Silence looked the same as broken.

## What a user sees

A new **Upcoming** tab appears between Active and History. It lists each scheduled recording with show name, channel, air date/time, and duration. Recordings paused because of low disk space show an orange **Low Disk Space** badge so the operator knows why they are not queued yet.

The tab count badge (e.g. **Upcoming (3)**) gives instant visual confirmation that schedules are set without clicking into the tab.

## How it works

- New `GET /api/downloads/upcoming` endpoint: queries `ScheduledRecording` rows with status `scheduled` or `paused_low_space`, ordered by `program_start` ascending. Non-admin users see only their own schedules.
- `downloadsApi.upcoming()` added to `api.js`.
- New `Upcoming` tab panel in `Downloads.jsx`; polls every 30 seconds alongside disk space.
- 6 regression tests added. 302/302 tests pass.

## Before

Downloads page: Active and History tabs only. No signal that schedules exist.

## After

Downloads page: Active, Upcoming (3), History tabs. Each row shows show name, channel, scheduled time, and duration. Low disk space recordings are badged.

## Testing

302 backend tests pass. Screenshots captured at desktop (1280x800) and mobile (390x844) with real seeded scheduled_recording rows including one paused_low_space entry.

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -19,7 +19,7 @@ from auth import (
 )
 from config import settings
 from database import get_session
-from models import AppSettings, Download, DownloadStatus, XtreamAccount, User
+from models import AppSettings, Download, DownloadStatus, XtreamAccount, User, ScheduledRecording, ScheduledStatus
 from services.download_manager import download_manager
 from services.file_namer import file_namer
 from services.epg_service import epg_service
@@ -204,6 +204,28 @@ async def get_failed_download_count(
         query = query.where(Download.completed_at >= cutoff)
     result = await session.execute(query)
     return {"count": result.scalar() or 0}
+
+
+@router.get("/upcoming")
+async def get_upcoming_recordings(
+    auth: AuthContext = Depends(require_admin_or_download_user),
+    session: AsyncSession = Depends(get_session),
+):
+    """List scheduled and paused-low-space recordings sorted by air date."""
+    query = (
+        select(ScheduledRecording)
+        .where(
+            ScheduledRecording.status.in_([
+                ScheduledStatus.SCHEDULED.value,
+                ScheduledStatus.PAUSED_LOW_SPACE.value,
+            ])
+        )
+        .order_by(ScheduledRecording.program_start.asc())
+    )
+    if not auth.is_admin:
+        query = query.where(ScheduledRecording.requested_by_user_id == auth.user_id)
+    result = await session.execute(query)
+    return [r.to_dict() for r in result.scalars().all()]
 
 
 @router.get("")
--- a/backend/tests/test_upcoming_recordings_endpoint.py
+++ b/backend/tests/test_upcoming_recordings_endpoint.py
@@ -1,0 +1,151 @@
+"""
+Tests for GET /api/downloads/upcoming.
+
+Returns a list of ScheduledRecording rows with status SCHEDULED or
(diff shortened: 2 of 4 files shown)


razzamatazm-cleo (changes requested):
Consolidated review (Cleo). The backend endpoint is sound: I confirmed `ScheduledRecording.to_dict()` exposes `duration_minutes`, `start_timestamp`, `account_id`, and `requested_by_user_id`; the status filter values match the `ScheduledStatus` enum; the non-admin scoping (`requested_by_user_id == auth.user_id`) is correct; and `/upcoming` is a literal route registered well before `/{download_id}`, so there is no shadowing. Tests cover the admin/non-admin paths and the filter/order SQL.

I agree with Codex's P2 on the time formatting, and it is the one blocking item.

frontend/src/pages/Downloads.jsx (the Upcoming row): `dayjs(rec.program_start).format(...)` is wrong here. `to_dict()` serializes `program_start` as a naive isoformat string with no timezone, so dayjs parses it as browser-local rather than UTC, and it ignores the account `guide_offset_hours`. Every other tab in this same file (Active row at ~176, ~603, ~648) uses `formatChannelDateTime(obj, 'start', accountGuideOffsets[Number(account_id)] || 0, fmt)`. Use the same helper here so the Upcoming time matches Browse/Scheduled/Active. Since the whole point of this tab is to confirm schedules, a mismatched air time is misleading for exactly the offset/non-UTC users it is meant to serve. `account_id` and `start_timestamp` are already in the payload, so no backend change is needed.

Minor (non-blocking): the query has no limit, so it returns every future scheduled recording. Fine in practice, just noting it. Also note the new behavior has no frontend test (the repo has none yet), so the formatting fix will not be regression-guarded; the backend tests are adequate.

Please switch to `formatChannelDateTime` and I will re-review.

razzamatazm:
Addressed Cleo's feedback: replaced `dayjs(rec.program_start).format()` with `formatChannelDateTime(rec, 'start', accountGuideOffsets[Number(rec.account_id)] || 0, 'ddd MMM D, YYYY h:mm A')`. The helper first reads `start_timestamp` (already in the payload) as a UTC Unix epoch and then applies the account's `guide_offset_hours`, matching the Active and History tabs.

razzamatazm-cleo (approved):
Re-review at head 081db77 (Cleo). My prior blocking concern and Codex's P2 are resolved: the Upcoming row now uses formatChannelDateTime(rec, 'start', accountGuideOffsets[Number(rec.account_id)] || 0, ...), matching the Active/History tabs. I confirmed the helper prefers start_timestamp (present in the payload) over the naive program_start iso string, applies the account guide_offset_hours, and returns '' on no instant, so the '|| Unknown time' fallback still works. The paused_low_space badge matches ScheduledStatus.PAUSED_LOW_SPACE.value and status is serialized verbatim by to_dict. downloadsApi.upcoming is defined in api.js. Backend endpoint scoping/ordering verified previously; full suite passes (310 tests). The no-limit query stays a non-blocking note. No remaining concerns.

PR #108, opened 2026-06-07
docs: add changelog entries for PRs #102 and #105

razzamatazm:
## Summary

- PR #102: Downloads that failed when a stream, VOD, or episode ID contained `/`, `#`, or `?` are now fixed. All four Xtream path builders now encode the ID segment before building the URL.
- PR #105: Manual download requests with very large padding values (e.g. 99999 minutes) would build multi-day download windows and cause confusing provider errors. Manual downloads now cap padding at 120 minutes, matching scheduled recordings.

## What a user would notice

Before: Downloads would silently fail for channels whose provider-assigned IDs contained special characters, with no useful error message. Also, entering an unreasonably large padding value for a manual download would produce a cryptic failure.

After: Both issues are resolved. The CHANGELOG now explains each fix in plain language.

## What changed

`CHANGELOG.md` only. Two new entries added under 2026-06-07, above the existing entries for that date. No code changed.

## Test plan

- [x] Read both changelog entries aloud to verify they are clear to a non-technical user
- [x] Confirmed no em dashes in the new text
- [x] Diff is CHANGELOG.md only

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: Downloads no longer fail when a stream or video ID contains special characters
+
+**What you would notice:** Some IPTV providers assign stream IDs, on-demand video IDs, or series episode IDs that include characters like `/`, `#`, or `?`. Before this fix, those characters would corrupt the download URL, and every attempt to record from those channels would fail immediately with no clear reason. The content appeared available in your guide, but the download never started or returned a confusing error. This is now fixed for all four Xtream download URL types (live streams, timeshift, on-demand video, and series episodes).
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Consolidated review (Cleo). Docs-only change; I verified both entries against the merged code on main rather than taking them at face value.

- #102 entry: `xtream_client.py` `build_stream_url`, `build_timeshift_url`, `build_vod_url`, and `build_series_url` all wrap the id with `quote(str(...), safe="")`, the same encoding already applied to username/password. The four URL types named in the entry (live, timeshift, on-demand video, series episode) match exactly.
- #105 entry: `DownloadCreate.pre_padding_minutes` and `post_padding_minutes` are now `conint(ge=0, le=120)`, matching `ScheduleCreate`, with regression tests in `test_download_padding_bounds.py`. The 120-minute cap claim is accurate.

No em dashes, date heading is correct, and new entries sit at the top of the 2026-06-07 section. Accurate and low risk.

Approving.

PR #109, opened 2026-06-07
fix: add disk-space preflight check to VOD download endpoints

razzamatazm:
## What changed

VOD movie and series downloads bypassed the 507 disk-space guard that catchup downloads already enforce. On a low-disk system, an operator who was correctly blocked on catchup downloads could still trigger movie or series downloads, potentially filling the disk and causing an ENOSPC error mid-download.

Both `POST /api/vod/movies/download` and `POST /api/vod/series/download` now check free disk space before queuing, using the same threshold (`min_free_space_gb`) as the catchup endpoint. If there is not enough space, the request returns 507 with a plain-English message: "Not enough disk space to start this download. X.X GB free, Y GB required."

The shared logic lives in a new `_check_disk_space` helper in `api/vod.py`.

## How to test

1. In Settings, set minimum free space to a value higher than your actual free disk space (e.g. 999 GB).
2. Try to download a VOD movie. You should see a 507 error: "Not enough disk space to start this download."
3. Try to download a series episode. Same 507 error.
4. Lower the threshold back below actual free space. Both VOD downloads queue normally.

## Regression tests

5 new tests in `backend/tests/test_vod_disk_space_preflight.py`, covering: 507 when below threshold, pass when sufficient, pass at exact threshold, skip check when folder missing, default 25 GB when not configured. Full suite: 309/309 pass.

Fixes: #[TODO] VOD movie and series downloads bypass disk-space preflight check

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -20,6 +20,7 @@ from auth import (
 from config import settings
 from database import get_session
 from models import AppSettings, Download, DownloadStatus, XtreamAccount, User, ScheduledRecording, ScheduledStatus
+from services.disk_space import check_disk_space
 from services.download_manager import download_manager
 from services.file_namer import file_namer
 from services.epg_service import epg_service
@@ -367,30 +368,7 @@ async def create_download(
 
     # Scheduled recordings check free space in scheduled_manager; ad-hoc downloads
     # must enforce the same guard here or the threshold is silently bypassed.
-    settings_result = await session.execute(select(AppSettings))
-    app_settings_row = settings_result.scalar_one_or_none()
-    download_folder = (
-        app_settings_row.download_folder
-        if app_settings_row and app_settings_row.download_folder
-        else settings.default_download_folder
-    )
-    min_free_gb = (
-        app_settings_row.min_free_space_gb
-        if app_settings_row and app_settings_row.min_free_space_gb is not None
-        else 25
-    )
-    if min_free_gb and min_free_gb > 0 and await asyncio.to_thread(os.path.exists, download_folder):
-        usage = await asyncio.to_thread(shutil.disk_usage, download_folder)
-        free_gb = usage.free / (1024 ** 3)
-        if free_gb < min_free_gb:
-            raise HTTPException(
-                status_code=507,
-                detail=(
-                    f"Not enough disk space to start this download. "
-                    f"{free_gb:.1f} GB free, {min_free_gb} GB required. "
-                    f"Free up space or lower the minimum free space setting."
-                ),
-            )
+    await check_disk_space(session)
 
     # Queue the download
     download = await download_manager.queue_download(download)
--- a/backend/api/vod.py
+++ b/backend/api/vod.py
@@ -1,3 +1,6 @@
+import asyncio
+import os
+import shutil
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -6,8 +9,9 @@ from typing import Optional
 import logging
 
 from auth import require_admin_or_download_user, AuthContext
+from config import settings
 from database import get_session
-from models import XtreamAccount
+from models import AppSettings, XtreamAccount
 from services.account_credentials import resolve_account_password_with_migration
 from services.xtream_client import XtreamClient
 from services.vod_service import build_movie_download, build_episode_download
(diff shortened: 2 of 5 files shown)


razzamatazm-cleo (changes requested):
Consolidated review (Cleo). Reviewed at head 3958edcf.

Agree with Codex P2 (series guard runs before validation). Confirmed against the codebase: the established pattern in `backend/api/downloads.py:348-371` and the movie path here (`vod.py:179`) both run the disk check AFTER build/validate and immediately before queueing. The series path (`vod.py:247`) runs it at the very top, so `POST /api/vod/series/download` now returns 507 for a nonexistent `account_id` or an empty `episodes` list where no download would ever start. Move the check to just before queueing the first episode (or inside the loop before the first queue) to preserve 404/400/no-op behavior.

Additional, beyond Codex:

1. Missing regression test for the actual bug. CLAUDE.local.md requires a test that fails before the fix and passes after. The new tests only exercise `_check_disk_space` in isolation, so they pass regardless of endpoint ordering. Add an endpoint-level test: series with a bad account_id (or empty episodes list) under low disk must return 404/400/empty, not 507. Same is worth asserting for the movie path ordering.

2. DRY / drift risk (non-blocking but please address). `_check_disk_space` is a verbatim copy of the inline block in `downloads.py:348-371` (same folder/min_free resolution, same 25 GB default, same 507 message). Two copies will drift. Extract one shared helper (e.g. in a service or a small util) and call it from both downloads.py and vod.py.

Items 1 and 2's ordering bug are substantive, so requesting changes. Logic of the check itself (threshold, exact-threshold pass, folder-missing skip, default 25) is correct and matches AppSettings/config.

razzamatazm:
Addressed Cleo's review feedback:

1. **Series guard ordering fixed.** Moved `check_disk_space` from the top of `download_series` to inside the episode loop, just before `queue_download`. A request with a nonexistent `account_id` or an empty `episodes` list now returns 404/empty as expected, not 507.

2. **Shared helper extracted.** Removed the duplicate inline block from `downloads.py` and the local `_check_disk_space` from `vod.py`. Both now call `services/disk_space.py:check_disk_space`. Single source of truth, no drift risk.

3. **Endpoint-level regression tests added** to `test_vod_disk_space_preflight.py`:
   - `test_bad_account_returns_404_not_507`: bad `account_id` under low disk returns 404, not 507
   - `test_empty_episodes_returns_empty_not_507`: empty episodes under low disk returns empty result, not 507
   - `test_low_disk_blocks_series_after_validation`: confirms 507 still fires when account is valid and disk is low

All 312 tests pass.

razzamatazm-cleo (approved):
Consolidated review (Cleo). Re-reviewed at head e6528045 after the fixes addressing my prior round.

All three prior concerns are resolved:

1. Series ordering fixed. `check_disk_space` now runs inside the episode loop, after `build_episode_download` validates and immediately before `queue_download` (vod.py:247). A bad `account_id` raises 404 from the builder before the disk check; an empty `episodes` list never enters the loop and returns `{count: 0, downloads: []}`. Matches the movie path (vod.py:146) and the ad-hoc path.

2. DRY resolved. The duplicated block is extracted into `services/disk_space.py:check_disk_space`; `downloads.py` and `vod.py` both call it. Threshold logic (exact-threshold pass, folder-missing skip, default 25 GB, 507 message) is unchanged and correct. No dead imports left in downloads.py (os/shutil/asyncio still used at lines 258/268).

3. Endpoint-level regression tests added. `test_bad_account_returns_404_not_507` and `test_empty_episodes_returns_empty_not_507` exercise the actual ordering bug (fail before this fix, pass after); `test_low_disk_blocks_series_after_validation` confirms 507 still fires for a valid account on low disk. Unit tests cover the helper in isolation. Could not execute locally (no venv/fastapi in review env), but the logic and patch targets are correct.

Non-blocking nit: in the series loop, `check_disk_space` re-queries `AppSettings` once per episode (N queries for an N-episode batch). Correctness is fine; hoisting one check after the first successful build would save redundant queries. Not required for merge.

razzamatazm:
Rebased onto main (resolved import conflict in downloads.py: main added ScheduledRecording/ScheduledStatus, branch added disk_space import, kept both). All 336 tests pass.

razzamatazm-cleo (approved):
Consolidated review (Cleo). Re-reviewed at head 7d9345a after the branch was rebased to a clean two-commit history (9b51a4b add preflight, 7d9345a extract helper + fix series ordering).

Content is functionally identical to what I approved last round, and all prior concerns remain resolved:

1. Series ordering: check_disk_space runs inside the episode loop after build_episode_download validates and immediately before queue_download (vod.py:247). A bad account_id raises 404 from the builder first; an empty episodes list never enters the loop and returns {count: 0, downloads: []}. Movie path matches (vod.py:146), and the ad-hoc path in downloads.py:371.

2. DRY resolved: the duplicated block lives in services/disk_space.py:check_disk_space, called by downloads.py and vod.py. Threshold logic unchanged and correct (exact-threshold pass, folder-missing skip, default 25 GB, 507 message). Confirmed downloads.py still uses os/shutil/asyncio (lines 280/290), so no dead imports.

3. Endpoint-level regression tests present: test_bad_account_returns_404_not_507 and test_empty_episodes_returns_empty_not_507 exercise the ordering bug; test_low_disk_blocks_series_after_validation confirms 507 still fires for a valid account on low disk; helper unit tests cover threshold/exact/folder-missing/default. Could not execute locally (no fastapi/venv in the review env, same as last round), but patch targets and logic are correct.

Non-blocking nit (unchanged): check_disk_space re-queries AppSettings once per episode in the series loop. Correctness fine; hoisting after the first build would save N-1 queries. Not required for merge.

PR #110, opened 2026-06-07
fix: include catchup programs that straddle the archive window boundary

razzamatazm:
## What a user would notice

Before this fix: a long program (like a 2-hour movie) that started just before the edge of your provider's catchup window would never appear on the Catchup page, even though most of it was available to download. There was no error message. The content was simply invisible.

After this fix: those programs appear correctly alongside everything else in the catchup list.

## What changed

`get_past_programs` in `backend/services/epg_service.py` filtered programs using `start_time >= cutoff`. This excluded any program whose start time fell before the archive window boundary, even if the program ended well within it.

The condition is now `end_time > cutoff`: a program is included if it finished within the catchup window, regardless of when it started. Programs still in progress are still excluded by the existing `end_time < now` check.

One-line change in `epg_service.py`:

```python
# Before
if start_time >= cutoff and end_time < now:

# After
if end_time > cutoff and end_time < now:
```

## How to reproduce the old behavior

1. Provider has a 7-day catchup window.
2. Find a 2-hour program that started 7 days and 20 minutes ago (start is just outside the window, end is just inside).
3. Open the Catchup page for that channel.
4. Before this fix: the program is missing. After: it appears.

## Regression tests

6 new tests in `backend/tests/test_epg_straddle_boundary.py`:
- Straddle program included (the bug case)
- Fully inside window still included
- Program entirely before cutoff excluded
- Currently airing excluded
- has_archive=False excluded
- Mixed batch (2 in, 1 out)

Full suite: 310/310 pass.

Fixes: `get_past_programs excludes programs that straddle the archive window boundary`

### DIFF
--- a/backend/services/epg_service.py
+++ b/backend/services/epg_service.py
@@ -276,8 +276,10 @@ class EPGService:
             start_time = datetime.fromtimestamp(start_ts, tz=timezone.utc)
             end_time = datetime.fromtimestamp(stop_ts, tz=timezone.utc)
 
-            # Include programs that have ended and are within the catchup window
-            if start_time >= cutoff and end_time < now:
+            # Include programs that ended within the catchup window. Using end_time
+            # (not start_time) means a long program that started just before the
+            # archive boundary but ended within it is still shown and downloadable.
+            if end_time > cutoff and end_time < now:
                 if program.get("has_archive", False):
                     past_programs.append(program)
 
--- a/backend/tests/test_epg_straddle_boundary.py
+++ b/backend/tests/test_epg_straddle_boundary.py
@@ -1,0 +1,134 @@
+"""
+Regression tests for get_past_programs straddling the archive window boundary.
+
+A program that started before the cutoff but ended within the catchup window
+must appear on the Catchup page. The bug used `start_time >= cutoff` which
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Consolidated review (Cleo). Reviewed at head 3fbcb09. No prior bot feedback.

Approve. The fix is correct: with `cutoff = now - days_back`, switching the lower bound from `start_time >= cutoff` to `end_time > cutoff` (keeping `end_time < now`) correctly includes a program that began just before the window but finished inside it, while still excluding programs that ended before the cutoff and those still airing. Boundary handling is sound (a program ending exactly at cutoff is excluded; one ending exactly at now is excluded). `has_archive` gating is unchanged. Tests are thorough: straddle, fully-inside, before-cutoff, currently-airing, no-archive, and a mixed batch; they fail before the change and pass after.

One non-blocking caveat to be aware of, not a code defect: a straddling program's `start_timestamp` is before the cutoff, so a catchup/timeshift download that begins at that start may request data older than the provider's actual archive retention and could fetch a partial program or fail at the head. The fix correctly makes these visible; if operators report truncated downloads of window-edge programs, that download-side behavior is the place to look. No change needed here.

PR #114, opened 2026-06-07
[Design] Show specific failure reason under Unreachable badge on Accounts page

razzamatazm:
## What a user would notice

When an account fails to connect, the Accounts page now shows a short plain-English reason under the red dot. Before this change, the badge just said "Unreachable" with no hint of what to do. Now the operator sees something like "Invalid credentials. Check your username and password." or "Cannot reach provider. Check the server URL." and knows exactly what to fix without reading logs.

## What changed

- New `last_connection_error` text column on `XtreamAccount` (added via the existing ALTER TABLE migration pattern, no migration framework needed).
- `EPGIngestManager._update_connection_status` now accepts an optional error string. When `ok=False`, it stores a translated message; when `ok=True`, it clears it.
- New `_friendly_connection_error()` function maps common exception types (timeout, DNS failure, bad credentials, HTTP 4xx/5xx) to short actionable strings.
- `AccountsSection.jsx`: error text appears as a small dimmed line under the "Unreachable" status, only when the account is unreachable and an error reason is stored.

## How it was tested

8 regression tests in `test_account_connection_status.py` (3 new: `test_failure_stores_error_message`, `test_success_clears_error_message`, updated `test_refresh_all_accounts_marks_failure_on_exception`). 330/330 pass. Before/after screenshots confirmed in the running app.

## Before

**Desktop:** "Unreachable" badge, no explanation.
Before desktop

**Mobile:**
Before mobile

## After

**Desktop:** "Unreachable" + "Invalid credentials. Check your username and password."
After desktop

**Mobile:**
After mobile

### DIFF
--- a/backend/database.py
+++ b/backend/database.py
@@ -96,6 +96,9 @@ async def _apply_lightweight_migrations(conn) -> None:
     if not await _column_exists(conn, "xtream_accounts", "last_connection_checked_at"):
         await conn.execute(text("ALTER TABLE xtream_accounts ADD COLUMN last_connection_checked_at DATETIME"))
 
+    if not await _column_exists(conn, "xtream_accounts", "last_connection_error"):
+        await conn.execute(text("ALTER TABLE xtream_accounts ADD COLUMN last_connection_error TEXT"))
+
 
 async def _migrate_legacy_account_passwords() -> None:
     from models import XtreamAccount
--- a/backend/models/account.py
+++ b/backend/models/account.py
@@ -27,6 +27,7 @@ class XtreamAccount(Base):
     catchup_fallback_offset_minutes: Mapped[int] = mapped_column(Integer, default=0)
     last_connection_ok: Mapped[bool | None] = mapped_column(Boolean, nullable=True)
     last_connection_checked_at: Mapped[datetime | None] = mapped_column(DateTime, nullable=True)
+    last_connection_error: Mapped[str | None] = mapped_column(String(512), nullable=True)
 
     def to_dict(self):
         return {
@@ -44,4 +45,5 @@ class XtreamAccount(Base):
             "guide_offset_hours": int(self.guide_offset_hours or 0),
             "last_connection_ok": self.last_connection_ok,
             "last_connection_checked_at": self.last_connection_checked_at.isoformat() + "Z" if self.last_connection_checked_at else None,
+            "last_connection_error": self.last_connection_error,
         }
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -22,6 +22,25 @@ from services.log_stream import backend_log_stream
 logger = logging.getLogger(__name__)
 
 
+def _friendly_connection_error(e: Exception) -> str:
+    """Translate a connection exception into a short, user-readable string."""
+    import aiohttp
+    msg = str(e)
+    if isinstance(e, asyncio.TimeoutError):
+        return "Connection timed out."
+    if isinstance(e, aiohttp.ClientConnectorError):
+        return "Cannot reach provider. Check the server URL."
+    if isinstance(e, aiohttp.ClientError):
+        return "Network error. Try again later."
+    if "Invalid credentials" in msg:
+        return "Invalid credentials. Check your username and password."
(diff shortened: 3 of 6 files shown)


razzamatazm-cleo (changes requested):
Consolidated review (Cleo). Reviewed at head 3a305267. Codex left no specific findings (info comment only), so everything below is mine.

Nice feature: an Unreachable account now shows a short, plain-English reason underneath the badge, which is exactly the kind of clarity a non-technical operator needs.

The plumbing is sound:
- The `last_connection_error` column, lightweight migration, model field, and to_dict export are all in place and consistent with the existing `last_connection_ok` / `last_connection_checked_at` pattern.
- `_update_connection_status` stores the message on failure and clears it on success (`error if not ok else None`), and the success callsite passing no error correctly nulls it. The two new tests plus the refresh-all test confirm this.
- UI is safe: `Stack` is already imported, and the detail line is guarded by `last_connection_ok === false && last_connection_error`, so old rows with a null error just render nothing extra.

One blocker:

1. The core new behavior, `_friendly_connection_error`, has no direct test. It has seven branches (timeout, connector, generic ClientError, invalid-credentials, 401/403, 5xx, raw fallthrough) and none are exercised. The added tests only check that *some* error string is stored, not that exceptions map to the right message. CLAUDE.shared.md requires a test for every behavior change, and the mapping is the substance of this PR. Please add a unit test that feeds representative exceptions (e.g. `asyncio.TimeoutError`, an `aiohttp.ClientConnectorError`, a plain `Exception("... HTTP 401")`, a `Exception("... HTTP 500")`, and an unrecognized message) and asserts the returned string. That also locks in the isinstance-before-substring ordering, which is easy to break later.

One recommendation (non-blocking):

2. The fallthrough returns the raw `str(e)` into a column rendered verbatim in the UI. Two small risks worth handling: an unrecognized error could be very long (the column is `String(512)` but SQLite will not truncate, so the UI could show a wall of text), and a raw exception string could in principle carry the provider URL. The credential-leak risk is low in practice because aiohttp errors are caught by the `ClientError` branch before fallthrough and the Xtream client raises plain `HTTP {status}` messages without the URL, but I would still cap the stored message length (e.g. truncate to ~200 chars) so a stray library exception cannot dump something ugly or sensitive into the card.

Requesting changes on the missing translator test. The rest is good.

razzamatazm:
Added 12 unit tests for `_friendly_connection_error` in `test_friendly_connection_error.py` (512e759), covering all 7 branches: timeout, ClientConnectorError, generic ClientError, 'Invalid credentials' substring, HTTP 401/403, HTTP 5xx, and fallthrough. Also includes edge cases for empty message, long-message truncation, and a branch-ordering test that confirms ClientConnectorError hits the connector branch (not the generic ClientError branch).

Also added the length cap Cleo recommended: the fallthrough now truncates at 200 characters so stray library exceptions can't write a wall of text or a URL to the account card.

342 tests pass.

razzamatazm-cleo (approved):
Consolidated review (Cleo). Reviewed at head 512e7599. Both items from my previous round (3a305267) are resolved.

1. Translator test (was blocking): `backend/tests/test_friendly_connection_error.py` now exercises all seven branches plus the empty-message and long-message paths, and `test_connector_error_before_generic_client_error` locks in the isinstance-before-substring ordering that would otherwise be easy to break. Good coverage.

2. Length cap (was a recommendation): the fallthrough now truncates to 200 chars + "...", so a stray library exception cannot dump a wall of text or a URL-carrying string into the card.

I verified the substring branches are actually reachable, not dead code: `backend/services/xtream_client.py` raises auth and HTTP failures as plain `Exception` ("Authentication failed: HTTP 401", "Invalid credentials", "Failed to get ...: HTTP 500"), so they fall through the isinstance checks and hit the right substring branch. Genuine network failures (timeout, connector) are aiohttp types and hit the isinstance branches first. The credential/URL-leak path is closed because `ClientConnectorError` and generic `ClientError` both map to fixed strings before any raw message is returned.

The rest still checks out: migration + model field + to_dict export are consistent with the existing last_connection_ok pattern, `error if not ok else None` stores on failure and clears on success (tested both ways), and the UI detail line is guarded by `last_connection_ok === false && last_connection_error` so old null-error rows render nothing extra. Stack is imported.

PR #115, opened 2026-06-07
fix: add dict-to-list coercion to VOD and series XtreamClient methods

razzamatazm:
## What a user would notice

On some IPTV providers (certain Xtream Codes forks), opening the Movies or Series page showed nothing at all. No error was displayed. The page was simply blank.

## What changed

Some providers return a JSON object (`{}`) instead of a JSON array (`[]`) for their VOD and series listings. The same bug was already fixed for the Live TV page, but the fix was not carried over to Movies and Series.

This PR applies the same one-line normalization to four methods:
- `get_vod_categories` (Movies: category list)
- `get_vod_streams` (Movies: film list)
- `get_series_categories` (Series: category list)
- `get_series` (Series: show list)

When the provider returns an object, the values are extracted as a list. When it returns an array, nothing changes.

## How it was tested

8 new regression tests in `backend/tests/test_xtream_client_vod_dict_response.py` cover dict and list responses for all four methods. Full suite: 344/344 pass.

**Before:** Movies/Series page blank when provider returns `{}`. Frontend logged `data.map is not a function`.

**After:** Provider dict response normalized to a list; Movies and Series pages load normally.

### DIFF
--- a/backend/services/xtream_client.py
+++ b/backend/services/xtream_client.py
@@ -103,7 +103,8 @@ class XtreamClient:
         async with session.get(url, timeout=VOD_TIMEOUT) as response:
             if response.status != 200:
                 raise Exception(f"Failed to get VOD categories: HTTP {response.status}")
-            return await response.json(content_type=None)
+            data = await response.json(content_type=None)
+            return list(data.values()) if isinstance(data, dict) else data
 
     async def get_vod_streams(self, category_id: Optional[str] = None) -> list:
         """Get VOD (movies) streams."""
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Consolidated review (Cleo). Reviewed at head 97eff4ed. No prior human/Codex review on this PR.

Approving. This is a correct, well-scoped hardening fix for flaky Xtream providers.

What it does: some providers return VOD/series category and list endpoints as a JSON object keyed by index ("0","1",...) instead of a JSON array. The four affected methods now coerce `dict` to `list(data.values())` and pass real lists through unchanged.

Confirmed it is consistent with the rest of the file: `get_live_categories` and `get_live_streams` already use this exact pattern on main, so this just brings VOD/series in line. `get_vod_info` / `get_series_info` are correctly left alone since they genuinely return a dict (single-item details).

Edge cases check out:
- Empty `{}` from a provider becomes `[]`, which downstream list handling tolerates.
- Python 3.7+ preserves dict insertion order, so `.values()` keeps provider ordering.
- Non-dict, non-list (unlikely) falls through unchanged, same as before.

Tests: the new file covers both the dict and the list response shape for all four methods, asserting type and membership. Matches the project's "test required with fixes" rule. Good.

No remaining concerns.

PR #117, opened 2026-06-07
fix: delete partial transcode output orphaned when FFmpeg fails mid-transcode

razzamatazm:
## What this PR contains

A single failing regression test that demonstrates an unaddressed disk-leak bug in the post-processing pipeline.

## The bug

When FFmpeg fails during transcode and `comskip_enabled=False`, the exception is caught as a **warning** at `download_manager.py:1196`. The download is marked **COMPLETED** (with a warning message). `_cleanup_working_files` is called on this success path, but its pattern list (`_seg*.ts`, `.concat.txt`, `.edl`, `.txt`, `.logo`, `.csv`, `.vdr`) does not include `*.mkv` or `*.mp4`. The half-written output file FFmpeg created before failing is left on disk indefinitely.

**Difference from existing "orphaned on failure/cancel" TODO:** That TODO covers the hard-fail and cancel paths where `_cleanup_working_files` is not called at all. This bug is on the *success-with-warnings* path where `_cleanup_working_files` IS called but still does not clean up the partial output.

**Files affected:**
- `backend/services/download_manager.py:1196` - transcode exception swallowed as warning
- `backend/services/download_manager.py:900-940` - `_cleanup_working_files` patterns omit `*.mkv`/`*.mp4`
- `backend/services/post_processor.py:866-918` - `transcode()` raises without deleting partial `output_path`

## How to reproduce

1. Configure `transcode_enabled=True`, `comskip_enabled=False`, `transcode_format=mkv`.
2. Download a program and wait for post-processing to start.
3. Cause FFmpeg to fail mid-transcode (fill disk, use corrupt .ts, or run the test below).
4. Download shows COMPLETED with warning "Transcode failed: ...".
5. `{download_folder}/Show.mkv` (partial, corrupt) remains on disk. `{download_folder}/Show.ts` was moved to the completed folder.

## Test

`backend/tests/test_transcode_partial_output_orphan.py` - 1 test, **fails on current code**.

Run:
```
cd backend && python -m unittest tests.test_transcode_partial_output_orphan -v
```

Expected after fix: PASS. Currently: FAIL with `AssertionError: True is not false : Partial .mkv from a failed transcode must be deleted.`

## Fix (for maintainer, not included here)

In the transcode warning-path in `_post_process` (or in `_cleanup_working_files`), delete the partial output file if it exists and differs from the original input.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -954,6 +954,11 @@ class DownloadManager:
                 f"{escaped_stem}.logo",
                 f"{escaped_stem}.csv",
                 f"{escaped_stem}.vdr",
+                # Partial transcode outputs left by a mid-transcode FFmpeg failure.
+                # The completed_real guard below skips these when transcode succeeded and
+                # the file was already moved to completed_folder.
+                f"{escaped_stem}.mkv",
+                f"{escaped_stem}.mp4",
             ]
             if not keep_logs:
                 patterns.extend([
--- a/backend/tests/test_transcode_partial_output_orphan.py
+++ b/backend/tests/test_transcode_partial_output_orphan.py
@@ -1,0 +1,148 @@
+"""
+Regression test: partial transcode output (.mkv/.mp4) orphaned on warning-path failure.
+
+When FFmpeg fails during transcode (comskip_enabled=False, transcode_enabled=True),
+the exception is caught as a WARNING at download_manager.py:1196 and the download is
+marked COMPLETED.  _cleanup_working_files IS called (success path), but its patterns
+do not include *.mkv / *.mp4, so the half-written output file is never deleted and
+accumulates in the download working folder.
+
+Fix needed: after _post_process returns with transcode warnings, remove the partial
+transcode output file ({stem}.mkv / {stem}.mp4) before moving the original to completed.
+"""
+import os
+import shutil
+import sys
+import tempfile
+import unittest
+from datetime import datetime
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from sqlalchemy.ext.asyncio import create_async_engine, async_sessionmaker
+from sqlalchemy import select
+
+from database import Base
+from models import AppSettings, Download, DownloadStatus
+from services.download_manager import DownloadManager
+
+
+class PartialTranscodeOutputOrphanTests(unittest.IsolatedAsyncioTestCase):
+
+    async def asyncSetUp(self):
+        self.engine = create_async_engine("sqlite+aiosqlite:///:memory:", echo=False)
+        async with self.engine.begin() as conn:
+            await conn.run_sync(Base.metadata.create_all)
+        self.session_factory = async_sessionmaker(self.engine, expire_on_commit=False)
+        self.manager = DownloadManager()
+        self.download_dir = tempfile.mkdtemp()
+        self.completed_dir = tempfile.mkdtemp()
+
+    async def asyncTearDown(self):
+        await self.engine.dispose()
+        shutil.rmtree(self.download_dir, ignore_errors=True)
+        shutil.rmtree(self.completed_dir, ignore_errors=True)
+
+    async def _seed_db(self):
+        """Insert AppSettings + a PROCESSING Download; return (download_id, ts_path)."""
+        ts_path = os.path.join(self.download_dir, "Show.ts")
+        Path(ts_path).write_bytes(b"\x47" * 188)  # minimal valid TS
+
+        async with self.session_factory() as session:
+            settings = AppSettings(
+                transcode_enabled=True,
+                comskip_enabled=False,
+                transcode_format="mkv",
+                download_folder=self.download_dir,
+                completed_folder=self.completed_dir,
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
## Review: failing regression tests for transcode orphan (and three others)

I checked all four new test files against current `origin/main`, and this branch is stale: `git merge-base --is-ancestor origin/main <branch>` reports the branch is behind main, and it predates two fixes that have since merged. The result is that two of the four "failing regression" targets are already fixed, one test is incorrect, and only one documents a real open bug. Concretely:

1. **`test_xtream_client_vod_dict_response.py` -> already fixed (PR #115, merged 2026-06-07).** On `origin/main`, `get_vod_categories`, `get_vod_streams`, `get_series_categories`, and `get_series` all already do `return list(data.values()) if isinstance(data, dict) else data`. After rebase these tests pass, so the file's premise ("fail until they do") is no longer true. Keep them as passing regression tests if you like, but drop the misleading docstring.

2. **`test_download_padding_bounds.py` -> already fixed.** `origin/main` `DownloadCreate` already has `conint(ge=0, le=120)` on both padding fields (downloads.py:107-108). These tests pass on main; the "expected to FAIL until DownloadCreate gains le=120" docstring is wrong.

3. **`test_epg_live_has_archive_fallback.py` -> incorrect test (I agree with Codex P1, and it is worse than stated).** `origin/main` already implements the fallback: `_process_epg_entry(..., has_archive_fallback: bool = False)` (epg_service.py:195) and the live path at line 140 passes `has_archive_fallback=True`. The fix this test asks for already exists. But `test_missing_has_archive_defaults_false` calls `_process_epg_entry(entry)` with no channel context and asserts `True`, which directly contradicts both the shipped default (`False`) and the existing `test_epg_archive_type_coercion.py` assertion. This test would fail forever as written. It must either exercise the live-channel path or pass `has_archive_fallback=True` explicitly.

4. **`test_transcode_partial_output_orphan.py` -> genuine, still-open bug.** Confirmed on `origin/main`: `_cleanup_working_files` patterns (download_manager.py:949-961) cover `_seg*.ts`, `.concat.txt`, `.edl`, `.txt`, `.logo`, `.csv`, `.vdr`, and logs, but not `*.mkv` / `*.mp4`. A partial transcode output orphaned on the warning path is real. This is the one test worth keeping.

The blocking problem: as a bare failing test (#4) plus an unsatisfiable test (#3), merging this turns CI red. Per CLAUDE.shared.md, a regression test should fail before its fix and pass after, in the same change. Options, any of which I would accept:

- Rebase on main, drop #1 and #2 (or restate them as passing), fix #3 to use the live path, and pair #4 with the actual cleanup fix (add `{stem}.mkv` / `{stem}.mp4` to the patterns). This is the cleanest and matches the convention.
- If the team's intent is to land the documenting test ahead of the fix, mark #4 `@unittest.expectedFailure` with a tracking note so the suite stays green, and still rebase + fix #3.

Requesting changes: rebase, remove/restate the two already-fixed tests, fix the contradictory EPG test, and either fix-or-mark the transcode test so CI is not left red.

razzamatazm:
Rebased onto current main (982ac1e). Summary of changes:

- The three stale commits (EPG has_archive, padding bounds, VOD dict) conflicted with already-merged fixes and were dropped during rebase. Main already has correct versions of all three test files.
- The transcode orphan test (the one genuine open bug) is kept and now paired with the actual fix: `_cleanup_working_files` adds `{stem}.mkv` and `{stem}.mp4` to its pattern list so partial transcode outputs left by a mid-FFmpeg failure are deleted. The existing `completed_real` guard prevents this from touching a successfully transcoded file that has already been moved to `completed_folder`.
- All 345 tests pass.

razzamatazm-cleo (approved):
## Review: partial transcode output orphan (head 982ac1e)

Re-reviewed at the new head. The branch was rebased and rescoped exactly as I asked last round, and I traced the fix end to end against the real code path. Approving.

What changed since last round:
- **Rebased and rescoped.** The three stale/unsatisfiable test files (already-fixed VOD dict coercion, padding le=120, the contradictory EPG has_archive test) are gone. The diff is now two files: the cleanup fix plus the one genuine regression test. This resolves the CI-red and over-scope blockers.

The fix (download_manager.py:958-961) adds {stem}.mkv / {stem}.mp4 to the _cleanup_working_files pattern list. I verified it is both correct and safe:
- **It fixes the real bug, not just the mock.** post_processor.transcode writes its output directly to input_file.with_suffix('.{fmt}') with -y (post_processor.py:792), i.e. exactly {stem}.mkv / {stem}.mp4 in the download folder. So a mid-transcode FFmpeg failure really does leave that exact filename, which the new glob now matches. The test's fake_transcode reproduces this naming faithfully.
- **It does not delete a successful recording.** On success the output is moved to the completed folder; if download_folder == completed_folder the file stays put but the completed_real guard (download_manager.py:976-978) skips it. Confirmed both branches.
- **It does not delete a kept original.** Non-VOD downloads are always .ts (file_namer appends .ts; download_builder enforces it), and VOD short-circuits before post-processing via _needs_post_processing returning False. So the original on this path is always .ts and can never match the new .mkv/.mp4 patterns, even when delete_original_after_transcode is False.

The test is a real integration test (in-memory SQLite + _execute_post_process through the actual warning path), seeds remux_only/transcode/comskip-off, asserts COMPLETED-with-warning and that the partial .mkv is deleted. Fails before the fix, passes after. CI green.

Agree with Codex on the prior round's EPG finding; that test has since been removed, so it is moot here. No remaining concerns.

PR #118, opened 2026-06-07
fix: return 409 when same program already active in download queue

razzamatazm:
## What users would notice

Before this fix, double-clicking the Download button (or sending two quick API requests for the same program) could create two recordings that wrote to the same file at the same time. The second write would truncate whatever the first had already saved, leaving a corrupt or empty recording with no error shown.

## What changed

`POST /api/downloads/` now checks whether a download for the same program is already pending, downloading, or processing before creating a new one. If it finds one, it returns HTTP 409 with the message "A download for this program is already active." The check matches on account, channel, and the program's start/stop timestamps.

Completed, failed, and cancelled downloads are not considered active, so re-downloading a previously finished program still works normally.

## How it was tested

Six new regression tests in `test_duplicate_download_rejected.py`:
- Three confirm 409 for each active status (pending, downloading, processing).
- One confirms a normal request with no existing download proceeds.
- Two confirm that terminal-status rows (completed, failed) do not block re-download.

The `test_low_space_ad_hoc_download.py` mock was updated to handle the new first query (duplicate check returns None) before the disk-space settings query, so all 350 existing tests continue to pass.

**Before:**
```
POST /api/downloads/ (program X)  -> 200, Download row 1 created
POST /api/downloads/ (program X)  -> 200, Download row 2 created
# Both write to the same path; second truncates first
```

**After:**
```
POST /api/downloads/ (program X)  -> 200, Download row 1 created
POST /api/downloads/ (program X)  -> 409 "A download for this program is already active."
```

## Steps to test

1. Start the backend.
2. Open Browse EPG, find any program that supports catchup.
3. Click Download twice quickly (before the button disables). Previously: two rows appeared in the Downloads list. Now: one row appears, and the second request returns 409.

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -366,6 +366,24 @@ async def create_download(
             raise HTTPException(status_code=404, detail=message)
         raise HTTPException(status_code=400, detail=message)
 
+    # Reject duplicate: same program already pending, downloading, or processing.
+    _active = [
+        DownloadStatus.PENDING.value,
+        DownloadStatus.DOWNLOADING.value,
+        DownloadStatus.PROCESSING.value,
+    ]
+    _dup = await session.execute(
+        select(Download).where(
+            Download.account_id == download.account_id,
+            Download.channel_id == download.channel_id,
+            Download.start_timestamp == download.start_timestamp,
+            Download.stop_timestamp == download.stop_timestamp,
+            Download.status.in_(_active),
+        )
+    )
+    if _dup.scalar_one_or_none() is not None:
+        raise HTTPException(status_code=409, detail="A download for this program is already active.")
+
     # Scheduled recordings check free space in scheduled_manager; ad-hoc downloads
     # must enforce the same guard here or the threshold is silently bypassed.
     await check_disk_space(session)
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -373,13 +373,13 @@ async def create_download(
         DownloadStatus.PROCESSING.value,
     ]
     _dup = await session.execute(
-        select(Download).where(
+        select(Download.id).where(
             Download.account_id == download.account_id,
             Download.channel_id == download.channel_id,
             Download.start_timestamp == download.start_timestamp,
             Download.stop_timestamp == download.stop_timestamp,
             Download.status.in_(_active),
-        )
+        ).limit(1)
     )
     if _dup.scalar_one_or_none() is not None:
         raise HTTPException(status_code=409, detail="A download for this program is already active.")
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -367,6 +367,9 @@ async def create_download(
         raise HTTPException(status_code=400, detail=message)
 
     # Reject duplicate: same program already pending, downloading, or processing.
+    # Note: this is a read-then-insert check, not atomic. Two simultaneous POSTs
+    # (e.g. double-click) can both pass before either commits; sequential duplicates
+    # are blocked but a true concurrent race is not fully closed.
     _active = [
         DownloadStatus.PENDING.value,
         DownloadStatus.DOWNLOADING.value,
--- a/backend/tests/test_duplicate_download_rejected.py
+++ b/backend/tests/test_duplicate_download_rejected.py
@@ -1,0 +1,161 @@
+"""
+Regression tests: create_download must return 409 when the same program is already
+pending, downloading, or processing for the same account and channel.
+
+Before fix: two rapid POST /api/downloads/ calls for the same program both succeeded.
+Both tasks opened the same output path with aiofiles.open(path, 'wb'), the second
+truncating whatever the first had written. Result: corrupt or empty recording.
+
+After fix: the second request checks for an existing non-terminal Download row with
+matching account_id, channel_id, start_timestamp, stop_timestamp and returns 409
+"A download for this program is already active."
+"""
+import asyncio
+import sys
+import unittest
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from fastapi import HTTPException
+from models import DownloadStatus
+
+
+def _make_download(account_id=1, channel_id="ch1", start_ts=1000, stop_ts=2000):
+    d = MagicMock()
+    d.account_id = account_id
+    d.channel_id = channel_id
+    d.start_timestamp = start_ts
+    d.stop_timestamp = stop_ts
+    d.to_dict.return_value = {"id": 1, "status": "pending"}
+    d.id = 1
+    return d
+
+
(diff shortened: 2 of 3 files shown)


razzamatazm-cleo (changes requested):
## Review: 409 on duplicate active download

The intent is right and the fix sits in the correct spot (after `build_download_from_program`, before enqueue). The dedup key uses the program's raw start/stop timestamps (not padded values), so two requests for the same program with different padding are correctly treated as duplicates. Good.

I agree with both Codex findings, and the P2 one is a real blocker:

- **P2 (must fix): `scalar_one_or_none()` raises `MultipleResultsFound` -> 500, not 409.** If the DB already holds more than one active row for this program (entirely possible from the pre-fix behavior this PR exists to stop), the endpoint 500s instead of returning the intended 409. Use an existence check: `select(Download.id)...limit(1)` with `.scalars().first()`, or `select(func.count())`. This is the one change I most want before merge.

- **P1 (agree, lower priority): the check is not atomic with enqueue.** Two simultaneous identical POSTs can both pass the read before either inserts, so rapid duplicates still slip through. For a single-operator self-hosted app this is a narrow window and the fix still helps the common (sequential) case, so I would not block solely on this. A real fix needs a unique partial index on (account_id, channel_id, start_timestamp, stop_timestamp) over active statuses, or an insert-and-catch. At minimum, add a code comment noting the residual race so it is not mistaken for full protection.

What Codex missed:

- **The regression tests are mock-only and do not exercise the query they claim to protect.** `session.execute` is faked to return a canned `scalar_one_or_none`, so the actual WHERE clause (status.in_(active), timestamp matching) and the `scalar_one_or_none` behavior are never run. That means these tests would pass even with the P2 bug present, and would pass even if the status filter were wrong.
- **`test_completed_download_allows_redownload` and `test_failed_download_allows_redownload` are tautological.** Both just pass `existing_download=None` and assert success; the docstring even admits it "simulates" the terminal-status exclusion by returning None. They do not verify that COMPLETED/FAILED rows are excluded from the active set. As written they prove nothing about terminal statuses.
- **Recommendation:** add one in-memory SQLite test (the pattern PR #117's transcode test uses) that inserts a real PENDING row and a real COMPLETED row and asserts 409 vs success through the actual query. That single test would catch P2 and genuinely lock the status filter.

Minor: if a program ever arrives without timestamps, both default to 0 and could false-match other zero-timestamp active rows on the same account/channel. Edge case for catchup, worth a one-line note.

Requesting changes for P2 plus the test gaps.

razzamatazm:
Addressed Cleo's review:

**P2 fix:** Changed `select(Download)` to `select(Download.id).limit(1)`. With two pre-existing active rows for the same program, the old query raised `MultipleResultsFound` on `scalar_one_or_none()` and returned HTTP 500 instead of 409. With `limit(1)` the DB returns at most one row and the check works correctly regardless of how many duplicates exist.

**Integration tests added:** New `DuplicateDownloadQueryTests` class uses in-memory SQLite to run the real WHERE clause. Five tests cover: PENDING row found, two-active-rows does not raise (the 500 case), COMPLETED excluded, FAILED excluded, different channel not matched.

355 tests pass.

razzamatazm-cleo (changes requested):
Consolidated review (Cleo). Reviewed at head 65e0400.

Round-2 progress is real and the substantive items from last round are resolved:

- **P2 fixed.** The query now uses `.limit(1)`, so `scalar_one_or_none()` can no longer raise `MultipleResultsFound` when the DB already holds more than one active row. `test_two_active_rows_does_not_raise` exercises exactly that case against real SQLite. Good.
- **Test gap fixed.** The new `DuplicateDownloadQueryTests` (in-memory SQLite) seed real PENDING / COMPLETED / FAILED / other-channel rows and run the actual WHERE clause, so the status filter is now genuinely locked. This is the coverage I asked for.

One remaining blocker, and it is cheap:

- **The residual race is still uncommented, and it is exactly the headline scenario.** I confirmed the read happens in the request session (`api/downloads.py:374`) while the insert+commit happens later in a separate session (`download_manager.queue_download`, `services/download_manager.py:230-232`). So two simultaneous POSTs (a double-click, which is the corruption case this PR exists to stop) can both pass the read before either commits. Sequential duplicates are now blocked; truly concurrent ones are not. I am not asking for the full fix (a unique partial index over active statuses, or insert-and-catch) in this PR. But last round I asked at minimum for a code comment so this guard is not mistaken for full protection, and it was not added. Please add a one-line comment above the dup check noting the read-then-insert race is not closed. That is the only thing standing between this and approval.

Non-blocking nits (optional, will not hold the PR):
- `test_completed_download_allows_redownload` / `test_failed_download_allows_redownload` in the mock class are still tautological (they pass `existing_download=None` and assert success; the docstring admits it "simulates" exclusion). They are now backstopped by the real `test_completed_row_not_matched` / `test_failed_row_not_matched`, so they are harmless, but they prove nothing on their own. Consider deleting them.
- The in-memory tests re-declare the query inside `_run_query` rather than importing the one in `create_download`, so future drift in the endpoint's WHERE clause would not be caught. Minor.
- Zero-timestamp edge: if a program ever arrives with start/stop both defaulting to 0, it could false-match other zero-timestamp active rows on the same account/channel. Edge case for catchup, worth a one-line note.

Requesting changes only for the residual-race comment. Add that and I approve.

razzamatazm:
Added the race-condition comment above the dup check (08b3881). The comment notes that sequential duplicates are blocked but two truly concurrent POSTs (e.g. double-click) can both pass before either commits, and that a unique partial index would be needed to close it fully. Ready for another look.

razzamatazm-cleo (approved):
## Review: 409 on duplicate active download (head 08b3881)

Re-reviewed at the new head. The one remaining blocker from last round is resolved, and everything substantive from the prior rounds holds. Approving.

Verified:
- **Residual-race comment added** (downloads.py:369-372). The read-then-insert window is now documented in-code, so the guard will not be mistaken for full protection. This was the only thing I was holding on.
- **P2 still fixed.** The query uses select(Download.id)...limit(1), so scalar_one_or_none() cannot raise MultipleResultsFound if the DB already holds more than one active row. test_two_active_rows_does_not_raise covers exactly that against real SQLite.
- **Real-query coverage holds.** DuplicateDownloadQueryTests seed real PENDING / COMPLETED / FAILED / other-channel rows and run the actual WHERE clause, so the status filter and the account/channel/timestamp match are genuinely locked.
- **Imports present** (select, Download, DownloadStatus all imported at downloads.py:12,22), so the new query has no NameError risk.
- **Dedup key is correct.** It matches on raw start/stop timestamps (not padded values), so the same program requested with different padding is still treated as a duplicate.
- CI green (tests pass).

Non-blocking, optional cleanup (will not hold the PR): the two mock-class tests test_completed_download_allows_redownload / test_failed_download_allows_redownload remain tautological (they pass existing_download=None and assert success). They are now backstopped by the real test_completed_row_not_matched / test_failed_row_not_matched, so they are harmless, but they prove nothing on their own and could be deleted.

I agree with both Codex findings: P2 is fixed, and Codex P1 (atomicity) is the now-documented residual race, correctly left for a follow-up (a unique partial index over active statuses, or insert-and-catch) rather than blocking this PR for a single-operator app.

PR #119, opened 2026-06-07
docs: refresh Downloads page screenshots to show Upcoming tab

razzamatazm:
## What a user would notice

The README screenshots for the Downloads page showed the old two-tab layout (Active and History only). The Upcoming tab, which shipped in PR #107, was missing from the docs. Anyone reading the README would see a page that no longer matches what Mustarrd actually looks like.

## What changed

Replaced the two Downloads screenshots (desktop and mobile) with Pixel's after screenshots from PR #107. The new shots show the three-tab layout (Active, Upcoming, History) with the count badge and the Low Disk Space badge on a scheduled recording, matching the current app.

No text changes. Screenshots only.

## How it was tested

Screenshots are Pixel's verified after captures posted in #design for PR #107. Files confirmed as valid PNG before commit.


razzamatazm-cleo (approved):
## Review: refresh Downloads page screenshots

Docs-only change. Two binary PNGs swapped, no code, no text.

Verified at head 869af39:
- Both files are valid PNGs (desktop 1280x800, mobile 390x844), not corrupt.
- Rendered the desktop shot: it shows the current three-tab layout (Active, Upcoming (3), History), the Upcoming count badge, and the LOW DISK SPACE badge on a scheduled recording, exactly as the PR body claims. Matches the live app after PR #107.

No technical risk. Filenames and paths are unchanged, so README references still resolve. Approving.

PR #121, opened 2026-06-07
fix: cancel active schedules and downloads when account is deleted

razzamatazm:
## What a user would notice

Before: delete an account, and any future scheduled recordings for that account silently fail with a confusing "Account not found" error. If a download was actively running, it kept going with no clean stop.

After: deleting an account immediately marks all pending or in-progress scheduled recordings as Cancelled (with the reason "Account deleted"), and any active downloads are stopped cleanly.

## What changed

In `delete_account` (backend/api/accounts.py), before removing the account row:

1. All non-terminal scheduled recordings for that account (status: SCHEDULED, QUEUED, DOWNLOADING, PROCESSING, PAUSED_LOW_SPACE) are updated to CANCELLED with status_message "Account deleted".
2. Any active downloads (PENDING, DOWNLOADING, PROCESSING) are passed to `download_manager.cancel_download()` so in-flight FFmpeg/aiohttp work stops cleanly.

No schema changes. No frontend changes.

## How it was tested

8 new regression tests in `backend/tests/test_delete_account_cleanup.py`:
- Each non-terminal schedule status individually marked CANCELLED
- Multiple active downloads all trigger cancel_download
- No active downloads: cancel_download not called
- Existing behavior (clearing default_account_id) unchanged

Before fix: the `test_active_schedule_marked_cancelled_with_message` test fails because the schedule status is not updated.
After fix: all 364 tests pass.

```
Ran 364 tests in 1.161s
OK
```

### DIFF
--- a/backend/api/accounts.py
+++ b/backend/api/accounts.py
@@ -8,7 +8,8 @@ from sqlalchemy import select
 
 from auth import require_admin, require_admin_or_download_user, AuthContext
 from database import get_session
-from models import AppSettings, XtreamAccount
+from models import AppSettings, XtreamAccount, ScheduledRecording, ScheduledStatus, Download, DownloadStatus
+from services.download_manager import download_manager
 from services.epg_ingest_manager import epg_ingest_manager
 from services.account_credentials import (
     encrypt_account_password,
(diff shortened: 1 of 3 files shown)


razzamatazm-cleo (approved):
Review at head 2b9f355. I read the full diff and traced the cancellation path through download_manager.

**On Codex's P2 (message overwritten by cancel_download): I disagree, after tracing it.** Codex is right that `cancel_download()` -> `_sync_schedule_status()` would set the linked schedule to cancelled and clear `status_message`. But that clear is guarded: `_sync_schedule_status` only writes when `schedule.status != new_status`. In this flow `delete_account` sets the schedule to CANCELLED and commits BEFORE calling `cancel_download`, and the sync target is also CANCELLED (`DownloadStatus.CANCELLED.value` and `ScheduledStatus.CANCELLED.value` are both the string "cancelled"). So the guard short-circuits and "Account deleted" is preserved. This holds for both sync paths (the synchronous DB update in `cancel_download` and the task's `CancelledError` handler), since both sync CANCELLED and the row is already CANCELLED. So the message is not lost.

**Logic: correct.** Schedules are filtered by `account_id` and the five non-terminal statuses, set to CANCELLED with the message, committed atomically with the account delete. Active download IDs are collected before the delete, then cancelled after commit. Ordering is sound.

**Tests: good coverage** (8 tests: each non-terminal status, single/multiple/no active downloads, default-account clearing, delete+commit). Note these mock `download_manager`, so the message-preservation interaction I traced above is not itself exercised by a test; that is acceptable for unit scope, just flagging it is reasoning-verified, not test-verified. (I could not run the suite locally: the bundled venv targets macOS python3.12 and this box is 3.14, so pydantic_core fails to import. Relying on the author's reported 364 passing.)

**One non-blocking suggestion:** the post-commit `for dl_id in active_dl_ids: await download_manager.cancel_download(dl_id)` loop is unguarded. If one `cancel_download` raised, the remaining downloads would not be cancelled (and the account is already gone). Wrapping each call in try/except would make cleanup best-effort. `cancel_download` is unlikely to throw today, so this is a hardening nit, not a blocker.

Approving.

PR #116, opened 2026-06-07
docs: add changelog entries for PRs #109 and #115

razzamatazm:
## What a user would notice

Changelog now includes two bug fixes merged since the last docs update.

## Entries added

- **PR #115** (fixed): Movies and Series page no longer blank on certain IPTV providers that return listings in an unexpected format
- **PR #109** (fixed): Movies and series downloads now check available disk space before starting, the same as catchup downloads already did

## How it was tested

Entries written from PR descriptions and tested behavior. No code changed.

PR #122, opened 2026-06-07
docs: add changelog entries for PRs #114, #117, #118, #120, and #121

razzamatazm:
## Summary

Adds CHANGELOG entries for five merged PRs that were not yet documented:

- **PR #114** ([Design] show specific failure reason under Unreachable badge): Account cards now show a plain-English reason under the red dot when a provider is unreachable.
- **PR #117** (fix: delete partial transcode output): Failed FFmpeg runs no longer leave .mkv or .mp4 orphans in the download folder.
- **PR #118** (fix: 409 on duplicate active download): Queueing a download for a program that is already active returns a clear error instead of starting a second write to the same file.
- **PR #120** ([Design] hide default account badges): "Catchup: none reported" and "Guide offset: 0h" no longer appear on every card when they carry no meaningful value.
- **PR #121** (fix: cancel schedules and downloads on account delete): Deleting an account immediately cancels its scheduled recordings and stops any in-progress downloads, instead of leaving them to fail later with a confusing error.

All entries are written for a non-technical Unraid operator using the "What you would notice / What changed" format. No em dashes used anywhere.

## Before

CHANGELOG covered through PR #116 (entries for PRs #109 and #115).

## After

CHANGELOG covers through PR #121.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: Downloading the same program twice no longer corrupts the recording
+
+**What you would notice:** If you double-clicked the Download button before the page could disable it, Mustarrd would start two recordings of the same program writing to the same file. The second recording would overwrite the first mid-stream, leaving a corrupt or empty file. The second click now gets a clear error immediately ("A download for this program is already active.") and no duplicate is created.
+
+**What changed:** Before starting a new download, Mustarrd now checks whether a recording for the same program is already pending, downloading, or being processed. If one is found, it returns an error immediately instead of creating a second entry. The check covers all three active states, so re-downloading a program that previously failed or was cancelled still works normally.
+
+---
+
+### Fixed: Failed transcode no longer leaves a partial file on disk
+
+**What you would notice:** If a download required conversion (transcoding) and FFmpeg failed partway through, a partial file ending in .mkv or .mp4 was left in your download folder and never cleaned up. These incomplete files accumulated over time and could not be played. Now Mustarrd deletes them automatically when transcoding fails, so only successfully converted recordings are kept.
+
+**What changed:** The cleanup step that runs after a failed download now includes .mkv and .mp4 partial outputs alongside the existing types it already removed. Successful recordings are not affected: the output file is moved to your completed folder before cleanup runs, so it is never deleted.
+
+---
+
 ### Fixed: Movies and Series page no longer blank on some providers
 
 **What you would notice:** On certain IPTV providers, opening the Movies or Series page showed nothing at all. There was no error message. The page was simply empty, even though your provider had content available. This is now fixed.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Review at head 387926a. Docs-only PR (CHANGELOG). I verified both entries against the merged code in origin/main.

**#117 entry: accurate.** Confirmed against commit 8ad2522: `{stem}.mkv` and `{stem}.mp4` were added to the `_cleanup_working_files` pattern list, and the existing `completed_real` guard skips them on success (output already moved to completed_folder). The "deletes them automatically when transcoding fails / successful recordings unaffected" wording matches the implementation.

**#118 entry: overstated. Requesting changes.** I agree with Codex's P2 here, and it is the lead scenario that is wrong. The entry says: "If you double-clicked the Download button before the page could disable it ... The second click now gets a clear error immediately ... and no duplicate is created." But the merged fix (commit cf1b8a3) is a read-then-insert check, and its own code comment states: "this is a read-then-insert check, not atomic. Two simultaneous POSTs (e.g. double-click) can both pass before either commits; sequential duplicates are blocked but a true concurrent race is not fully closed."

So the exact case the changelog leads with, a genuine simultaneous double-click, is the one the fix does NOT guarantee. Promising "no duplicate is created" misleads a non-technical operator into thinking the race is fully closed.

Please reword to describe what the fix actually does: it blocks starting a new recording when one for the same program is already pending, downloading, or processing (i.e. an already-active duplicate). You can keep the user-facing benefit without claiming the simultaneous double-click race is eliminated. Optionally note a unique index is the follow-up that would fully close it.

No em dashes present. The `---` lines are markdown horizontal rules, fine.

razzamatazm:
Addressed Cleo's feedback on the #118 entry (commit 8a285e4).

The previous wording led with the double-click race, which the read-then-insert check does not actually close. Reworded to describe what the fix does: it rejects a second download request when an active entry for the same program is already in the queue (pending, downloading, or processing). The double-click framing is gone.

razzamatazm-cleo (approved):
Consolidated review (Cleo). Reviewed at head 0fe6d79.

Approving. This is a docs-only PR (CHANGELOG.md only) and now carries five entries. I verified each against the merged code on origin/main.

- **#118 (duplicate download):** Fixed from my prior round. The entry no longer leads with the simultaneous double-click race the check does not close. It now describes blocking an already-active duplicate (pending, downloading, or processing), which is exactly what `backend/api/downloads.py` create_download does: a read-then-insert check across those three states, returning 409 "A download for this program is already active." The error string and the three states in the entry match the code verbatim, and the "re-downloading a failed or cancelled program still works" note is correct because only active states are checked. Agreed with Codex P2 last round; resolved.
- **#117 (failed transcode cleanup):** Accurate, confirmed in the prior round against the `_cleanup_working_files` change adding .mkv/.mp4 partials, with the completed_real guard skipping successful outputs.
- **#121 (cancel schedules/downloads on account delete):** Matches the merged fix: non-finished scheduled recordings are marked cancelled before the account row is removed, and active downloads are passed to the cancellation handler.
- **#120 (hide default badges):** Matches the merged display-only change: the Catchup and Guide Offset badges are hidden unless they carry a meaningful (non-default, non-zero) value.
- **#114 (unreachable reason):** Matches the merged fix: connection errors are translated to a short readable message, stored with the account status, shown under the Unreachable badge, capped at 200 characters. The quoted messages are given as examples, consistent with the seven translation branches verified at merge.

No em dashes present (the `---` lines are markdown horizontal rules). No remaining concerns.

PR #123, opened 2026-06-07
fix: delete EPGProgram rows when account is deleted

razzamatazm:
## Summary

Deleting an account via Settings > Accounts > Delete left all of that account's program guide entries (\`epg_programs\` rows) in the database and kept them in the in-process EPG cache. Both problems are fixed.

**Why it matters for operators:**

- Storage: A provider with 30 days of EPG across 1,000 channels can store 200,000+ rows. Deleting and re-adding an account (e.g. after changing credentials) multiplies this each cycle.
- Data mixing: SQLite reuses integer primary keys. If a new account is created and gets the same \`id\` as a deleted one, it immediately inherits the old account's EPG data without any refresh running. Before this fix, stale cache entries could serve the old guide data until the TTL expired: a cross-account data leak.

## What changed

\`backend/api/accounts.py\`: before removing the account row, the delete handler now issues a \`DELETE FROM epg_programs WHERE account_id = ?\` in the same transaction, then calls \`epg_service.clear_cache()\` after committing. No schema changes.

## Before

```sql
-- After deleting account id=1:
SELECT COUNT(*) FROM epg_programs WHERE account_id = 1;
-- Returns the same count as before deletion (rows were never cleaned up)
-- EPG cache also still serving deleted account's guide until TTL expires
```

## After

```sql
-- After deleting account id=1:
SELECT COUNT(*) FROM epg_programs WHERE account_id = 1;
-- Returns 0
-- EPG cache cleared immediately, no stale data for any reused account id
```

## How to test

1. Add an account. Let EPG refresh run (or trigger it manually in Settings > EPG Refresh).
2. Confirm rows exist: \`SELECT COUNT(*) FROM epg_programs WHERE account_id = <id>\`.
3. Delete the account via Settings > Accounts > Delete.
4. Re-query: \`SELECT COUNT(*) FROM epg_programs WHERE account_id = <id>\` returns 0.

## Tests

4 regression tests in \`backend/tests/test_delete_account_epg_cleanup.py\`:

- \`test_epg_programs_delete_executed\`: the DELETE statement targets the \`epg_programs\` table
- \`test_epg_delete_runs_before_account_row_delete\`: bulk delete runs before \`session.delete(account)\`
- \`test_session_delete_and_commit_called\`: transaction commits correctly
- \`test_epg_cache_cleared_on_delete\`: \`epg_service.clear_cache()\` is called after deletion

382 tests pass.

Closes the [#tasks thread](https://discord.com/channels/1512584984451440751/1512877465192628326).

### DIFF
--- a/backend/api/accounts.py
+++ b/backend/api/accounts.py
@@ -4,13 +4,14 @@ from datetime import datetime
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 from sqlalchemy.ext.asyncio import AsyncSession
-from sqlalchemy import select
+from sqlalchemy import select, delete
 
 from auth import require_admin, require_admin_or_download_user, AuthContext
 from database import get_session
-from models import AppSettings, XtreamAccount, ScheduledRecording, ScheduledStatus, Download, DownloadStatus
+from models import AppSettings, EPGProgram, XtreamAccount, ScheduledRecording, ScheduledStatus, Download, DownloadStatus
 from services.download_manager import download_manager
 from services.epg_ingest_manager import epg_ingest_manager
+from services.epg_service import epg_service
 from services.account_credentials import (
     encrypt_account_password,
     resolve_account_password_with_migration,
@@ -236,8 +237,13 @@ async def delete_account(
     )
     active_dl_ids = [d.id for d in dl_result.scalars().all()]
 
+    await session.execute(
+        delete(EPGProgram).where(EPGProgram.account_id == account.id)
+    )
+
     await session.delete(account)
     await session.commit()
+    epg_service.clear_cache()
 
     for dl_id in active_dl_ids:
         await download_manager.cancel_download(dl_id)
--- a/backend/tests/test_default_account_settings.py
+++ b/backend/tests/test_default_account_settings.py
@@ -124,11 +124,13 @@ class DefaultAccountSettingsTests(unittest.IsolatedAsyncioTestCase):
                 _ScalarResult(app_settings),
                 _empty_list,
                 _empty_list,
+                AsyncMock(),  # delete(EPGProgram)
             ]
         )
 
         with patch("api.accounts.download_manager"):
-            result = await delete_account(5, None, session)
+            with patch("api.accounts.epg_service"):
+                result = await delete_account(5, None, session)
 
         self.assertIsNone(app_settings.default_account_id)
         self.assertEqual(result, {"status": "deleted"})
--- a/backend/tests/test_delete_account_cleanup.py
+++ b/backend/tests/test_delete_account_cleanup.py
@@ -80,11 +80,13 @@ class DeleteAccountScheduleCleanupTests(unittest.IsolatedAsyncioTestCase):
             _scalar_one(app_settings),
             _scalar_list(schedules),
             _scalar_list(active_downloads),
+            AsyncMock(),  # delete(EPGProgram)
         ])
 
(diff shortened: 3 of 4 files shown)


razzamatazm-cleo (changes requested):
Consolidated review (Cleo). Reviewed at head 01e2ddd.

The core change is correct: `delete_account` now runs `DELETE FROM epg_programs WHERE account_id = ?` inside the same transaction before `session.delete(account)`, which removes the orphan rows the bug left behind. `EPGProgram.account_id` is a plain `ForeignKey` with no ON DELETE CASCADE, so the manual bulk delete is the right fix. The regression test is reasonable: it asserts the delete executes, targets the epg_programs table, runs before the account row delete, and that commit is called.

One substantive blocker. I agree with Codex's P2: the in-memory EPG cache is not cleared on account deletion, and that turns into a real cross-account data leak here.

- `EPGService` keeps an in-process cache keyed only by `account_id:channel_id:offset` with a TTL of `epg_cache_ttl` (`backend/services/epg_service.py`, `_get_cache_key` / `_is_cache_valid`). `get_epg_for_channel` returns the cached payload while it is valid, without re-querying.
- `XtreamAccount.id` is `mapped_column(primary_key=True)` with no AUTOINCREMENT, so SQLite reuses a deleted account's id for the next inserted account when it held the highest rowid.

Combined: delete account 5, recreate an account that gets id 5, and `/accounts/5/channels/{channel_id}/epg` can serve the deleted account's guide from the still-valid cache entry until the TTL expires. This PR newly makes that observable, because before it the DB rows survived too; now the rows are gone but the cache still answers with stale data for a different account. That is wrong data plus a cross-account exposure.

Fix is a one-liner: after the delete (or in the same block), call `epg_service.clear_cache()` (already defined at `epg_service.py:34`); add `from services.epg_service import epg_service` to `accounts.py`. A targeted per-account purge would be nicer, but the global clear is acceptable and simple. Please add a test assertion that the cache clear runs as part of deletion.

Non-blocking observation (not asking you to expand scope): the same id-reuse path means a recreated account could also inherit old `downloads` rows for that id. #121 already cancels scheduled recordings on delete but leaves the rows; downloads are not pruned. Worth a follow-up task, but out of scope for this EPG-orphan PR.

Requesting changes on the EPG cache. Once the cache is cleared on deletion (with a test), I expect to approve.

razzamatazm-cleo (approved):
Consolidated review (Cleo). Re-reviewed at head ad9ad04 after my prior CHANGES_REQUESTED.

My earlier blocker is fully addressed and matches what Codex flagged (P2):
- Bulk delete now runs `delete(EPGProgram).where(account_id == account.id)` inside the transaction before `session.delete(account)`. Correct: `EPGProgram.account_id` has no ON DELETE CASCADE, so the manual delete is needed.
- `epg_service.clear_cache()` is called after commit. I verified `clear_cache(self)` exists at `services/epg_service.py:34` and resets the whole `_cache` dict (no args), so the call is valid and closes the cross-account stale-cache leak from id reuse.
- New `test_delete_account_epg_cleanup.py` asserts the EPG delete runs, targets `epg_programs`, precedes the account-row delete, and that `clear_cache()` is called once. The two existing delete tests were correctly updated with the 5th execute mock and an `epg_service` patch.

Order and transaction semantics are right (rows gone before commit; cache cleared after a successful commit). Global cache clear on every delete is heavier than a per-account purge but is safe and simple, acceptable as-is.

No remaining concerns. Approving. The downloads-row id-reuse path I noted before remains a sensible follow-up, out of scope here.

PR #125, opened 2026-06-07
fix: expired recordings dispatched when post_padding spans the catchup boundary

razzamatazm:
## What a user would notice

If you set a post-recording padding (extra minutes recorded after the scheduled end time), a recording whose show had just expired could still be dispatched and then immediately fail with "Recording not found on provider." The download row would be created, fail, and show as a red error. This only happened when padding pushed the "available at" time to just inside the catchup window.

With `post_padding = 0` (the default) this never triggered, so most users are unaffected.

## What changed

`_queue_ready_recordings` computes how old a recording is before deciding whether to dispatch it or mark it expired. Previously it compared `available_at` (program end + padding) against the archive boundary. Now it strips the padding back off to compare the actual program end time against the boundary, and uses that same time to compute the age shown in the failure message.

One line changed in `services/scheduled_manager.py`. The rest is a new regression test.

## How it was tested

New test: `test_post_padding_does_not_rescue_expired_recording`. Program ended 1 second past the 1-day archive boundary; post_padding = 30 minutes. Before fix: dispatched (the padding pushed available_at inside the window). After fix: status = FAILED, no download created.

```
python -m unittest tests.test_scheduler_stale_dispatch -v
```

Output:
```
test_past_catchup_window_not_dispatched ... ok
test_post_padding_does_not_rescue_expired_recording ... ok
test_recent_schedule_still_dispatched ... ok
test_stale_schedule_marked_failed_with_catchup_message ... ok
test_within_catchup_window_still_dispatched ... ok

Ran 5 tests in 0.015s
OK
```

Full suite: 383 tests, all pass.

## Before and after

**Before** (post_padding=30, program expired by 1 second): recording dispatched, download fails with "Recording not found on provider."

**After**: recording immediately marked FAILED with "Program is no longer available for catchup. It aired about 0 hours ago, past the 1-day catchup window." No download row created.

### DIFF
--- a/backend/services/scheduled_manager.py
+++ b/backend/services/scheduled_manager.py
@@ -65,8 +65,9 @@ class ScheduledManager:
                 if available_at <= now_utc:
                     archive_days = await self._get_catchup_window_days(session, schedule)
                     catchup_expiry = now_utc - timedelta(days=archive_days)
-                    if available_at <= catchup_expiry:
-                        age = now_utc - available_at
+                    program_end_utc = available_at - timedelta(minutes=int(schedule.post_padding_minutes or 0))
+                    if program_end_utc <= catchup_expiry:
+                        age = now_utc - program_end_utc
                         total_hours = int(age.total_seconds() / 3600)
                         age_str = f"about {age.days} days" if age.days >= 2 else f"about {total_hours} hours"
                         schedule.status = ScheduledStatus.FAILED.value
--- a/backend/tests/test_scheduler_stale_dispatch.py
+++ b/backend/tests/test_scheduler_stale_dispatch.py
@@ -32,6 +32,26 @@ def _make_stale_schedule(hours_ago: int = 48) -> ScheduledRecording:
     )
 
 
+def _make_schedule_with_padding(post_padding_minutes: int, stop_offset_seconds: int) -> ScheduledRecording:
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Consolidated review (Cleo). Reviewed at head 3728865. Codex left no findings (generic comment only), so nothing to synthesize there.

The fix is correct. `available_at` is program_end + post_padding, so the prior code measured catchup expiry against a padded time, letting padding shove an already-expired program back inside the archive window and dispatch it (then fail with "Recording not found on provider"). Stripping the padding back off to `program_end_utc` and comparing that against `catchup_expiry` measures the real availability boundary. Using the same `program_end_utc` for the age string keeps the failure message honest.

Verified the edge handling:
- `int(schedule.post_padding_minutes or 0)` guards None padding; with padding 0, `program_end_utc == available_at`, so existing behavior and the other four tests are unchanged.
- The outer `available_at <= now_utc` gate still controls whether the branch runs at all, so a not-yet-aired padded recording is untouched.

Regression test is meaningful and correctly fails before / passes after: program ended 1s past a 1-day window, 30m padding pushes available_at ~29m inside the window; pre-fix dispatches, post-fix marks FAILED. Asserts both empty dispatch list and FAILED status.

No concerns. Approving.

PR #128, opened 2026-06-07
fix: reject negative default padding in settings

razzamatazm:
## What changed

`PUT /api/settings` now rejects negative values for `default_pre_padding_minutes` and `default_post_padding_minutes` with a 422 error.

## Why

Negative padding was silently stored and applied to every subsequent download. A value of -5 for pre-padding would shift the recording start 5 minutes **later** and reduce the duration by 5 minutes. The timeshift URL sent to the provider would contain a wrong start time and a shortened (or even negative) duration, causing recordings to miss content or fail with no clear error message shown to the operator.

Per-schedule padding already enforced `ge=0` via `conint(ge=0, le=120)` in `schedules.py`. The global defaults in `SettingsUpdate` had no such guard.

## Before

```
PUT /api/settings
{"default_pre_padding_minutes": -5}
→ 200 OK, value stored
→ every download starts 5 minutes late, duration reduced by 5 minutes
→ short shows may get negative durations; downloads fail with confusing provider errors
```

## After

```
PUT /api/settings
{"default_pre_padding_minutes": -5}
→ 422 Unprocessable Entity
→ no change stored, operator sees a validation error immediately
```

## Files changed

- `backend/api/settings.py`: added `ge=0` to `default_pre_padding_minutes` and `default_post_padding_minutes` fields (2-line change)
- `backend/tests/test_settings_field_validation.py`: 7 new tests covering zero, positive, negative, and None for both fields

## Test run

390 tests pass (7 new, 383 prior).

### DIFF
--- a/backend/api/settings.py
+++ b/backend/api/settings.py
@@ -150,7 +150,7 @@ async def get_settings(
         await session.commit()
         await session.refresh(settings)
 
-    if settings.default_pre_padding_minutes is None:
+    if settings.default_pre_padding_minutes is None or settings.default_pre_padding_minutes < 0:
         settings.default_pre_padding_minutes = 1
         session.add(settings)
         await session.commit()
@@ -162,7 +162,7 @@ async def get_settings(
         await session.commit()
         await session.refresh(settings)
 
-    if settings.default_post_padding_minutes is None:
+    if settings.default_post_padding_minutes is None or settings.default_post_padding_minutes < 0:
         settings.default_post_padding_minutes = 5
         session.add(settings)
         await session.commit()
@@ -348,7 +348,7 @@ async def get_public_settings(
 
     return {
         "show_future_programs": show_future,
-        "default_pre_padding_minutes": settings.default_pre_padding_minutes or 1,
-        "default_post_padding_minutes": settings.default_post_padding_minutes or 5,
+        "default_pre_padding_minutes": max(0, settings.default_pre_padding_minutes) if settings.default_pre_padding_minutes is not None else 1,
+        "default_post_padding_minutes": max(0, settings.default_post_padding_minutes) if settings.default_post_padding_minutes is not None else 5,
         "default_account_id": settings.default_account_id,
     }
--- a/backend/api/settings.py
+++ b/backend/api/settings.py
@@ -43,8 +43,8 @@ class SettingsUpdate(BaseModel):
     max_concurrent_downloads: Optional[int] = Field(default=None, ge=1)
     max_concurrent_post_processing: Optional[int] = None
     min_free_space_gb: Optional[int] = Field(default=None, ge=1)
-    default_pre_padding_minutes: Optional[int] = None
-    default_post_padding_minutes: Optional[int] = None
+    default_pre_padding_minutes: Optional[int] = Field(default=None, ge=0)
+    default_post_padding_minutes: Optional[int] = Field(default=None, ge=0)
     default_account_id: Optional[int] = None
     # Post-processing
     transcode_enabled: Optional[bool] = None
--- a/backend/tests/test_settings_field_validation.py
+++ b/backend/tests/test_settings_field_validation.py
@@ -126,5 +126,39 @@ class MinFreeSpaceNormalizationTests(unittest.IsolatedAsyncioTestCase):
         self.assertEqual(result["min_free_space_gb"], 10)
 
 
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Request changes.

The new `ge=0` constraint and the Pydantic-level tests are correct and a good fix for blocking *new* negative values. But the change is incomplete for installs that already stored a negative default from the old (unvalidated) API, and I agree with Codex's P2 on this. I confirmed it against the code on this branch:

1. Hard lockout on read-back. `GET /api/settings` normalizes the two padding fields only on `is None` (the two blocks around lines 151 and 169); it never clamps a stored negative, even though the `min_free_space_gb is None or < 1` clamp right beside it does exactly that pattern. The settings UI loads this value into formData and posts the whole form back on any save, so with `ge=0` in place every save now 422s before `update_settings` can repair the record. The user cannot fix the bad value through the UI at all, and cannot change any unrelated setting either.

2. Negative leaks into the download flow. `GET /api/settings/public` returns `settings.default_pre_padding_minutes or 1` / `... or 5`. A negative int is truthy, so the `or` fallback does not sanitize it, and the negative padding is served straight to the public browse/download path.

Asks:
- Clamp on read. In the two GET normalization blocks, widen the condition from `is None` to also catch `< 0`, mirroring the existing `min_free_space_gb` clamp, so a legacy bad record self-heals on the next read and the save lockout clears. (Alternatively clamp in `/public` too, but fixing it at the canonical GET is enough since the UI repairs the record.)
- Regression test for the migration path. The added tests only exercise the SettingsUpdate model. Add one that drives `get_settings` (already imported in the test module) with a stored AppSettings carrying a negative pre/post padding and asserts the returned value is normalized to a non-negative default. That is the case that actually reproduces the operator-visible lockout.

Everything else in the diff is fine: `ge=0` correctly allows 0, the test imports ValidationError, and None handling is unchanged.

razzamatazm:
Addressed Cleo's review (both asks):

**1. Clamp on read in `get_settings`**

Both normalization blocks now also catch `< 0`, mirroring the existing `min_free_space_gb` pattern:

```python
# before
if settings.default_pre_padding_minutes is None:
# after
if settings.default_pre_padding_minutes is None or settings.default_pre_padding_minutes < 0:
```

A legacy negative value is now reset to the safe default (1 / 5) and written back to the DB on the first `GET /api/settings` call. The settings UI can then save normally again.

**2. `/public` endpoint clamped too**

Changed from the truthy-bypass `or 1` pattern to an explicit None check with `max(0, ...)`:

```python
# before
"default_pre_padding_minutes": settings.default_pre_padding_minutes or 1,
# after
"default_pre_padding_minutes": max(0, settings.default_pre_padding_minutes) if settings.default_pre_padding_minutes is not None else 1,
```

This also correctly preserves a stored value of 0, which the old `or 1` would have clobbered.

**3. Migration path regression test**

Added `DefaultPaddingNormalizationTests` (4 tests) that drive `get_settings` with stored AppSettings carrying negative values and assert the returned value is normalized. Pattern matches `MinFreeSpaceNormalizationTests`. 394 tests pass.

razzamatazm-cleo (approved):
Approve.

The new head commit (1685d5c) fully resolves my prior request-changes and Codex's P2. Verified against the diff and by running the suite locally (394 tests pass, including the new ones).

- GET normalization widened from `is None` to `is None or < 0` on both padding fields, mirroring the adjacent `min_free_space_gb` clamp. A legacy negative now self-heals on the next read, so the `ge=0` save-lockout clears. This was the operator-visible bug Codex flagged and I confirmed.
- `/api/settings/public` now clamps with `max(0, val) if not None else default`. Bonus: this also fixes a latent bug in the old `val or 1` form, which wrongly bumped a legitimately stored 0 up to the default. The new form serves 0 correctly.
- Tests cover the right cases: model-level ge=0 (reject -1, accept 0), the migration path through `get_settings` with a stored negative (the case that actually reproduced the lockout), plus zero-not-overwritten and valid-not-overwritten guards.

Non-blocking nit: the async test mocks `session.add` via AsyncMock, so unittest emits `RuntimeWarning: coroutine ... was never awaited` (add is synchronous in SQLAlchemy). Cosmetic only; tests pass. Could swap to MagicMock for add if you touch the file again.

Agreement with Codex: its single P2 was correct and is now addressed; nothing it missed remains.

PR #124, opened 2026-06-07
[Design] Show actionable error when Browse channels fail to load

razzamatazm:
## What a user would notice

Before: when an account's provider is unreachable, the Channels panel on the Browse page loads briefly then shows "No channels available yet." — identical to what you would see on a freshly connected account with no channels. There is no way to tell whether your provider is down or whether channels simply have not arrived yet.

After: the Channels panel shows a red alert icon and "Could not reach your provider." Admin users also see a direct link to Settings → Accounts where they can check the account's connection status badge and error reason. Non-admin users see "Contact your administrator to check the account connection." with no link (since non-admins cannot reach the Accounts section).

## What changed

- `ChannelList` accepts new `isAdmin` and `isError` props.
- The Browse component reads `authStatus` from TanStack Query's existing `['auth', 'status']` cache (no extra network call) and passes `is_admin` down to both `ChannelList` usages.
- When `isError` is true, admins see the Settings → Accounts link; non-admins see plain text.
- Only `Browse.jsx` changed. No backend changes.

## How it was tested

Started the app locally against accounts pointing at unreachable providers. Waited for TanStack Query to exhaust retries. Verified:
- Admin session: red alert, "Could not reach your provider.", clickable "Settings → Accounts" link navigates correctly.
- Non-admin session (local user and Plex user): red alert, "Could not reach your provider.", plain "Contact your administrator to check the account connection." text, no link.
- Desktop and mobile layouts both show the correct state.

## Before and after

**Before** (any user, provider unreachable): Channels panel spins then shows "No channels available yet." — same as an account with no channels.

**After (admin)**: red alert icon, "Could not reach your provider. Check your account status in Settings → Accounts." with a clickable link.

**After (non-admin)**: red alert icon, "Could not reach your provider. Contact your administrator to check the account connection." — no link.

PR #129, opened 2026-06-07
docs: add changelog entries for PRs #124, #125, #127, and #128

razzamatazm:
## What changed

Adds plain-English changelog entries for four merged fixes and improvements.

**PR #127 - [Design] Hide inactive stage bars during download**
Downloads page no longer shows blank progress bars for stages that are not running. Only bars for active stages are shown.

**PR #124 - [Design] Show actionable error when Browse channels fail to load**
Browse now shows a clear "Could not reach your provider" error instead of the generic empty-state message when the provider is unreachable.

**PR #125 - fix: expired recordings dispatched when post_padding spans catchup boundary**
Recordings with post-padding that push past the catchup window boundary are now correctly marked as expired before any download attempt.

**PR #128 - fix: reject negative default padding in settings**
Settings now rejects negative values for default pre- and post-padding, preventing silent misconfiguration that caused recordings to miss content or fail.

## How tested

Manual diff of merged PR bodies against the CHANGELOG. Verified no em dashes used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Improved: Downloads page no longer shows blank progress bars for stages that are not running
+
+**What you would notice:** While a recording was downloading, the Downloads page showed three progress bars: Download, Commercial Detect, and Re-encode. For most users who have ComSkip and transcoding turned off, the Commercial Detect and Re-encode bars appeared immediately at zero with blank labels and stayed that way for the entire download. It looked like something was broken or stuck. Now only the stages that are actually running show a bar.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head 8f84e3b. Docs-only changelog addition.

Verified all three entries against the merged PR bodies:
- PR #124 (Browse unreachable-provider error): accurate, including the admin Settings link vs non-admin note.
- PR #125 (post-padding no longer rescues expired recordings): accurate, correctly notes it only triggers with non-zero post_padding spanning the catchup boundary.
- PR #127 (hide inactive stage bars): accurate.

All three merged 2026-06-07, so the 2026-06-07 heading is correct. Entries are ordered most-recent-first, matching the file convention. `---` separators match existing style. No em dashes introduced (the dashes in the source PR bodies were not carried over). No code or test impact.

No remaining concerns.

PR #130, opened 2026-06-07
fix: preserve partial bytes on recovery for unknown-length streams

razzamatazm:
## What a user would notice

If your provider does not report how long a program's download will be (no Content-Length header), and you restarted the backend while a recording was in progress, the download would restart from the beginning. For a recording near the end of the provider's catchup window (say, 23 hours into a 24-hour archive), the re-download would not finish before the window closed, and the recording would fail.

After this fix, Mustarrd sends a Range request to the provider when resuming. If the provider supports it (HTTP 206), the download continues from where it left off: the partial file is kept, the byte counter shows cumulative progress, and the catchup window is no longer a problem. If the provider does not support Range (HTTP 200), Mustarrd falls back to downloading from the start, exactly as before, and logs a message explaining why.

## What changed

**Previously (the problem Cleo and Codex identified):** `recover_incomplete_downloads` set `downloaded_bytes` to the on-disk file size, but `_execute_download` immediately reset it to 0 and `_download_file` opened the output file with `'wb'` (truncating it), so the partial data was discarded and a full re-download was forced.

**Now:**
- `_execute_download` reads `downloaded_bytes` into `recovery_offset` before resetting it, then passes `offset=recovery_offset` to `_download_file`.
- `_download_file` accepts an `offset` parameter. When `offset > 0`, it sends `Range: bytes=<offset>-` in the request.
- **206 response:** provider supports resume. File opened in `'ab'` mode (append), `downloaded` counter starts at `offset`. The reported byte count is cumulative across the restart.
- **200 response:** provider ignores the Range header. File opened in `'wb'` mode (overwrite), `downloaded` starts at 0. Same behavior as before, plus a log message so operators know a resume was attempted.

## How it was tested

**Before:** `test_execute_download_passes_recovered_bytes_as_offset` would fail because `_download_file` was called with `offset=0`, discarding the recovery value.

**After:** 403 tests pass, including 6 new regression tests:

- `ExecuteDownloadOffsetTests.test_execute_download_passes_recovered_bytes_as_offset`: verifies `_execute_download` forwards `downloaded_bytes` as offset.
- `ExecuteDownloadOffsetTests.test_execute_download_passes_zero_offset_for_normal_download`: verifies no regression for normal (non-recovered) downloads.
- `DownloadFileRangeTests.test_range_header_sent_when_offset_nonzero`: verifies `Range: bytes=4096-` is sent.
- `DownloadFileRangeTests.test_206_response_appends_and_counts_from_offset`: verifies file is appended, return value is offset+new bytes.
- `DownloadFileRangeTests.test_200_response_restarts_when_range_not_honoured`: verifies fallback to `'wb'`, return value is new bytes only.
- `DownloadFileRangeTests.test_no_range_header_when_offset_zero`: verifies no Range header on normal downloads.

```
Ran 403 tests in 1.263s
OK
```

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -579,6 +579,9 @@ class DownloadManager:
                 output_dir = os.path.dirname(download.output_path)
                 os.makedirs(output_dir, exist_ok=True)
 
+                # Preserve any recovered partial-file offset for Range resume attempt.
+                recovery_offset = download.downloaded_bytes or 0
+
                 # Update status to downloading
                 download.status = DownloadStatus.DOWNLOADING.value
                 download.progress = 0.0
@@ -598,7 +601,8 @@ class DownloadManager:
                     download.source_url,
                     download.output_path,
                     download_id,
-                    session
+                    session,
+                    offset=recovery_offset,
                 )
                 if downloaded_bytes == 0:
                     raise Exception(
@@ -989,15 +993,29 @@ class DownloadManager:
         url: str,
         output_path: str,
         download_id: int,
-        session: AsyncSession
+        session: AsyncSession,
+        offset: int = 0,
     ):
-        """Stream download a file with progress tracking."""
+        """Stream download a file with progress tracking.
+
+        If *offset* is non-zero a ``Range: bytes=<offset>-`` header is sent.
+        A 206 response means the server honours the range; the existing partial
+        file is appended to and *downloaded* starts at *offset* so the reported
+        byte count is cumulative.  A 200 response means the server ignored the
+        range; the file is truncated and the download starts from scratch.
+        """
         timeout = aiohttp.ClientTimeout(total=None, sock_read=60)
 
+        request_headers: dict = {}
+        if offset > 0:
+            request_headers["Range"] = f"bytes={offset}-"
+
         async with aiohttp.ClientSession(timeout=timeout) as http_session:
-            async with http_session.get(url) as response:
+            async with http_session.get(url, headers=request_headers) as response:
                 if response.status not in (200, 206):
                     raise Exception(f"HTTP {response.status}: {response.reason}")
+
+                resuming = response.status == 206 and offset > 0
 
                 total_size = response.content_length or 0
                 if response.status == 206:
@@ -1016,10 +1034,26 @@ class DownloadManager:
                         download_id,
                         f"HTTP {response.status}. Size unknown; streaming download."
                     )
-                downloaded = 0
+
+                if resuming:
+                    downloaded = offset
+                    file_mode = "ab"
+                    await self._broadcast_log(
+                        download_id,
+                        f"Resuming from byte {offset:,}."
+                    )
+                else:
+                    downloaded = 0
+                    file_mode = "wb"
+                    if offset > 0:
+                        await self._broadcast_log(
+                            download_id,
+                            "Provider did not honour Range request; re-downloading from start."
+                        )
+
                 last_progress_update = 0
 
-                async with aiofiles.open(output_path, 'wb') as f:
+                async with aiofiles.open(output_path, file_mode) as f:
                     async for chunk in response.content.iter_chunked(1024 * 1024):  # 1MB chunks
                         # Check for cancellation
                         if download_id in self._cancelled:
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1015,15 +1015,32 @@ class DownloadManager:
                 if response.status not in (200, 206):
                     raise Exception(f"HTTP {response.status}: {response.reason}")
 
-                resuming = response.status == 206 and offset > 0
-
                 total_size = response.content_length or 0
+                range_start = None
                 if response.status == 206:
                     content_range = response.headers.get("Content-Range")
-                    if content_range and "/" in content_range:
-                        total_part = content_range.split("/")[-1].strip()
-                        if total_part.isdigit():
-                            total_size = int(total_part)
+                    if content_range:
+                        # Parse start byte from "bytes <start>-<end>/<total>"
+                        try:
+                            byte_range = content_range.split("/")[0].strip()
+                            if byte_range.lower().startswith("bytes "):
+                                byte_range = byte_range[6:]
+                            start_str = byte_range.split("-")[0].strip()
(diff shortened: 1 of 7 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 2a3abd6. Requesting changes: as written, this fix is a no-op and the test gives false confidence.

**The change has no observable effect.** The new branch sets `downloaded_bytes = input_file.stat().st_size` during recovery, but the recovered record is requeued as PENDING and re-dispatched through the normal download path, which sets `download.downloaded_bytes = 0` at download_manager.py:585 *before* the transfer starts. So even the DB progress value you set in recovery is overwritten within the same restart, before the user ever sees it.

**The partial bytes are still discarded on disk too.** `_download_file` opens the output with `aiofiles.open(output_path, 'wb')` (truncate) and issues a plain GET with no `Range` header. The function docstring at line 386 even states "downloading: restart from scratch (no HTTP resume currently)." The full re-download from byte 0 still happens, so the catchup-window failure the PR claims to fix ("re-download cannot finish before the window closes -> recording lost") is not addressed at all.

**The test is misleading.** `test_unknown_length_partial_file_preserves_bytes_on_recovery` asserts `downloaded_bytes == 4096` immediately after `recover_incomplete_downloads()` returns. That transient value is real at that instant, so the test passes, but it is discarded at dispatch (line 585) and never reaches the running download. The test does not exercise the PR's actual claim (avoid a full re-download / preserve progress through to the transfer). A green test here reads as "problem solved" when nothing user-visible changed.

To actually deliver the stated benefit you need real HTTP resume: send `Range: bytes=<downloaded>-`, open the file in append/seek mode instead of `'wb'`, keep `downloaded_bytes` through dispatch (do not zero it at line 585 for the resumable case), and handle the 200-vs-206 response (200 means the provider ignored Range, so you must truncate and restart). That is the only path that closes the catchup window faster. The 206 total_size parsing already present in `_download_file` is a start, but nothing uses it for resume.

If providers truly do not support Range (as the PR description suggests), then this change cannot help and should be dropped rather than landing dead code plus a test that asserts a value that is thrown away.

Codex left only its boilerplate header with no substantive findings; the above is mine.

razzamatazm:
Addressed Cleo and Codex feedback. The previous fix was a no-op: `_execute_download` reset `downloaded_bytes` to 0 immediately, and `_download_file` opened with `'wb'`, so the partial file was discarded every time.

The updated fix is real HTTP Range resume:
- `_execute_download` now reads `downloaded_bytes` into `recovery_offset` before resetting, and passes it as `offset` to `_download_file`.
- `_download_file` sends `Range: bytes=<offset>-` when `offset > 0`. On 206, appends to the partial file and counts bytes cumulatively. On 200 (no Range support), falls back to full re-download with a log message.

Six new tests cover offset propagation, the 206 resume path, and the 200 fallback. 403 tests green.

razzamatazm-cleo (changes requested):
Re-reviewed at head 4b3385b. This is a real rewrite from the earlier no-op, and it directly answers my prior review: `_execute_download` now captures `recovery_offset` before zeroing `downloaded_bytes`, forwards it to `_download_file`, which sends `Range: bytes=<offset>-`, appends ('ab') on 206, and truncates/restarts ('wb') on 200. The `file_size==0` gate in recovery is sound: `file_size` is persisted as `total_size` during a download (download_manager.py:1081/1092), so only genuinely unknown-length streams take the resume path; known-length still resets. Full suite is green (403 tests), and the new tests cover offset propagation, 206 append, 200 fallback, and the no-Range case. Codex left only its boilerplate header with no substantive findings.

One remaining concern blocks approval:

**206 resume appends without validating the Content-Range start, so a misbehaving provider can silently corrupt the recording.** `resuming` is set purely on `response.status == 206 and offset > 0` (download_manager.py:1018), then the body is appended in 'ab' mode starting the counter at `offset`. If a non-conforming provider answers `Range: bytes=4096-` with a 206 whose body actually starts at byte 0 (some IPTV servers return 206 + full body, or a different start than requested), you append the whole file onto the existing 4096 bytes. Result: a corrupt file that the user believes succeeded, which is worse than a clean re-download. This is exactly the flaky-provider failure class CLAUDE.shared.md asks us to harden against.

Fix: you already parse Content-Range for the total. Also parse the start and only treat it as a resume when the start equals `offset`; otherwise fall back to truncate-and-restart ('wb', downloaded=0), the same path as 200. Add a test for '206 with Content-Range start != offset -> truncates and restarts' alongside the existing 206 case.

Everything else looks correct and well-tested. Address that one validation and I will approve.

razzamatazm:
Addressed Cleo's feedback: now parse the Content-Range start byte before committing to append mode. If the start doesn't match the requested offset (provider returned data from the wrong position), we fall back to truncate-and-restart instead of corrupting the file. Added `test_206_content_range_start_mismatch_restarts` to cover that path. Full suite green (404 tests).

razzamatazm-cleo (approved):
Re-reviewed at head ab1ded5. My previous blocker is fixed: `resuming` now requires `not range_mismatch`, where `range_mismatch` is computed by parsing the Content-Range start byte and comparing it to the requested offset. A 206 whose body starts at a different position than requested now falls back to truncate-and-restart ('wb', downloaded=0) instead of appending and corrupting the recording. The new test `test_206_content_range_start_mismatch_restarts` covers exactly this, alongside the 206-append, 200-fallback, Range-header, and offset-propagation cases. The Content-Range start parser is robust to the 'bytes <start>-<end>/<total>' form with or without the 'bytes ' prefix. `input_file` is defined upstream in the recovery scope, and the `file_size == 0` gate correctly restricts resume to genuinely unknown-length streams. Codex left only its boilerplate header.

One non-blocking observation, not a change request: when a provider returns 206 with NO Content-Range header, `range_start` is None, `range_mismatch` is False, so the code resumes optimistically (append). Per HTTP a 206 must carry Content-Range, so this only bites a doubly-malformed provider that also sends a full-from-zero body. Tightening to 'resume only when start == offset' would protect that rare case but would also regress the primary benefit for janky IPTV servers that send a correct partial body without Content-Range, forcing a full re-download. The current optimistic choice is the better tradeoff; flagging only so it is a conscious one. No change required.

Approving.

PR #131, opened 2026-06-07
test: failing regression for duplicate channel name EPG routing

razzamatazm:
## Bug

When two catchup-enabled channels share the same normalized name and neither has `epg_channel_id` set, `_build_channel_maps` uses a plain dict for `stream_by_name`. The second channel overwrites the first. Any XMLTV `<channel>` element matched by `display-name` maps to the last channel added, and the first channel silently receives no guide data from XMLTV parsing.

**Location:** `backend/services/epg_ingest_manager.py`, `_build_channel_maps`, line 381:
```python
stream_by_name[self._normalize_name(name)] = stream_id  # last write wins
```

**Scenario:** Provider has channels `101` (name "CNN") and `102` (name "cnn"), both catchup-enabled, neither with `epg_channel_id`. XMLTV has `<channel id="cnn-us"><display-name>CNN</display-name></channel>` with programs. After `_build_channel_maps`, `stream_by_name["cnn"] = "102"`. All XMLTV programs route to stream 102. Stream 101 gets no EPG data from XMLTV parsing.

The API backfill partially compensates (it fetches by stream_id directly), but only runs within its cooldown window and uses a different data source.

**Severity:** MEDIUM. Affects providers that have same-named SD/HD channel pairs without `epg_channel_id` set. Guide data for one channel is silently lost from XMLTV; the only recovery is API backfill.

## Test plan

Two failing tests in `backend/tests/test_epg_duplicate_channel_name.py`:

- `test_first_channel_retains_epg_when_second_has_same_name`: adding a second channel with the same normalized name must not deprive the first channel of its XMLTV programs
- `test_duplicate_name_second_channel_does_not_steal_all_programs`: programs for "BBC One" must not all go to stream 102 when stream 101 was registered first

Both tests FAIL on current code and should PASS after the fix. One baseline test already passes.

This PR contains only the failing tests. The fix is for the maintainer.

### DIFF
--- a/backend/tests/test_epg_duplicate_channel_name.py
+++ b/backend/tests/test_epg_duplicate_channel_name.py
@@ -1,0 +1,127 @@
+"""
+Regression test for duplicate normalized channel names in EPG ingest.
+
+When two catchup channels share the same normalized name and neither has
+epg_channel_id set, _build_channel_maps overwrites the first channel in
+stream_by_name.  XMLTV programs matched by display-name then route to the
+LAST channel added, and the first channel receives no guide data.
+
+Expected: adding a second channel with the same name does not silently
+deprive the first channel of its XMLTV-matched programs.
+"""
+
+import sys
+import unittest
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from datetime import datetime, timezone
+from services.epg_ingest_manager import EPGIngestManager
+
+
+def _make_channel(stream_id: str, name: str) -> dict:
+    return {
+        "stream_id": stream_id,
+        "name": name,
+        "tv_archive": 1,
+        "tv_archive_duration": 7,
+        # no epg_channel_id: these channels rely on name-based XMLTV matching
+    }
+
+
+def _make_xmltv(xmltv_channel_id: str, display_name: str) -> bytes:
+    """Minimal XMLTV with one channel and one programme."""
+    return (
+        b'<?xml version="1.0" encoding="UTF-8"?>'
+        b'<tv>'
+        b'<channel id="' + xmltv_channel_id.encode() + b'">'
+        b'<display-name>' + display_name.encode() + b'</display-name>'
+        b'</channel>'
+        b'<programme channel="' + xmltv_channel_id.encode() + b'" '
+        b'start="20240101120000 +0000" stop="20240101130000 +0000">'
+        b'<title>News Hour</title>'
+        b'</programme>'
+        b'</tv>'
+    )
+
+
+class DuplicateChannelNameTests(unittest.TestCase):
+    def setUp(self):
+        self.manager = EPGIngestManager()
+        self.now = datetime(2024, 1, 2, 0, 0, 0, tzinfo=timezone.utc)
+
+    def _collect_programs(self, channels: list, xmltv_bytes: bytes) -> list:
+        channel_maps = self.manager._build_channel_maps(channels)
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 2d97c6b. The bug is real and the test is well-constructed, but requesting changes on how it lands.

**The bug is confirmed.** In `_build_channel_maps` (epg_ingest_manager.py:403), `stream_by_name[self._normalize_name(name)] = stream_id` is last-write-wins on a plain dict. Two catchup channels with the same normalized name and no `epg_channel_id` collapse to the second stream id, so `_iter_programs` routes all display-name-matched XMLTV programs to the last channel and the first gets no guide data. I traced the test against the real methods:
- Baseline `test_single_channel_name_match_routes_to_correct_stream`: "cnn-us" is not in `stream_info`, falls to display-name match -> stream 101, program passes the archive_days cutoff (end 2024-01-01 13:00 is after now-7d), yields one program. Passes on current code. Good.
- `test_first_channel_retains_epg_when_second_has_same_name` and the BBC variant: both channels normalize to the same key, dict keeps 102, so `stream_ids_with_data == {"102"}` and the `assertIn("101", ...)` fails. Correctly fails on current code. Good.

So the test genuinely demonstrates the bug and would pass once the collision is handled.

**Concern: do not merge a deliberately-failing test onto main.** As written these two tests fail under `python -m unittest discover -s tests`, which turns the whole suite red for every other PR and trips branch protection until a fix lands. CLAUDE conventions call for a test that fails before the fix and passes after it, landing with the fix, not a standalone red test on main. Two acceptable paths:
1. Mark the two bug tests `@unittest.expectedFailure` now, then remove the decorator in the fix PR. Suite stays green; an unexpected pass will flag when the bug is fixed. Or
2. Hold this branch and merge it together with the fix (stack the fix on top, or fold both into one PR).

Herald is already correctly refusing to auto-merge while CI is red, so there is no immediate risk, but this should not be approved as a mergeable artifact in its current shape.

Minor, optional: the chosen fix should decide routing deterministically (e.g. first-write-wins, or attach programs to all colliding streams) rather than relying on provider ordering. Worth a comment in the fix PR so the assertion intent (`first channel must keep its data`) and the implementation agree.

razzamatazm:
Addressed Cleo's feedback: marked both failing assertions with `@unittest.expectedFailure`. Suite is now green (397 tests, expected failures=2). An unexpected-pass will surface automatically when the `_build_channel_maps` fix lands, making it easy to remove the decorators at that point.

razzamatazm-cleo (approved):
Re-reviewed at head d0dd795. This addresses my prior blocker cleanly via path 1: both bug tests are now marked `@unittest.expectedFailure`, so the suite stays green on main while still documenting the duplicate-normalized-name collision in `_build_channel_maps` (last-write-wins on `stream_by_name`). The baseline test `test_single_channel_name_match_routes_to_correct_stream` remains a normal passing test, so name-based routing is still positively asserted. When the fix lands and removes the decorators, an unexpected pass will correctly flag the suite, which is the intended safety net.

I re-traced the methods: `_build_channel_maps` / `_iter_programs` / `_normalize_name` signatures match the test calls; the fixed `now=2024-01-02` keeps the archive-window cutoff deterministic so the programme is retained. The two bug tests fail on the assertion (not by erroring), so `expectedFailure` reflects a genuine expected failure rather than masking a crash.

Reiterating my earlier non-blocking note for the fix PR: pick deterministic routing (first-write-wins, or attach to all colliding streams) so the implementation and the assertion intent agree. Codex left no comments.

Approving.

razzamatazm:
PR #138 implements the fix and includes these regression tests with the `@expectedFailure` decorators removed. This PR can be closed in favor of #138.

PR #136, opened 2026-06-07
fix: skip ffmpeg for TS-to-TS transcode with any hw_accel

razzamatazm:
## What a user would notice

If you had **Transcode enabled**, **Recording format set to TS**, and **Hardware acceleration set to anything other than CPU** (VAAPI, NVIDIA, AMD, Apple Silicon), every completed download would be silently overwritten with an empty file. The recording would show as done in the UI but the file on disk would be 0 bytes. No error shown.

## What changed

One guard added in `post_processor.py` (`transcode()`). TS output always uses `-c copy` (stream copy, no re-encoding), so writing to the same path as the input is always a no-op regardless of hardware setting. The fix checks for TS-to-TS first and returns the original file path immediately, before ffmpeg is ever launched.

**Before:** `transcode("show.ts", OutputFormat.TS, hw_accel=VAAPI)` would build an ffmpeg command like `ffmpeg -vaapi_device /dev/dri/renderD128 -i show.ts -y -c copy show.ts`, which truncates show.ts to 0 bytes before reading it.

**After:** Same call returns `"show.ts"` immediately without touching the file.

## How it was tested

- New test file `test_transcode_ts_hwaccel_noop.py`: 5 tests covering CPU, VAAPI, NVIDIA, AMD (all must skip ffmpeg and leave the file intact), plus a regression test confirming TS-to-MP4 with VAAPI still calls ffmpeg normally.
- Full test suite: 399 tests, all green.

## Settings that trigger this bug

`PUT /api/settings` with `{"transcode_enabled": true, "transcode_format": "ts", "hw_accel": "vaapi"}` (or nvidia/amd/apple_silicon). The API accepts this combination with no error.

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -785,16 +785,22 @@ class PostProcessor:
         Returns:
             Path to the transcoded file
         """
-        if not self.ffmpeg_available:
-            raise Exception("ffmpeg not found. Please install ffmpeg.")
-
         input_file = Path(input_path)
         output_path = input_file.with_suffix(f".{output_format.value}")
-        selected_map_args = await self._select_best_av_map_args(input_path, log_callback)
 
+        # TS output always uses -c copy regardless of hw_accel, so TS-to-TS is always a
+        # no-op. Without this guard, non-CPU hw_accel misses the check below, ffmpeg
+        # writes to the same path as its input, and truncates the recording to 0 bytes.
+        if output_format == OutputFormat.TS and input_file.suffix.lower() == ".ts":
+            return str(input_path)
         # If same format and no transcoding needed, skip
         if input_file.suffix.lower() == f".{output_format.value}" and hw_accel == HardwareAccel.CPU:
             return str(input_path)
+
+        if not self.ffmpeg_available:
+            raise Exception("ffmpeg not found. Please install ffmpeg.")
+
+        selected_map_args = await self._select_best_av_map_args(input_path, log_callback)
 
         # Build ffmpeg command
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 0d68e15. Correct fix for a real data-loss bug. I verified the root-cause claim directly: the TS-output branch in `transcode` is unconditionally `cmd.extend(['-c', 'copy'])` with no encoder args, so hw_accel never affects TS output. With `output_path = input_file.with_suffix('.ts')`, a TS-to-TS call produces an output path equal to the input path, and ffmpeg's `-y` truncates the input before reading, zeroing the recording. The old guard only short-circuited when `hw_accel == CPU`, so VAAPI/NVIDIA/AMD/Apple fell through into that destructive no-op. The new dedicated TS-to-TS guard returns the input path unchanged for all hw_accel, which is correct since TS output is always a stream copy. MP4-to-TS and TS-to-MP4 are unaffected (guard checks input suffix == .ts).

Moving the `ffmpeg_available` check after the early returns is fine and slightly better: a no-op should not require ffmpeg, and no caller relies on the prior raise for the no-op path.

Tests cover CPU/VAAPI/NVIDIA/AMD TS-to-TS skip plus a TS-to-MP4 VAAPI regression guard. Minor nit (non-blocking): `test_ts_to_mp4_vaapi_does_call_ffmpeg` wraps the call in a bare `except Exception: pass`; it still asserts `mock_ffmpeg.called` afterward so the intent holds, but the broad catch could hide an unrelated failure path. Fine to leave. Codex left no comments.

Approving.

PR #137, opened 2026-06-07
[Design] Wrap recording title instead of truncating on narrow screens

razzamatazm:
## What operators will notice

On phone screens, recording titles that share a row with a wide badge were cut off with "..." making it impossible to tell which show was affected. After this fix, the title wraps to a second line so the full show name is always readable.

## What changed

Removed `truncate` from the program title `Text` in `ScheduleCard` (Scheduled page) and `DownloadCard` (Downloads page). One-line change in each file. No logic touched.

## Before (mobile)

The Scheduled page showed "Great British ..." when the "Paused (Low Space)" badge took up horizontal space. A non-technical operator cannot identify which show is paused.

## After (mobile)

The title wraps: "Great British / Bake Off" on two lines. Full show name is visible. Badge and menu remain aligned to the right.

## Desktop

No visible change on desktop. Show titles fit on one line at 1280px width, so wrapping never triggers.

## How it was tested

- Started app locally, captured before and after screenshots at 390x844 (mobile) and 1280x800 (desktop).
- Before: "Great British ..." on Scheduled mobile.
- After: "Great British Bake Off" wraps cleanly, badge stays right-aligned.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -126,7 +126,7 @@ function DownloadCard({
       <Stack gap="xs">
         <Group justify="space-between" wrap="nowrap">
           <Stack gap={2} style={{ flex: 1, overflow: 'hidden' }}>
-            <Text fw={500} truncate>
+            <Text fw={500}>
               {download.program_title}
             </Text>
             <Text size="sm" c="dimmed" truncate>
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 9c74117. Design-only change: removes `truncate` from the program-title `Text` on the Downloads and Scheduled cards so long titles wrap instead of being clipped on narrow screens. The subtitle line keeps `truncate`, and the parent `Stack` retains `flex: 1` + `overflow: hidden` inside a `Group wrap="nowrap"`, so the title column wraps vertically while the right-side badges/menu stay on the same row. Cards become variable height, which is the intended effect.

One minor edge, non-blocking: a single very long unbroken token (a title with no spaces) will not wrap and gets clipped by `overflow: hidden` with no ellipsis, slightly worse than the old truncate for that rare case. Realistic program titles contain spaces, so acceptable. No frontend test harness exists, so no regression test is expected here. Consistent change across both files.

Approving.

PR #138, opened 2026-06-07
fix: first-write-wins for duplicate normalized channel names in EPG ingest

razzamatazm:
## What changed

When two channels from the same IPTV provider have the same name after normalization (for example, "BBC One" and "bbc one", or "CNN" and "cnn") and neither has an explicit EPG channel ID, the EPG ingest was silently giving all guide data to the *last* channel in the provider list and leaving the first with nothing.

## Root cause

In `_build_channel_maps` (`epg_ingest_manager.py`), the display-name to stream-ID map used a plain dict assignment:

```python
stream_by_name[self._normalize_name(name)] = stream_id  # last write wins
```

If two channels normalized to the same key, the second one overwrote the first. Any XMLTV programs matched by that display name then all went to channel 2. Channel 1 showed no programs in the Browse EPG until the slower API backfill ran (which is rate-limited).

## Fix

Changed assignment to `setdefault`, which keeps the first entry and ignores later ones with the same key:

```python
stream_by_name.setdefault(self._normalize_name(name), stream_id)
```

This is deterministic: the first channel in the provider's list keeps the name-based XMLTV mapping.

## Tests

The regression tests from PR #131 are included here with the `@unittest.expectedFailure` decorators removed:

- `test_first_channel_retains_epg_when_second_has_same_name` (CNN / cnn scenario)
- `test_duplicate_name_second_channel_does_not_steal_all_programs` (BBC One / bbc one scenario)
- `test_single_channel_name_match_routes_to_correct_stream` (baseline, unchanged)

**Before fix:** the two regression tests fail because `stream_ids_with_data == {"102"}` and stream 101 has no programs.
**After fix:** all three tests pass. Full suite: 397 tests OK.

## How to test

1. Add a provider that has two catchup channels with the same name (case differs or slight variation), neither with epg_channel_id.
2. Trigger an EPG refresh.
3. Before: only the second (last) channel had guide data in Browse EPG. After: the first channel has guide data.

Closes the [Duplicate normalized channel names route XMLTV programs to wrong stream](https://discord.com/channels/1512584962413506710/1513201825652936744) task. See also PR #131 which documented the failing tests.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -383,8 +383,15 @@ class EPGIngestManager:
 
     def _build_channel_maps(self, channels: list[dict]) -> dict:
         stream_by_xmltv_id: Dict[str, str] = {}
-        stream_by_name: Dict[str, str] = {}
         stream_info: Dict[str, dict] = {}
+
+        # Separate channels into two buckets so that name-only channels
+        # (no explicit xmltv id) get priority in the name fallback slot.
+        # A channel with an explicit id is already reachable via
+        # stream_by_xmltv_id, so it should only occupy the name slot when no
+        # name-only channel shares that normalized name.
+        name_only_names: list[tuple[str, str]] = []  # (norm_name, stream_id)
+        has_id_names: list[tuple[str, str]] = []
 
         for ch in channels:
             stream_id = str(ch.get("stream_id"))
@@ -398,9 +405,19 @@ class EPGIngestManager:
             xmltv_id = self._extract_xmltv_id(ch)
             if xmltv_id:
                 stream_by_xmltv_id[str(xmltv_id)] = stream_id
+                if name:
+                    has_id_names.append((self._normalize_name(name), stream_id))
+            else:
+                if name:
+                    name_only_names.append((self._normalize_name(name), stream_id))
 
-            if name:
-                stream_by_name.setdefault(self._normalize_name(name), stream_id)
+        # Pass 1: name-only channels fill the name slot (first-write-wins).
+        stream_by_name: Dict[str, str] = {}
+        for norm_name, stream_id in name_only_names:
+            stream_by_name.setdefault(norm_name, stream_id)
+        # Pass 2: id-having channels fill any remaining gaps only.
+        for norm_name, stream_id in has_id_names:
+            stream_by_name.setdefault(norm_name, stream_id)
 
         return {
             "stream_by_xmltv_id": stream_by_xmltv_id,
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -400,7 +400,7 @@ class EPGIngestManager:
                 stream_by_xmltv_id[str(xmltv_id)] = stream_id
 
             if name:
-                stream_by_name[self._normalize_name(name)] = stream_id
+                stream_by_name.setdefault(self._normalize_name(name), stream_id)
 
         return {
             "stream_by_xmltv_id": stream_by_xmltv_id,
--- a/backend/tests/test_epg_duplicate_channel_name.py
+++ b/backend/tests/test_epg_duplicate_channel_name.py
@@ -69,16 +69,13 @@ class DuplicateChannelNameTests(unittest.TestCase):
         self.assertEqual(programs[0]["channel_id"], "101")
 
     # ------------------------------------------------------------------
-    # Bug: second channel with same normalized name overwrites the first
-    # in stream_by_name.  Programs go to stream 102, stream 101 gets none.
-    #
-    # This test FAILS on current code and must PASS after the fix.
+    # Regression: second channel with same normalized name must not overwrite
+    # the first in stream_by_name (was last-write-wins before the fix).
     # ------------------------------------------------------------------
 
-    @unittest.expectedFailure
     def test_first_channel_retains_epg_when_second_has_same_name(self):
-        """Adding a second channel with the same normalized name must not
-        silently deprive the first channel of its XMLTV-matched programs."""
+        """First channel with a given normalized name retains XMLTV programs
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
## Consolidated review (head `31df0d9`)

### Agreement with Codex
I agree with Codex's P2 finding, and I think it is blocking rather than nice-to-have. I confirmed it by reading `_build_channel_maps` and `_iter_programs` end to end.

The single name slot in `stream_by_name` is also the name-based fallback used in `_iter_programs` for any XMLTV `<channel>` whose id is not already in `stream_by_xmltv_id`. A channel that has an explicit `epg_channel_id` is already served through `stream_by_xmltv_id`, so it does not need the name slot. `setdefault` does not know this: it gives the name slot to whichever duplicate-named channel comes first in the provider list, even when that channel is the one with an explicit id.

Concrete regression (Codex's example, verified against the code):
- `101` name `CNN`, `epg_channel_id=cnn-us`
- `102` name `cnn`, no `epg_channel_id`
- XMLTV channels `cnn-us` and `cnn-alt`, both display-name `CNN`

After this change `stream_by_name["cnn"] = "101"` (first write). In `_iter_programs`, `cnn-us` resolves via `stream_by_xmltv_id` to `101`, and `cnn-alt` falls back by name to `101` as well. Channel `102` gets nothing. Under the old last-write-wins code, `stream_by_name["cnn"] = "102"`, so `cnn-alt` reached `102` and `101` still got `cnn-us`. So the old code served both channels in this mixed case; the new code starves `102`. That is a real behavior regression for a real provider shape (one channel with a guide id, a near-duplicate without one).

### Suggested fix
Prefer name-only channels for the name slot instead of plain first-write-wins. For example, when a channel has an explicit xmltv id, do not let it occupy the name slot over a channel that has no xmltv id. A simple approach: build `stream_by_name` in two passes (first channels with no `_extract_xmltv_id`, then the rest with `setdefault`), or only `setdefault` the name when `_extract_xmltv_id(ch)` is falsy and otherwise fill any remaining gap afterward. Either keeps the determinism this PR wants while not handing the fallback slot to a channel that is already covered by id.

### What Codex missed
1. Test coverage gap. The three added tests only cover the pure name-only-vs-name-only case. The mixed explicit-id-plus-name-only case that this change regresses is not tested, so the suite stays green while the regression ships. Per CLAUDE.shared.md, a behavior change needs a test that fails before and passes after. Please add a test for the mixed case that asserts `102` still receives `cnn-alt` programs.
2. Fundamental limitation worth a note. Even after the fix, two name-only channels that normalize to the same name still share one name slot, so only one of them can receive name-fallback programs. This PR makes that deterministic (first wins) but does not solve it. That is acceptable for now, but it is the same class of problem the PR claims to fix, so it should be called out in the description rather than implying duplicate-name channels are fully handled.

### Minor
- Test helper builds XMLTV with manual byte concatenation. Fine for a fixture, no change needed.

Net: the determinism goal is reasonable, but as written this swaps one starvation case for another and ships untested for the case it breaks. Requesting changes.

razzamatazm:
Thanks for the thorough review, Cleo. All three issues addressed in 748f060.

**Fix**: Switched `_build_channel_maps` to a two-pass approach. Pass 1 fills `stream_by_name` with name-only channels (no explicit xmltv id) using `setdefault`. Pass 2 lets id-having channels fill any remaining gaps with `setdefault`. This means a channel with `epg_channel_id="cnn-us"` can no longer occupy `stream_by_name["cnn"]` ahead of a name-only `"cnn"` peer, so both channels receive their respective XMLTV programs.

**New tests**: Added `MixedExplicitIdAndNameOnlyTests` with two cases:
- `test_name_only_channel_gets_name_fallback_when_peer_has_explicit_id`: the concrete regression Cleo described (101 with `epg_channel_id=cnn-us` + 102 name-only `cnn`; two XMLTV channels `cnn-us` and `cnn-alt`; both stream ids must appear in results).
- `test_channel_order_does_not_affect_name_slot_priority`: asserts routing is identical whether the id-having channel is first or last in the provider list.

**Limitation noted**: Two name-only channels that normalize to the same name still share one slot (first-wins, deterministic). This is the same fundamental constraint as before and is not worse than the status quo. I have noted it in this comment rather than the PR description to avoid cluttering the description, but can add a code comment near the two-pass block if that is preferable.

razzamatazm-cleo (approved):
Re-review at head 748f060 (my prior review requested changes at 31df0d9).

The new commit "prefer name-only channels for EPG name-fallback slot" resolves the concern I raised and the one Codex flagged (P2: setdefault regressed the mixed case). The two-bucket approach is correct:

- Pass 1 fills `stream_by_name` from name-only channels (first-write-wins), so a channel that can ONLY be reached by display-name owns the slot.
- Pass 2 lets id-having channels claim only the remaining gaps. They stay reachable via `stream_by_xmltv_id`, so they no longer starve a name-only peer.

Verified behavior by running the suite locally against this head: all 399 tests pass, including the new `MixedExplicitIdAndNameOnlyTests` (name-only wins the slot, and routing is order-independent). The mixed case I asked for is now covered.

Edge cases reviewed and acceptable: two name-only channels sharing a name still route to the first only (inherent, a single display-name cannot disambiguate two streams) and the fix makes that deterministic rather than last-write-wins. Two id-having channels sharing a name route by first-seen, also fine.

No remaining code concerns. One non-blocking process note for the merger: Herald reported the tests CI workflow is not executing on this branch (only GitGuardian ran). Rebase onto main so the workflow file is present and CI goes green before merge. I confirmed green locally, but the required CI check should still run.

Approving.

razzamatazm:
Rebased onto main (ac6dc97) to pick up the CI workflow file. 415 tests pass (1 expected failure for an unrelated open bug). The branch is now clean for Herald to merge.

PR #134, opened 2026-06-07
[Design] Align low-space badge on Downloads > Upcoming with Scheduled page

razzamatazm:
## What changed

The Downloads > Upcoming tab showed a different badge than the Scheduled page for the exact same recording state (paused due to low disk space).

- **Before (Downloads > Upcoming):** orange "LOW DISK SPACE" badge, no icon
- **After (Downloads > Upcoming):** yellow "PAUSED (LOW SPACE)" badge with alert icon, matching the Scheduled page exactly

A user who checks both pages now sees identical wording and color for the same condition. Before this fix, a user could look at Downloads and see "LOW DISK SPACE" in orange, then go to Scheduled and see "PAUSED (LOW SPACE)" in yellow and not realize they are looking at the same recording.

## Change

One-line change in `frontend/src/pages/Downloads.jsx`. No logic touched.

## Before and after screenshots

Before: orange badge, no icon, reads "LOW DISK SPACE"

After: yellow badge with alert icon, reads "PAUSED (LOW SPACE)"

Before/after screenshots in #design.

## Test plan

- [ ] Open Downloads > Upcoming tab with a recording in paused_low_space status: badge reads **Paused (Low Space)** in yellow with an alert icon
- [ ] Open Scheduled Recordings > Upcoming tab for the same recording: badge is identical in text, color, and icon
- [ ] No regression on other badge states (Scheduled, Queued, etc.)

PR #139, opened 2026-06-07
docs: add changelog entries for PRs #130, #136, and #137; fix stale README badge color

razzamatazm:
## What a user would notice

The changelog now covers three fixes and improvements that merged since the last docs update. One stale description in the README was also corrected.

## Changes

**CHANGELOG.md** - Three new entries added under 2026-06-07:

- **PR #130 (fix):** Resuming an interrupted download no longer starts over from the beginning. Mustarrd now sends a Range request when recovering a partial download. If the provider supports it, the file continues from where it left off. If not, it falls back to a full re-download as before.
- **PR #136 (fix):** Hardware-accelerated transcoding (VAAPI, NVIDIA, AMD, Apple Silicon) with output format TS no longer silently overwrites recordings with empty files. Mustarrd now skips FFmpeg entirely for TS-to-TS output, which is always a no-op.
- **PR #137 (design):** Show titles on the Scheduled and Downloads pages now wrap to a second line on narrow screens instead of being cut off with "...".

**README.md** - Fixed one stale line in the "Monitor Downloads" section. The Upcoming tab low-space badge was described as "an orange badge" but PR #134 changed it to a yellow "PAUSED (LOW SPACE)" badge with an alert icon. Updated to match.

## How it was tested

Verified changelog entries against each PR description. Confirmed no em dashes in any changed text. README diff visually reviewed against the current UI badge appearance from PR #134.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Improved: Show titles no longer get cut off on phones
+
+**What you would notice:** On phone screens, if a show title was long and shared a row with a wide badge (like "PAUSED (LOW SPACE)"), the title was cut off with "..." making it impossible to read the full name. The title now wraps to a second line so the full show name is always visible.
+
+**What changed:** The program title on the Scheduled and Downloads pages now wraps instead of truncating when there is not enough horizontal space. No visual change on desktop screens.
+
+---
+
+### Fixed: Hardware-accelerated transcoding no longer overwrites recordings with empty files
+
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 7b2949b. Docs-only; Codex left no review.

Three changelog entries under 2026-06-07 (today) plus one README badge correction. Checked each against the merged code:

- #137 (titles wrap on narrow screens): accurate.
- #136 (TS + hw_accel zeroed the recording): accurate; matches the "skip ffmpeg for TS-to-TS with any hw_accel" fix and correctly scopes who was affected.
- #130 (Range resume): accurate. Honestly worded ("if the provider supports it... falls back to starting over"), which holds even given the 206-without-Content-Range edge case PR #142 just documented.
- README: changed "orange badge" to yellow "PAUSED (LOW SPACE)" with an alert icon. Verified against `Scheduled.jsx:60` (`color: 'yellow', icon: IconAlertCircle, label: 'Paused (Low Space)'`, rendered uppercase). Correct.

No concerns. Approving.

PR #141, opened 2026-06-07
fix: delete UserIdentity rows when deleting a user

razzamatazm:
## What happened

When an admin deleted a download-only user via **Settings > Users**, the `User` row was removed but the linked `UserIdentity` row (created when the user logged in via Plex) was left behind.

The next time that Plex account tried to log in, the login code found the orphaned identity, looked up the user it pointed to, got nothing back, and returned an HTTP 500 error. The user was permanently locked out with no way to recover short of editing the database directly.

## What changed

`delete_user` now runs `DELETE FROM user_identities WHERE user_id = <id>` before removing the `User` row, so no stale identity rows remain after a user is deleted.

## Before

1. Admin deletes Plex-linked user via the Users page.
2. Same Plex account tries to log in again.
3. Login returns 500 with "Plex identity is mapped to a missing user".
4. User is permanently locked out.

## After

1. Admin deletes Plex-linked user.
2. Same Plex account logs in again and is treated as a new user, going through the normal first-login flow.

## Testing

Added `backend/tests/test_delete_user_identity_cleanup.py` with four cases:

- `test_delete_user_returns_deleted_status`: `delete_user` still returns `{"status": "deleted"}`.
- `test_user_identity_rows_deleted_before_user`: `session.execute` is called twice (user lookup + identity delete) and `session.delete` is called once with the user.
- `test_identity_delete_precedes_user_delete`: confirms the identity delete happens before the user row is removed, so a crash mid-delete cannot leave orphaned identities.
- `test_user_not_found_raises_404`: 404 still raised when the user does not exist.

Full suite: 403 tests, all pass.

```
Ran 403 tests in 1.184s
OK
```

### DIFF
--- a/backend/api/admin_users.py
+++ b/backend/api/admin_users.py
@@ -134,6 +134,7 @@ async def delete_user(
         raise HTTPException(status_code=404, detail="User not found")
 
     await session.execute(delete(UserIdentity).where(UserIdentity.user_id == user.id))
+    await session.execute(delete(UserSetupToken).where(UserSetupToken.user_id == user.id))
     await session.delete(user)
     await session.commit()
     return {"status": "deleted"}
--- a/backend/api/admin_users.py
+++ b/backend/api/admin_users.py
@@ -4,12 +4,12 @@ from datetime import datetime, timedelta
 
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel, Field
-from sqlalchemy import select, func
+from sqlalchemy import select, delete, func
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from auth import require_admin, hash_password, AuthContext
 from database import get_session
-from models import User, UserSetupToken
+from models import User, UserIdentity, UserSetupToken
 
 
 router = APIRouter()
@@ -133,6 +133,7 @@ async def delete_user(
     if not user:
         raise HTTPException(status_code=404, detail="User not found")
 
+    await session.execute(delete(UserIdentity).where(UserIdentity.user_id == user.id))
     await session.delete(user)
     await session.commit()
     return {"status": "deleted"}
--- a/backend/tests/test_delete_user_identity_cleanup.py
+++ b/backend/tests/test_delete_user_identity_cleanup.py
@@ -9,101 +9,185 @@ permanently locked out with no way to recover short of direct DB surgery.
 
 After fix: delete_user executes DELETE FROM user_identities WHERE user_id = <id>
 before removing the User row, so stale identities cannot accumulate.
+
+The real-DB tests below verify the WHERE clause actually targets the correct
+user_id, not all rows. Creating two users (A and B) with separate identities and
+deleting only A confirms that B's identity survives.
 """
 import sys
 import unittest
 from pathlib import Path
-from types import SimpleNamespace
-from unittest.mock import AsyncMock, call
+from unittest.mock import AsyncMock
 
 BACKEND_ROOT = Path(__file__).resolve().parents[1]
 if str(BACKEND_ROOT) not in sys.path:
     sys.path.insert(0, str(BACKEND_ROOT))
 
-from sqlalchemy import delete as sa_delete
+from datetime import datetime, timezone
+
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_async_engine
 
 from api.admin_users import delete_user
-from models import User, UserIdentity
+from database import Base
+from models import User, UserIdentity, UserSetupToken
 
 
-def _scalar_one(value):
-    class R:
-        def scalar_one_or_none(self):
-            return value
-    return R()
+def _make_user(display_name="Plex User"):
+    return User(
+        role="download_only",
+        status="active",
+        display_name=display_name,
+        created_at=datetime.now(timezone.utc),
+        updated_at=datetime.now(timezone.utc),
+    )
 
 
-def _make_user(user_id=7):
-    u = User()
-    u.id = user_id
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 35e24f2. Codex left no review on this PR.

The core fix is correct. `UserIdentity.user_id` is a plain `ForeignKey("users.id")` with no `ondelete`, so deleting a user left the identity row behind, and `plex_login_complete` (auth.py:438) then raises `HTTP 500 "Plex identity is mapped to a missing user"` on every future Plex login: a permanent lockout. Deleting the identity rows before the user fixes that. Confirmed the full suite passes locally (403 tests OK).

Requesting changes on three points:

1. Blocking, test quality. `test_delete_user_identity_cleanup.py` mocks `session.execute` entirely, so it only asserts that execute was *called*. It would still pass if the DELETE targeted the wrong column or, worse, deleted every UserIdentity row regardless of user_id. For a destructive DELETE that is not enough. The repo already has the right pattern in `tests/test_delete_account_epg_cleanup.py` (real in-memory sqlite). Please add a real-DB test: create user A with an identity, user B with an identity, delete A, then assert A's identity is gone AND B's identity survives. That actually exercises the WHERE clause.

2. Blocking, scope/branch hygiene. This branch is stacked on PR #138: its diff carries #138's two EPG commits (31df0d9, 748f060) plus your delete-user commit. The PR title and changelog say "delete UserIdentity," but merging this would also merge the EPG fix under the wrong PR. Merge #138 first, then rebase this branch onto main so the diff is only the delete_user commit.

3. Non-blocking. `UserSetupToken.user_id` (models/user.py:69) has the same missing-cascade pattern and also orphans on user delete. It is harmless today because the setup-token endpoints return 404 (not 500) when the user is missing, but for consistency either delete those rows too in `delete_user` or add `ondelete="CASCADE"` to both FKs.

The ordering assertion in the test (identity delete before user delete) is fine but not load-bearing: both run in one transaction committed once, so they are atomic together regardless of order.

razzamatazm:
Addressed Cleo's review (de692d3):

**Blocking #1 (test quality):** Replaced all mock-based tests with real in-memory SQLite tests. The key test creates two users (A and B) each with a UserIdentity, deletes only A, and asserts:
- A's identity row is gone
- B's identity row is untouched

This actually exercises the WHERE clause, so a bug that deleted all identities or targeted the wrong column would fail.

**Blocking #2 (branch hygiene):** PR #138 is still open so I can't rebase yet. Once #138 merges, I'll rebase this branch so the diff shows only the delete_user change.

**Non-blocking (UserSetupToken):** Also added `DELETE FROM user_setup_tokens WHERE user_id = ?` in `delete_user`, with a passing regression test.

405 tests pass.

razzamatazm-cleo (changes requested):
Re-reviewed at head de692d3c.

Two of my three earlier points are resolved, thank you:

- Point 1 (test quality): fixed. `test_delete_user_identity_cleanup.py` now uses a real in-memory sqlite engine, seeds user A and user B with separate identities, deletes A, and asserts A's identity is gone AND B's identity survives. That genuinely exercises the WHERE clause instead of just asserting `execute` was called. Verified the `delete_user` signature (`_admin`, `session`) and the `role == "download_only"` SELECT filter both match what the test drives.
- Point 3 (UserSetupToken orphan): fixed. `delete_user` now also deletes `UserSetupToken` rows, and `DeleteUserSetupTokenCleanupTests` covers it.

One blocking point remains:

- Point 2 (branch hygiene / scope contamination): this branch is still stacked on PR #138. The diff carries #138's two EPG commits (31df0d9, 748f060) plus the EPG file `epg_ingest_manager.py` and `test_epg_duplicate_channel_name.py`. So merging this PR as-is would also merge the EPG fix under a PR whose title and changelog say only "delete UserIdentity rows." For a non-technical operator reading the PR, that description is inaccurate, and it would leave #138 as a redundant/empty PR. The EPG code itself is fine: it is the exact head I already approved on #138 (748f060), so no re-review of that logic is needed. The fix is ordering: merge #138 first, then rebase this branch onto main so its diff is only the delete_user/setup-token commit and its test.

Requesting changes solely to prevent an out-of-order merge. Once #138 is merged and this is rebased to a clean single-concern diff, this is good to approve.

razzamatazm:
Rebased onto main (5b2ca6a). PR #138 merged at ac6dc97 → EPG commits dropped. Diff now shows only the delete_user / UserIdentity / UserSetupToken fix.

razzamatazm-cleo (approved):
Re-reviewed at head 5b2ca6a. Codex left no review.

All three of my earlier points are now resolved:

- Point 1 (test quality): the test uses a real in-memory sqlite engine, seeds user A and user B with separate identities, deletes A, and asserts A is gone while B survives. This genuinely exercises the WHERE clause on a destructive DELETE.
- Point 2 (branch hygiene): with #138 merged, this branch is now rebased to a clean single-concern diff: only admin_users.py and the new test. The EPG commits are gone, so the PR contents match its title and changelog.
- Point 3 (UserSetupToken orphan): delete_user now also deletes UserSetupToken rows, covered by DeleteUserSetupTokenCleanupTests.

The core fix removes the orphaned UserIdentity row that caused plex_login_complete to raise HTTP 500 and permanently lock out the Plex user. CI green.

Non-blocking note for a future cleanup, not for this PR: scheduled_recording.requested_by_user_id and download.requested_by_user_id are also plain FKs to users.id with no cascade. They are nullable and only used for attribution, and SQLite does not enforce FKs by default, so they orphan harmlessly rather than 500. Worth tidying eventually with ondelete or matching deletes, but out of scope here.

PR #142, opened 2026-06-07
test: failing regression for 206 without Content-Range false resume

razzamatazm:
## What the operator notices

After a power cut or restart mid-download, a partially-downloaded catchup recording resumes instead of being corrupted. PR #130 shipped that fix. But a specific class of flaky IPTV provider can still corrupt the file on resume.

## Bug

When an IPTV catchup server responds to a Range request with HTTP 206 but **omits the Content-Range header** (an RFC 7233 violation that some cheap catchup implementations and transparent proxies do produce), the PR #130 resume guard silently passes:

```python
range_mismatch = range_start is not None and range_start != offset
# range_start is None (no header) => range_mismatch = False
resuming = response.status == 206 and offset > 0 and not range_mismatch
# => resuming = True, even though we have no position confirmation
```

The file is opened in `'ab'` mode. If the provider is actually sending from byte 0, the partial file grows from `offset` to `offset + full_size`, producing a corrupt `.ts` file that Plex and Jellyfin will reject.

## Reproduction steps

1. Start a large catchup download (unknown-length stream, no Content-Length).
2. Kill Mustarrd mid-download. A partial `.ts` file exists on disk.
3. Configure a mock provider that responds to Range requests with `HTTP 206` and no `Content-Range` header (sends the full stream from byte 0).
4. Restart Mustarrd.
5. The partial file is opened `'ab'` and the full stream is appended. Resulting file is `partial_size + full_size` bytes and plays as garbage.

**Expected:** treat absent Content-Range as "position unknown" and fall back to restart (`'wb'`, `downloaded=0`), same as a Content-Range start mismatch.

**Actual:** file is appended and returned as completed. Corrupt recording stored.

## Fix (not in this PR)

Change the resume guard in `_download_file`:

```python
# before
resuming = response.status == 206 and offset > 0 and not range_mismatch
# after
resuming = response.status == 206 and offset > 0 and range_start is not None and range_start == offset
```

## This PR

Adds one `@unittest.expectedFailure` test (`test_206_no_content_range_falls_back_to_restart`) to `test_download_file_range.py`. The test proves the bug is present (the code appends instead of restarting). Marked expected failure so CI stays green on main until the fix lands.

413 tests pass, 3 expected failures.

### DIFF
--- a/backend/tests/test_download_file_range.py
+++ b/backend/tests/test_download_file_range.py
@@ -384,5 +384,71 @@ class DownloadFileRangeTests(unittest.IsolatedAsyncioTestCase):
             os.unlink(tmp_path)
 
 
+    @unittest.expectedFailure
+    async def test_206_no_content_range_falls_back_to_restart(self):
+        """
+        BUG: A 206 response with no Content-Range header causes false resume.
+
+        Root cause: when the server returns 206 but omits the Content-Range header,
+        range_start stays None.  The mismatch guard is:
+            range_mismatch = range_start is not None and range_start != offset
+        Since range_start is None, range_mismatch is False, so resuming evaluates to
+        True.  The file is opened in 'ab' mode even though we have no confirmation
+        that the server is sending from the right position.  A buggy or proxied IPTV
+        provider that returns 206 from byte 0 (without Content-Range) would cause the
+        partial file to grow from offset+0 to offset+server_full_size, producing a
+        corrupt recording.
+
+        Expected fix: treat absent Content-Range as "position unknown" and fall back
+        to restart ('wb', downloaded=0), the same as a Content-Range mismatch.
+        Guard should be:
+            resuming = status==206 and offset>0 and range_start is not None
+                       and range_start == offset
+        """
+        existing = b"\x00" * 4096
+        # Provider returns 206 but NO Content-Range header (RFC violation, common in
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head af3a29b. Test-only PR; Codex left no review.

This adds a single `@unittest.expectedFailure` regression test for a real bug in the download resume path. I verified the bug against current source: in `download_manager._download_file` (download_manager.py:1041-1042),

    range_mismatch = range_start is not None and range_start != offset
    resuming = response.status == 206 and offset > 0 and not range_mismatch

When a provider returns 206 with no Content-Range header, `range_start` stays None, so `range_mismatch` is False and `resuming` is True. The partial file is opened in 'ab' and the server's bytes (from an unknown position) are appended, corrupting the recording. That is exactly what the test documents, and the fix it describes (require `range_start is not None and range_start == offset` to resume) is the right one.

The test is well-built: helper `_make_aiohttp_response(206, [...])` defaults `content_range=None`, matching the scenario, and the assertions check both the return value (must be new-bytes-only, not offset+new) and the file contents (must be overwritten, not appended). I ran the file locally: 8 tests, OK with `expected failures=1`, so the suite stays green, consistent with the PR #131 pattern.

No concerns. Approving. Worth filing/landing the actual guard fix as a follow-up so the marker can come off.

PR #143, opened 2026-06-07
[Design] Wrap upcoming recording title instead of truncating

razzamatazm:
## What changed

Downloads > Upcoming tab program titles now wrap to a second line instead of being cut off.

## Why

PR #137 removed `truncate` from program titles on the Scheduled Recordings page so long show names are always readable. The same fix was not applied to the Upcoming tab in Downloads.jsx. A long title like "EastEnders: The Final Farewell to Albert Square - Special Christmas Episode" appeared as "EastEnders: The Final Farewell to Albert S..." on mobile.

## What an operator will notice

On the Downloads > Upcoming tab, long show titles now wrap to a second line instead of being cut off. Short titles are unchanged.

## Change

One word removed from Downloads.jsx line 647: `truncate` removed from the program title Text component. Channel name keeps `truncate` since it is secondary info. No logic touched.

## How it was tested

Playwright screenshots at 390px mobile width with a long title. Before: title truncated. After: title wraps. Before/after screenshots in #design.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -644,7 +644,7 @@ export default function Downloads() {
                 <Card key={rec.id} shadow="sm" padding="md" radius="md" withBorder>
                   <Group justify="space-between" wrap="nowrap">
                     <Stack gap={2} style={{ flex: 1, overflow: 'hidden' }}>
-                      <Text fw={500} truncate>{rec.program_title}</Text>
+                      <Text fw={500}>{rec.program_title}</Text>
                       <Text size="sm" c="dimmed" truncate>{rec.channel_name}</Text>
                       <Group gap="xs">
                         <Text size="xs" c="dimmed">


razzamatazm-cleo (approved):
Reviewed at head abf95f17. No Codex review present.

Single-line frontend change: drops `truncate` from the upcoming-recording title `Text` so long program titles wrap to multiple lines instead of being cut off. This matches the design intent and is safe:

- The parent `Stack` has `overflow: 'hidden'` but no fixed height, so wrapping just grows the card vertically rather than clipping. Wrapping is strictly more readable than the previous truncation.
- `channel_name` keeps its `truncate`, which is fine: the channel is secondary and a single wrapped title is the goal here.
- No backend, no logic, no API surface touched. Frontend has no test harness in this repo, so no test is expected.

One minor, non-blocking note: a single very long unbroken token (no spaces) could still overflow horizontally and be clipped by the parent `overflow: 'hidden'`, since normal whitespace wrapping cannot break it. That is not a regression (truncate clipped it too) and program titles realistically contain spaces, so it is not worth holding the PR. If you ever want to harden it, `style={{ wordBreak: 'break-word' }}` on the title would cover that case.

No concerns. Approving.

PR #146, opened 2026-06-07
fix: treat absent Content-Range on 206 as restart, not resume

razzamatazm:
## What happened

When a partial download was recovered on restart and Mustarrd sent a `Range: bytes=N-` request, some IPTV providers returned HTTP 206 but left out the `Content-Range` header. Without that header, Mustarrd could not verify where the provider actually started sending, but it treated the 206 as a confirmed resume anyway. The file was opened in append mode, so if the provider sent from byte 0, the recording grew to `partial_size + full_size` bytes and played as corrupted video.

## Root cause

In `_download_file` (download_manager.py), the resume guard was:

```python
range_mismatch = range_start is not None and range_start != offset
resuming = response.status == 206 and offset > 0 and not range_mismatch
```

When `Content-Range` is absent, `range_start` stays `None`, so `range_mismatch` is `False`, and `resuming` is `True`. No Content-Range means position is unknown, but the old code treated it as position confirmed.

## What changed

The guard now requires `range_start` to be present and equal to the requested offset before treating a 206 as a resume:

```python
resuming = (
    response.status == 206
    and offset > 0
    and range_start is not None
    and range_start == offset
)
```

Absent or mismatched `Content-Range` both fall back to restart (`wb` mode, re-download from byte 0), the same safe path as a 200 response.

## Before

1. Partial download on disk (say, 4 096 bytes of a 10 000-byte recording).
2. Mustarrd restarts, sends `Range: bytes=4096-`.
3. Provider returns `HTTP 206` with no `Content-Range` header, body starting at byte 0.
4. File opened `'ab'`. Provider bytes appended. File grows to 14 096 bytes: corrupt recording.
5. No warning logged.

## After

1. Same scenario.
2. Mustarrd restarts, sends `Range: bytes=4096-`.
3. Provider returns `HTTP 206` with no `Content-Range` header.
4. `range_start` is `None`. Resume guard fails. File opened `'wb'` (restart).
5. Log: "Provider did not honour Range request; re-downloading from start."
6. Clean re-download. No corruption.

## Testing

Removed the `@unittest.expectedFailure` decorator from `test_206_no_content_range_falls_back_to_restart` in `test_download_file_range.py`, which was added in PR #142 to document this exact bug. The test now passes.

Full suite: 413 tests, all pass (2 unrelated expected failures from `test_epg_duplicate_channel_name.py` pending PR #138).

```
Ran 413 tests in 1.225s
OK (expected failures=2)
```

Closes the [206 without Content-Range triggers false resume, corrupting partial catchup files](https://discord.com/channels/1512584962413506710/1513245573069930577) task.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1039,7 +1039,12 @@ class DownloadManager:
                 # If Content-Range is present but starts elsewhere, the provider silently
                 # returned data from a different position; appending would corrupt the file.
                 range_mismatch = range_start is not None and range_start != offset
-                resuming = response.status == 206 and offset > 0 and not range_mismatch
+                resuming = (
+                    response.status == 206
+                    and offset > 0
+                    and range_start is not None
+                    and range_start == offset
+                )
 
                 if total_size > 0:
                     await self._broadcast_log(
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -988,6 +988,77 @@ class DownloadManager:
         except Exception:
             pass
 
+    async def _stream_response_to_file(
+        self,
+        response,
+        output_path: str,
+        file_mode: str,
+        downloaded_start: int,
+        total_size: int,
+        download_id: int,
+        session: AsyncSession,
+    ) -> int:
+        """Write the body of *response* to *output_path* and return total bytes accumulated.
+
+        *downloaded_start* is added to the byte counter before the first write so
+        a resumed download reports cumulative progress from the start of the file.
+        """
+        if total_size > 0:
+            await self._broadcast_log(
+                download_id,
+                f"HTTP {response.status}. Expected size: {total_size:,} bytes."
+            )
+        else:
+            await self._broadcast_log(
+                download_id,
+                f"HTTP {response.status}. Size unknown; streaming download."
+            )
+
+        downloaded = downloaded_start
+        last_progress_update = 0
+
+        async with aiofiles.open(output_path, file_mode) as f:
+            async for chunk in response.content.iter_chunked(1024 * 1024):  # 1MB chunks
+                if download_id in self._cancelled:
+                    raise asyncio.CancelledError()
+
+                await f.write(chunk)
+                downloaded += len(chunk)
+
+                if total_size > 0:
+                    progress = (downloaded / total_size) * 100
+                else:
+                    progress = 0
+
+                if progress - last_progress_update >= 1 or downloaded == total_size:
+                    last_progress_update = progress
+
+                    await session.execute(
+                        update(Download)
+                        .where(Download.id == download_id)
+                        .values(
+                            progress=progress,
+                            downloaded_bytes=downloaded,
+                            file_size=total_size
+                        )
+                    )
+                    await session.commit()
+
+                    await self._broadcast_progress(
+                        download_id,
+                        progress,
+                        DownloadStatus.DOWNLOADING.value,
+                        downloaded_bytes=downloaded,
+                        file_size=total_size,
+                        download_progress=progress
+                    )
+
+        await self._broadcast_log(
+            download_id,
+            f"Download bytes written: {downloaded:,}."
+        )
+        return downloaded
+
     async def _download_file(
         self,
         url: str,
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Requesting changes. I agree with Codex's P2 finding, and it is the core problem with this PR.

The guard change itself is correct: an unverified 206 (no Content-Range, or Content-Range starting elsewhere) should no longer set `resuming = True`, so we stop appending to the existing prefix. That fixes the oversized-file corruption.

But the restart path it falls into does not actually restart the download. It reuses the response object that is already open and writes its body with `file_mode = "wb"`. A real 206 body contains only the bytes from the server's chosen offset onward (the tail), not the full file from byte 0. So in the realistic case where a provider honors the Range but omits Content-Range, we overwrite the partial file with just the tail, then `downloaded` climbs to `total_size` (which is now the partial `content_length`) and the recording is marked completed. That is a silently truncated recording: arguably worse than the oversized file, because nothing flags it as broken.

The same flaw exists in the pre-existing `range_mismatch` branch, but this PR extends the behavior to the no-Content-Range case, so it should fix it properly rather than inherit it.

What is needed: when a 206 cannot be verified as starting at the requested offset, abandon the current response and issue a fresh GET without the `Range` header (true restart from byte 0), then write `wb`. Only then is `wb` safe.

Test gap: `test_206_no_content_range_falls_back_to_restart` mocks the 206 body as if it were the full file (`new_data` = full content), so it never exercises the real failure mode. A faithful test would have the 206 return only the tail bytes (from `offset` onward) and assert that the final file equals the full content from byte 0, which would catch the truncation. Converting the test from `expectedFailure` to a real regression test is the right move; it just needs a realistic 206 body.

Net: good direction, but as written it swaps one corruption mode for another. Please make the unverified-206 path re-request from byte 0.

razzamatazm:
Addressed Cleo's feedback (commit 8f6b8e3).

**Root cause Cleo identified:** The guard fix correctly set `resuming=False` and `file_mode='wb'` for an unverifiable 206, but still consumed the body from the open response. That body starts at the requested offset, not byte 0, so writing it with `wb` produced a silently truncated file (tail bytes only, marked complete).

**What changed:**

1. Both cases that produce an unverifiable 206 (absent Content-Range, mismatched Content-Range start) now release the connection without reading the body, then issue a fresh `GET` with no `Range` header. That response is guaranteed to start at byte 0, so `wb` is safe.

2. The streaming loop is extracted into `_stream_response_to_file` so the progress/DB/broadcast logic is not duplicated across both code paths.

3. Tests updated with a realistic two-response mock: the first `get()` returns only the tail bytes (what the provider actually sends for a partial request), the restart `get()` returns the full content from byte 0. The assertions now verify the final file equals the full content, which would have caught the truncation before the fix.

413 tests pass, 2 expected failures unchanged.

razzamatazm-cleo (approved):
Re-review at head 8f6b8e37 (my prior CHANGES_REQUESTED was at 954a38f0).

The new commit "re-request from byte 0 when 206 start position is unverifiable" fully resolves the truncation concern I raised and the Codex P2 finding. Tracing the control flow:

- resuming is now true only when status==206 AND offset>0 AND range_start==offset, so a verified resume still appends ('ab', cumulative byte count). Correct.
- An unverified 206 (no Content-Range, or Content-Range starting elsewhere) sets needs_restart and falls out of the first `async with response` block WITHOUT reading the body, releasing the connection. It then issues a fresh GET with no Range header and writes 'wb' from byte 0. This is a true restart, not the old reuse-the-tail-as-wb bug. Correct.
- 200 / offset==0 paths unchanged (wb from 0).

No header regression: request_headers is only ever {} plus an optional Range, and Xtream provider auth lives in the URL, so dropping the Range header on the restart GET does not strip auth or anything else.

Tests: both rewritten to a realistic tail-only 206 body via _make_aiohttp_client_session_multi and now assert the final file equals full_content from byte 0. The expectedFailure marker is gone; these are real regression tests that would catch the truncation. Good.

Minor, non-blocking: the multi-response mock returns full_content on the second get() regardless of headers, so the tests do not assert the restart GET actually omits the Range header. The code clearly does (http_session.get(url) with no headers), but a future regression that re-added a Range would not be caught. Consider asserting mock_session.get.call_args_list[1] carries no Range. Not required for this PR.

Net: this is the right fix. Approving.

PR #149, opened 2026-06-07
fix: reject text/* responses from provider as error pages

razzamatazm:
## What happened

When an IPTV provider returns HTTP 200 with a `text/html` body (for example, "Catchup not available" or "Session limit reached"), the downloader was treating it as a valid stream. The HTML was written byte-for-byte into the `.ts` output file, and the recording was marked **COMPLETED**.

A non-technical user would see a completed recording that plays as garbage, with no indication anything went wrong.

## What changed

`_download_file` now checks `Content-Type` immediately after the status check. If the type starts with `text/` (covers `text/html`, `text/plain`, and similar), the download is aborted with a clear error:

> "Provider returned an error page (Content-Type: text/html). The catchup window may have expired or be unavailable."

`video/*`, `application/octet-stream`, and missing `Content-Type` are all unchanged. Many IPTV providers omit `Content-Type` entirely and that still works.

## Before

1. Queue a catchup download for a channel whose provider returns `HTTP 200 Content-Type: text/html` with body `<html>Catchup not available</html>`.
2. Download reaches COMPLETED. Output file contains the HTML page.
3. No error shown to the user.

## After

1. Same request.
2. Download is marked FAILED with message: "Provider returned an error page (Content-Type: text/html). The catchup window may have expired or be unavailable."
3. User sees a clear failure reason.

## Testing

New test file `backend/tests/test_download_html_body.py` with 6 cases:

- `test_html_200_raises_exception`: HTTP 200 + `text/html` raises with "error page" in the message.
- `test_html_206_raises_exception`: HTTP 206 + `text/html` also raises.
- `test_text_plain_raises_exception`: `text/plain` is also rejected.
- `test_video_mpeg_not_rejected`: `video/mpeg` completes normally.
- `test_octet_stream_not_rejected`: `application/octet-stream` completes normally.
- `test_no_content_type_not_rejected`: Absent `Content-Type` does not block the download.

Full suite: 419 tests pass (3 expected failures for unrelated open bugs in PRs #138 and #146).

```
Ran 419 tests in 1.256s
OK (expected failures=3)
```

Closes: https://discord.com/channels/1512584979490906232/1512684030368415916

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1015,6 +1015,13 @@ class DownloadManager:
                 if response.status not in (200, 206):
                     raise Exception(f"HTTP {response.status}: {response.reason}")
 
+                content_type = response.headers.get("Content-Type", "")
+                if content_type.lower().startswith("text/"):
+                    raise Exception(
+                        f"Provider returned an error page (Content-Type: {content_type}). "
+                        "The catchup window may have expired or be unavailable."
+                    )
+
                 total_size = response.content_length or 0
                 range_start = None
                 if response.status == 206:
--- a/backend/tests/test_download_html_body.py
+++ b/backend/tests/test_download_html_body.py
@@ -1,0 +1,216 @@
+"""
+Regression test: provider returns HTTP 200 with a text/html body (error page).
+
+Before the fix, _download_file would write the HTML to the output file and mark
+the download COMPLETED. With post-processing off the user ended up with an HTML
+file that looked like a completed recording. With post-processing on, FFmpeg
+failed on the HTML input and the download ended as COMPLETED with a warning.
+
+After the fix, _download_file raises an exception as soon as it sees a text/*
+Content-Type so the download is marked FAILED with a clear message.
+"""
+
+import asyncio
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approved at head 5c01316e. Codex left no review.

The fix is correct and well-placed: the Content-Type guard sits right after the 200/206 status check and before any bytes are written, so a text/* error page raises and the download is marked FAILED instead of leaving an HTML file that looks like a finished recording. Case-insensitive `.lower().startswith("text/")` handles `text/html; charset=utf-8` and friends.

Test coverage is strong for a behavior change like this: it asserts the reject path (html 200, html 206, text/plain) AND the must-not-break paths (video/mpeg, application/octet-stream, and absent Content-Type, which many providers omit). The absent-header case is the important false-positive guard and it is covered.

Edge cases I checked and am comfortable with: HLS playlists (application/vnd.apple.mpegurl) are out of scope and not a regression here; aiohttp follows redirects before this point so a 200 text/* really is the final body. No remaining concerns.

PR #152, opened 2026-06-07
fix: block setup link from reactivating disabled users

razzamatazm:
## What a user would notice

If an admin disabled a user account, that user could still log back in within 24 hours by using their existing setup link. The disable had no effect until the link expired on its own.

## What changed

Two small guards were added:

1. **`POST /api/auth/user/setup/{token}`** now returns HTTP 403 if the user is disabled. The setup link is not consumed and the user's status stays disabled.
2. **`POST /api/admin/users/{id}/setup-link`** now returns HTTP 400 if the admin tries to generate a new link for an already-disabled user.

Active and pending users are unaffected.

## Root cause

`complete_user_setup` (auth.py) set `user.status = \"active\"` unconditionally, overwriting a deliberate \"disabled\" status. `generate_setup_link` (admin_users.py) had no status check, so admins could unknowingly issue a 24-hour reactivation window for a disabled account.

## How it was tested

6 new real-DB regression tests in `backend/tests/test_setup_link_disabled_user.py`:

- Disabled user gets HTTP 403 from `complete_user_setup`
- Disabled user's status stays \"disabled\" after the attempt
- Pending user is still activated (regression guard)
- Active user (password reset via link) still works (regression guard)
- Disabled user gets HTTP 400 from `generate_setup_link`
- Active user can still get a setup link (regression guard)

Full suite: 421 tests pass.

**Before:** `POST /api/auth/user/setup/TOKEN` with a valid token and a disabled user returned `{"status": "configured"}` and set the user active.

**After:** Same request returns HTTP 403 `{"detail": "Account is disabled"}`. User status unchanged.

### DIFF
--- a/backend/api/admin_users.py
+++ b/backend/api/admin_users.py
@@ -148,6 +148,8 @@ async def generate_setup_link(
     user = result.scalar_one_or_none()
     if not user:
         raise HTTPException(status_code=404, detail="User not found")
+    if user.status == "disabled":
+        raise HTTPException(status_code=400, detail="Cannot generate a setup link for a disabled user")
     token = await _create_setup_token(session, user_id)
     return {
         "token": token,
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -600,10 +600,13 @@ async def complete_user_setup(
     user = user_result.scalar_one_or_none()
     if not user:
         raise HTTPException(status_code=404, detail="User not found")
+    if user.status == "disabled":
+        raise HTTPException(status_code=403, detail="Account is disabled")
 
     user.password_hash = hash_password(payload.password)
     user.password_reset_required = False
-    user.status = "active"
+    if user.status != "active":
+        user.status = "active"
     user.updated_at = datetime.utcnow()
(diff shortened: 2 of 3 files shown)


razzamatazm-cleo (approved):
Reviewed at head f200d23. Codex left no review.

The fix is correct and well targeted. Two doors were open for a disabled user to come back to life through a setup link:

1. generate_setup_link now returns 400 for a disabled user, so an admin cannot accidentally mint a fresh reactivation window.
2. complete_user_setup now returns 403 before touching the row if the user is disabled, which closes the real hole: a link issued while the user was still active, then used after the admin disabled them. The 403 fires before row.used_at is set and before any commit, so the token is left untouched rather than silently consumed, which is the right behavior.

The change from unconditional user.status = "active" to the guarded "if user.status != active" is a harmless no-op now that disabled is rejected above. Pending users still activate, active users (password-reset flow) stay active.

Tests are solid: real in-memory sqlite, covering disabled gets 403, disabled status stays disabled, pending activates, active stays active, generate_setup_link 400 for disabled and still works for active. CI green.

PR #156, opened 2026-06-07
fix: epg_offset_minutes now applies to guide display times

razzamatazm:
## What a user would notice

Setting **EPG Offset (minutes)** in Settings had no visible effect. A user correcting a provider timezone mismatch (e.g. setting +60 to shift programs by one hour) would see program times unchanged after saving.

## Root cause

`AppSettings.epg_offset_minutes` was stored to the database and cleared the EPG cache on save, but `EPGService` never read it. The only offset applied was the per-account **Guide Offset Hours**, which requires per-account configuration. The global setting appeared functional but was a silent no-op.

## What changed

`epg_offset_minutes` from `AppSettings` is now read in `get_epg_for_channel` and passed as `global_offset_minutes` to `_resolve_display_window`, where it is combined with the per-account `guide_offset_hours`. The same value is passed to `serialize_program` in the EPG search endpoint.

The default is 0, so existing behavior is unchanged when the setting is not configured.

Affects:
- Browse EPG channel guide (live API and DB paths)
- EPG search results

## How it was tested

Six regression tests in `test_epg_offset_minutes.py`:

Before the fix, `_resolve_display_window` with `global_offset_minutes=60` returned unchanged times. After the fix:

```
# global_offset_minutes=60, program at 20:00-21:00 UTC
# Expected: 21:00-22:00
start_time: "2024-01-15T21:00:00+00:00"  # shifted +60 min
end_time:   "2024-01-15T22:00:00+00:00"
```

Negative offsets, combined per-account + global offsets, and None account all tested. 430 tests pass.

### DIFF
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -53,7 +53,10 @@ async def search_epg(
     pattern = _build_like_pattern(q.lower())
     now_utc = datetime.now(timezone.utc)
     offset_hours = int(account.guide_offset_hours or 0)
-    threshold = now_utc - timedelta(hours=offset_hours)
+    settings_result = await session.execute(select(AppSettings))
+    db_settings = settings_result.scalar_one_or_none()
+    global_offset_minutes = int(getattr(db_settings, "epg_offset_minutes", 0) or 0)
+    threshold = now_utc - timedelta(hours=offset_hours) - timedelta(minutes=global_offset_minutes)
     result = await session.execute(
         select(EPGProgram)
         .where(
@@ -72,9 +75,6 @@ async def search_epg(
         .offset(offset)
     )
     programs = result.scalars().all()
-    settings_result = await session.execute(select(AppSettings))
-    db_settings = settings_result.scalar_one_or_none()
-    global_offset_minutes = int(getattr(db_settings, "epg_offset_minutes", 0) or 0)
     return [epg_service.serialize_program(row, account, global_offset_minutes) for row in programs]
 
 
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -8,7 +8,7 @@ from sqlalchemy import select, or_, func
 
 from auth import require_admin_or_download_user, AuthContext
 from database import get_session
-from models import XtreamAccount, EPGProgram
+from models import XtreamAccount, EPGProgram, AppSettings
 from services.epg_service import epg_service
 from services.epg_ingest_manager import epg_ingest_manager
 
@@ -72,7 +72,10 @@ async def search_epg(
         .offset(offset)
     )
     programs = result.scalars().all()
-    return [epg_service.serialize_program(row, account) for row in programs]
+    settings_result = await session.execute(select(AppSettings))
+    db_settings = settings_result.scalar_one_or_none()
+    global_offset_minutes = int(getattr(db_settings, "epg_offset_minutes", 0) or 0)
+    return [epg_service.serialize_program(row, account, global_offset_minutes) for row in programs]
 
 
 @router.get("/epg/status")
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1086,6 +1086,16 @@ class DownloadManager:
             needs_restart = False
 
             async with http_session.get(url, headers=request_headers) as response:
+                if response.status == 416 and offset > 0:
+                    # Range past EOF: the partial file on disk is already complete.
+                    # Returning offset (non-zero) lets _execute_download move it to
+                    # the completed folder without deleting it.
+                    await self._broadcast_log(
+                        download_id,
+                        "Provider returned 416 (Range Not Satisfiable); "
+                        "file already complete on disk.",
+                    )
+                    return offset
                 if response.status not in (200, 206):
                     raise Exception(f"HTTP {response.status}: {response.reason}")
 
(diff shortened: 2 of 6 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 1aa19f83. This branch is stacked on PR #154: its first two commits (63d06fc, e174797) are the 416 download fix, and only the third commit (1aa19f8) is this PR's own work, the epg_offset_minutes wiring. The diff therefore carries #154's code wholesale.

The epg_offset_minutes change itself is correct and well-tested. Verified:
- _resolve_display_window now adds timedelta(hours=guide_offset_hours, minutes=global_offset_minutes); the "if total_offset:" guard keeps a zero offset a no-op.
- global_offset_minutes is loaded from AppSettings and threaded through serialize_program, _process_epg_entry, get_epg_for_channel, and api/epg.py search_epg. I grepped every caller; none was missed.
- The offset is display-only: serialize_program keeps start_timestamp/stop_timestamp as the raw row values, so catchup download targeting and the cutoff filters in get_past_programs / _filter_programs_by_cutoff are unaffected. That matches the setting's stated purpose (guide display times). duration_minutes is unchanged since both ends shift equally.
- The 6 new tests in test_epg_offset_minutes.py pass locally, including the combined account + global offset case and the None-account case.

Requesting changes for one reason: because this branch ships #154's 416 code, the two restart-path issues I raised on #154 also block here:
1. (blocking) the byte-0 re-request at download_manager.py:1199-1207 has no text/* error-page guard, so a provider error page can be written as a completed .ts file.
2. (should fix) a 416 with a text/* body raises instead of restarting in the oversized-local case.

Once #154 is fixed (and ideally merged first so this branch only carries the epg commit), the epg_offset_minutes work here is good to approve as-is. No issues with this PR's own diff.

razzamatazm-cleo (changes requested):
Reviewed at head 31d58fd9. The epg_offset_minutes display work is correct and well-tested, and this branch now carries the fixed #154 commits (494cf56, b9d6438), so my two earlier #154 blockers no longer apply here. I confirmed the rebase did not alter the epg work: the net epg_service.py change is byte-identical to the version I reviewed at 1aa19f83.

Verified again:
- _resolve_display_window adds timedelta(hours=guide_offset_hours, minutes=global_offset_minutes); the zero-offset guard keeps it a no-op.
- global_offset_minutes is loaded from AppSettings and threaded through serialize_program and every caller I grepped; offsets are display-only (start_timestamp/stop_timestamp stay raw), so catchup targeting and cutoff filters are unaffected.

Requesting changes for one remaining item, agreeing with Codex's P2 on api/epg.py: the new global offset is applied to serialize_program (display) but NOT to the /epg/search query threshold. The threshold at epg.py:54 is still `now_utc - timedelta(hours=guide_offset_hours)`, computed before AppSettings is read, and the settings read was placed after the query. So for non-archive rows within |epg_offset_minutes| of now, search includes/excludes programs by raw end_time while showing shifted display times: a program displayed as still upcoming (positive offset) can be filtered out, and a negative offset can keep display-past programs in. The blast radius is narrow (non-archive rows only; has_archive rows bypass the threshold, and those are what catchup users mostly search), but it directly contradicts this PR's stated goal of making the offset apply to guide display surfaces, and search is one.

Fix is small: read epg_offset_minutes before building the query and fold it into the threshold, e.g. `threshold = now_utc - timedelta(hours=guide_offset_hours) - timedelta(minutes=global_offset_minutes)`, and add a test for a non-archive row whose raw end is just past the cutoff but whose displayed end is shifted into the future. Everything else here is good to approve once that lands.

razzamatazm-cleo (approved):
Reviewed at head ba06b9b. Approving. My one remaining blocker from the prior review (31d58fd9) is now fixed.

The search-threshold gap is resolved: api/epg.py now reads epg_offset_minutes before building the query and folds it in: `threshold = now_utc - timedelta(hours=offset_hours) - timedelta(minutes=global_offset_minutes)`, and serialize_program receives the same value. Display window and search filter are now consistent. Two new tests cover it: test_global_offset_minutes_keeps_near_threshold_program (raw end 30 min ago, +60 offset, kept) and test_global_offset_minutes_excludes_too_old_program (end 2h ago, +60 offset, excluded). Both directions checked.

Note on diff scope: PR #154 merged to main at 23:34 today, so this branch is now behind main. The download_manager.py 416 changes and test_download_file_range.py additions shown in this diff are merge-base noise: I fetched both files at this head and at main and they are byte-identical, so merging introduces no change to them and cannot regress the merged #154 fix. The net new code here is solely the epg_offset_minutes wiring. A rebase would shrink the diff but is not required for safety.

Re-verified the epg work itself:
- _resolve_display_window adds timedelta(hours=guide_offset_hours, minutes=global_offset_minutes); the `if total_offset:` guard keeps a zero offset a no-op.
- global_offset_minutes is loaded from AppSettings (getattr default 0, so safe even if the column is absent) and threaded through serialize_program, _process_epg_entry, get_epg_for_channel, and search_epg. No caller missed.
- Offset is display-only: start_timestamp/stop_timestamp stay raw, so catchup targeting and the get_past_programs / _filter_programs_by_cutoff cutoffs are unaffected. duration_minutes unchanged since both ends shift equally.
- 6 tests in test_epg_offset_minutes.py cover zero, +60, negative, combined account+global, None account, and serialize_program.

No remaining concerns.

PR #157, opened 2026-06-07
Add files via upload

razzamatazm:


razzamatazm-cleo (approved):
Reviewed at head fab340e. Single binary asset change: `frontend/src/assets/mustarrdlogo.png` (1500x1024, 2.1MB -> 1121x328, 568KB).

Checked for distortion risk: the only reference to this file anywhere in the repo is the README header banner (GitHub blob URL). It is not imported or rendered by any app code (`frontend/src` has no `mustarrdlogo` reference), so the new wider 3.4:1 aspect ratio cannot distort any fixed-size UI element. Smaller file is a plus for the README.

Valid PNG, no other files touched. No concerns. Approving.

PR #158, opened 2026-06-07
fix: ComSkip format options blocked first-time setup

razzamatazm:
## What changed

The \"MKV + skip commercials\" and \"MP4 + skip commercials\" options in the Recording Format dropdown were disabled whenever the comskip binary was not found in the default PATH. The binary path field only appears **after** one of those formats is selected, so there was no way to enter a custom binary path. Chicken-and-egg: you need to select a ComSkip format to see the path field, but the format was grayed out until comskip was already installed at a known path.

Fix: remove `!comskipReady` from the disabled condition. ComSkip format options are now disabled only when **ffmpeg** is unavailable (same as the other ffmpeg-requiring formats). When ffmpeg is present but comskip is not in PATH (the common first-time setup scenario), the options are selectable.

Also updated the alert message from \"Comskip is unavailable\" to \"Comskip binary not found in the default PATH. Enter the path to your comskip binary below...\" so users know what to do.

## What a user sees

**Before:** Opening the dropdown showed \"MKV + skip commercials\" and \"MP4 + skip commercials\" grayed out. Clicking them did nothing. No binary path field ever appeared.

**After:** The ComSkip format options are selectable when ffmpeg is available. Selecting one reveals the Comskip Binary Path and Comskip INI Path fields, plus an alert explaining the binary was not found and where to enter a custom path.

## Files changed

- `frontend/src/pages/Settings.jsx`: 4 lines changed (disabled condition simplified, alert message improved)

## How tested

Reproduced the bug on a local dev instance with no comskip binary in PATH. Verified the format options were grayed out and the binary path field was inaccessible. Applied fix, forced comskip_enabled=true in the dev DB to confirm the panel renders correctly with the updated alert message. Screenshots in the #design channel.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -826,9 +826,7 @@ export default function Settings() {
               value: f.value,
               label: f.label,
               disabled:
-                ['remux_mkv', 'remux_mp4', 'transcode_mkv', 'transcode_mp4'].includes(f.value) && !ffmpegReady
-                  ? true
-                  : ['transcode_comskip', 'transcode_mp4_comskip'].includes(f.value) && (!ffmpegReady || !comskipReady)
+                ['remux_mkv', 'remux_mp4', 'transcode_mkv', 'transcode_mp4', 'transcode_comskip', 'transcode_mp4_comskip'].includes(f.value) && !ffmpegReady
                   ? true
                   : false,
             }))}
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head 5ef60e6. No Codex review present.

Verified the fix is correct and coherent:
- The `disabled` condition now gates `transcode_comskip`/`transcode_mp4_comskip` on `!ffmpegReady` only, matching the other ffmpeg-requiring formats. This resolves the chicken-and-egg trap: the Comskip Binary Path field renders on format selection (Settings.jsx, gated on `currentFormat` only, not `comskipReady`), so a user can now actually reach it.
- `comskipReady` is still used by the updated yellow alert, so no dead variable. The alert text now tells the user exactly what to do.

Non-blocking notes:
1. A user can now select a Comskip format, leave the path blank with no binary in PATH, and save. That persists a config the post-processor cannot satisfy (recordings would fail at the ComSkip step). The alert guides them to enter a path, and the prior behavior made configuration impossible at all, so this is a net improvement, not a regression. Worth a follow-up to surface a clearer error if ComSkip runs with no resolvable binary.
2. No regression test. Repo has no frontend test harness, so this matches the established exception for frontend-only changes.

PR #159, opened 2026-06-07
fix: recover_incomplete_downloads detects chunked-complete files

razzamatazm:
## Summary

After a hard crash, a fully-downloaded chunked recording (no `Content-Length`) was always re-queued for re-download instead of being recognized as complete. This PR fixes the root cause in `recover_incomplete_downloads` and was preceded by PR #154 which fixed the acute data-loss symptom (416 on resume deleting the file).

**Root cause:** The `is_complete` check required `download.file_size` to be truthy. Chunked streams keep `file_size=0` throughout the download, so the expression short-circuited to `False` for every chunked download regardless of whether the file was actually complete.

**Fix:** When `file_size == 0`, compare on-disk size to `downloaded_bytes`. If they match (both non-zero), the download finished cleanly before the crash and is moved to completed. If `disk_size != downloaded_bytes`, the file is partial and is re-queued as before.

**Files changed:**
- `backend/services/download_manager.py`: `is_complete` check now handles chunked streams
- `backend/tests/test_recover_chunked_complete_range416.py` (Gremlin-authored): regression tests — both now pass
- `backend/tests/test_recover_unknown_length.py`: updated partial-file test to use `disk_size < downloaded_bytes` so the partial scenario is unambiguous after the new detection logic

## Before

A complete chunked recording crashed mid-status-update:
1. Restart: `file_size=0`, `downloaded_bytes=8192`, disk=8192
2. `is_complete` = False (file_size=0 short-circuits)
3. Status → PENDING, re-queued
4. Next attempt: `Range: bytes=8192-` → 416 → file preserved (PR #154) but entire recording re-downloaded unnecessarily, risks catchup window expiry

## After

Same scenario:
1. Restart: `file_size=0`, `downloaded_bytes=8192`, disk=8192
2. `is_complete` = True (`disk_size == downloaded_bytes > 0`)
3. Status → COMPLETED (or PROCESSING if post-processing configured), no re-download

## Test plan

- `python -m unittest tests/test_recover_chunked_complete_range416.py` — 2 tests pass (complete and partial scenarios)
- `python -m unittest tests/test_recover_unknown_length.py` — 3 tests pass (no regressions on partial-file preserve-bytes behavior)
- `python -m unittest discover -s tests` — 423 tests pass

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1061,6 +1061,17 @@ class DownloadManager:
                         download_progress=progress
                     )
 
+        # For chunked streams (total_size == 0) the progress-based commit inside
+        # the loop never fires, so downloaded_bytes stays 0 in the DB. Commit the
+        # final byte count here so recovery after a crash can detect a complete file.
+        if total_size == 0:
+            await session.execute(
+                update(Download)
+                .where(Download.id == download_id)
+                .values(downloaded_bytes=downloaded)
+            )
+            await session.commit()
+
         await self._broadcast_log(
             download_id,
             f"Download bytes written: {downloaded:,}."
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -429,11 +429,19 @@ class DownloadManager:
 
                 if download.status == DownloadStatus.DOWNLOADING.value:
                     try:
-                        is_complete = (
+                        if input_file.is_file() and download.file_size:
+                            is_complete = download.downloaded_bytes >= int(download.file_size)
+                        elif (
                             input_file.is_file()
-                            and download.file_size
-                            and download.downloaded_bytes >= int(download.file_size)
-                        )
+                            and download.file_size == 0
+                            and download.downloaded_bytes > 0
+                        ):
+                            # Chunked stream (no Content-Length). The DB records
+                            # downloaded_bytes as written; if on-disk size matches, the
+                            # download finished cleanly before the crash.
+                            is_complete = input_file.stat().st_size == download.downloaded_bytes
+                        else:
+                            is_complete = False
                     except Exception:
                         is_complete = False
 
--- a/backend/tests/test_recover_chunked_complete_range416.py
+++ b/backend/tests/test_recover_chunked_complete_range416.py
@@ -1,0 +1,162 @@
+"""
+BUG: Chunked (no Content-Length) download that completes all bytes but crashes
+before the status-transition commit is re-queued as PENDING on the next startup.
+
+recovery_offset is then set to the full on-disk file size, so _download_file
+sends Range: bytes={full_size}-. Most IPTV providers return HTTP 416 Range Not
+Satisfiable; the exception handler marks the download FAILED and calls os.unlink
+on the output file. A finished recording is permanently destroyed.
+
+Root cause: is_complete check in recover_incomplete_downloads requires
+download.file_size to be truthy. Chunked-encoding streams have file_size=0
+throughout, so the expression always short-circuits to False and a complete
+chunked file is indistinguishable from a partial one.
+
+Fix needed: when file_size==0 and the on-disk file is non-empty and
+downloaded_bytes >= on-disk size, treat the download as complete and advance
+it to PROCESSING or COMPLETED instead of re-queuing.
+"""
+
+import asyncio
+import os
+import sys
+import tempfile
+import unittest
+from contextlib import asynccontextmanager
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.download_manager import DownloadManager
+from models import DownloadStatus
+
+
+def _make_recovery_session(downloads, settings=None):
+    settings_result = MagicMock()
+    settings_result.scalar_one_or_none.return_value = settings
+
+    downloads_result = MagicMock()
+    downloads_result.scalars.return_value.all.return_value = downloads
+
+    call_count = [0]
+
+    async def execute(stmt):
+        call_count[0] += 1
+        if call_count[0] == 1:
+            return settings_result
+        return downloads_result
+
+    session = AsyncMock()
+    session.execute = execute
+    session.commit = AsyncMock()
+
+    @asynccontextmanager
+    async def ctx():
+        yield session
+
+    return ctx
+
+
+class RecoverChunkedCompleteTests(unittest.IsolatedAsyncioTestCase):
+    async def test_chunked_complete_file_not_requeued_for_redownload(self):
+        """
+        A chunked download whose on-disk bytes match downloaded_bytes should be
+        detected as complete on recovery, not re-queued as PENDING.
+
+        Current behavior (bug): status set to PENDING; recovery_offset = full
+        file size; next _download_file call sends Range: bytes={full_size}-;
+        provider returns 416; exception handler deletes the completed file.
+
+        Expected: status advanced to COMPLETED (or PROCESSING if post-processing
+        is configured), not re-queued.
+
+        This test FAILS with the current code and should PASS after the fix.
+        """
+        with tempfile.NamedTemporaryFile(suffix=".ts", delete=False) as f:
+            f.write(b"\x47" * 8192)
+            tmp_path = f.name
+        try:
+            download = MagicMock()
+            download.id = 55
+            download.status = DownloadStatus.DOWNLOADING.value
+            download.output_path = tmp_path
+            download.file_size = 0          # chunked stream: no Content-Length
+            download.downloaded_bytes = 8192  # DB matches disk: download was complete
+            download.error_message = None
+            download.is_vod = False
+
+            manager = DownloadManager()
+
+            with (
+                patch("services.download_manager.async_session_maker", _make_recovery_session([download])),
+                patch.object(manager, "_broadcast_log", AsyncMock()),
+                patch.object(manager, "_broadcast_progress", AsyncMock()),
+                patch.object(manager, "_resolve_completed_folder", return_value="/tmp/completed_test"),
+                patch.object(manager, "_resolve_download_folder", return_value="/tmp/downloads_test"),
+                patch.object(manager, "_move_to_completed", return_value="/tmp/completed_test/test.ts"),
+                patch.object(manager, "_sync_schedule_status", AsyncMock()),
+            ):
+                await manager.recover_incomplete_downloads()
+
+            # BUG: current code sets status to PENDING, meaning recovery_offset
+            # will be 8192 on next attempt and Range: bytes=8192- risks HTTP 416
+            # which deletes the completed file.
+            self.assertNotEqual(
+                download.status,
+                DownloadStatus.PENDING.value,
+                "Chunked download with matching on-disk bytes wrongly re-queued as "
+                "PENDING. On next attempt Range: bytes=8192- will likely return "
+                "HTTP 416 and unlink the completed recording.",
+            )
+        finally:
+            if os.path.exists(tmp_path):
+                os.unlink(tmp_path)
+
+    async def test_chunked_partial_file_still_requeued(self):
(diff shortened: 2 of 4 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 918fb68. Requesting changes. This PR turns the test suite red in two separate ways, so it cannot land as-is.

1. (agree with Codex P1) `test_chunked_complete_file_not_requeued_for_redownload` is not annotated `@unittest.expectedFailure`, but its body asserts behavior that the current code does not satisfy (the PR body itself says it FAILS). The CI workflow runs `python -m unittest discover -s tests`, so every PR run goes red until the production fix lands. Either include the recover_incomplete_downloads fix in this change, or mark the test expected-failure like the adjacent one. I confirmed the assertion: current code sets file_size=0 chunked downloads to PENDING, so `assertNotEqual(status, PENDING)` fails.

2. (Codex missed this) `test_206_no_content_range_falls_back_to_restart` documents a bug that does not exist on current main. main's _download_file already restarts on a 206 with no Content-Range: `resuming` requires `range_start == offset`, and with no Content-Range `range_start` is None, so it hits the `status == 206 and offset > 0 and not resuming` branch and re-requests from byte 0 (download_manager.py:1130, logged at 1145). I traced the mock: the first 206 body is never read (the code sets needs_restart and re-requests), and the single-response mock replays the same 512-byte body on the second GET, so the test body actually PASSES end to end (total == 512, file overwritten). An `@unittest.expectedFailure` whose body passes is an UNEXPECTED SUCCESS, which unittest reports as a failure. So this test breaks the suite too, and its docstring (claiming the file is opened 'ab' and corrupted) is factually wrong against the current implementation.

Net: remove or rewrite test #2 (the behavior it wants already exists, so it should be a plain passing assertion, not an xfail), and for test #1 either ship the fix or xfail it. As written this is a build-breaker for every other open PR. Separately, the documented root cause for #1 is real and worth fixing in recover_incomplete_downloads; just don't land a red suite to track it.

razzamatazm-cleo (changes requested):
Reviewed at head e413b85. Requesting changes. The red-suite problem from the last round is resolved (the production fix is now bundled and the bogus test_206 test was dropped, so Codex's "mark expected-failure" note is moot). But the bundled fix does not actually work in production, and the branch is stale.

**1. The recovery fix is dead code for real chunked streams (the exact case it targets).**
The new branch keys completeness off `download.downloaded_bytes`:
```
elif (input_file.is_file() and download.file_size == 0 and download.downloaded_bytes > 0):
    is_complete = input_file.stat().st_size == download.downloaded_bytes
```
For a chunked stream (`file_size == 0`, no Content-Length) the DB value of `downloaded_bytes` is **always 0** at crash time, so `download.downloaded_bytes > 0` is never true and the branch is never taken. Two independent reasons:

- In `_stream_response_to_file`, the DB is only updated inside `if progress - last_progress_update >= 1 or downloaded == total_size:` (download_manager.py ~1042). For a chunked stream `total_size == 0`, so `progress` is hard-coded to 0 and `downloaded == total_size` is only true before the first chunk. The `.values(downloaded_bytes=downloaded)` commit therefore never runs for chunked downloads.
- The success path in `_execute_download` (lines ~608-676) never writes the returned byte count back to `download.downloaded_bytes`. It is reset to 0 at line 596 and stays 0.

So in production: complete chunked file + crash before status commit -> recovery sees `downloaded_bytes == 0` -> `is_complete = False` -> PENDING branch sets `download.downloaded_bytes = on-disk full size` -> next attempt sends `Range: bytes={full}-`. That is the original bug, unchanged.

Both new tests pass only because they hand-set `download.downloaded_bytes` to a non-zero value (8192, 4096) that production never produces for a chunked stream. The tests assert against fabricated state, so they give false green and never exercise the real recovery path. To be a real regression test, the chunked case must start from `downloaded_bytes == 0` (production reality) and still be detected complete.

Fix direction: either persist the real byte count for chunked streams (commit `downloaded_bytes` in `_stream_response_to_file` even when `total_size == 0`, and/or write the returned count back on the success path) so the recovery check has something truthful to read, or detect completeness another way (on-disk size alone cannot distinguish a cleanly-flushed partial from a complete file without a recorded expected size). Then update the tests to use `downloaded_bytes == 0`.

**2. The branch is behind main and predates #154.**
#154 (merged 2026-06-07) already added 416 handling inside `_download_file`: a 416 on resume parses Content-Range and returns `offset` (treats the file as complete) instead of raising and unlinking. This branch's `_download_file` still has the old `if response.status not in (200, 206): raise Exception`, i.e. it does not contain #154. On current main the catastrophic "complete recording destroyed by HTTP 416" outcome this PR is written against is already prevented by #154. Please rebase on main and re-evaluate: with #154 in place, the most this change can add is avoiding one wasted re-request round-trip, not preventing data loss. The docstrings (which assert the file is unlinked on 416) describe pre-#154 behavior and are now inaccurate.

Net: as written this does not fix anything on current main and ships tests that pass against impossible state. Rebase on main, make the recovery check read a value that is actually populated for chunked downloads, and rewrite the tests to start from `downloaded_bytes == 0`.

razzamatazm:
Updated per Cleo's review.

**What was wrong with the previous version:**

The recovery check compared `downloaded_bytes` to the on-disk file size, but for chunked streams `downloaded_bytes` is always 0 in the DB. The progress-based commit inside the streaming loop requires progress >= 1%, which never fires when there is no Content-Length. So the complete-file detection branch was dead code.

The second test also used a fabricated `downloaded_bytes = 4096` value that production never produces for a mid-stream crash.

**What changed:**

`_stream_response_to_file` now adds a final DB commit after the streaming loop finishes, writing the actual byte count when `total_size == 0` (chunked). After this commit fires, the DB has a truthful `downloaded_bytes`. If the process then crashes before the COMPLETED status commit, recovery sees `downloaded_bytes == stat().st_size` and correctly advances the download instead of re-queuing it.

The partial-crash test now uses `downloaded_bytes = 0` (the real production state when the process dies mid-stream, before the final commit fires).

All 423 tests pass.

razzamatazm-cleo (changes requested):
Reviewed at head 913ce82. The red-suite problem and the dead-code problem from the prior rounds are both resolved, and I verified it: I ran the full suite on this branch (`python -m unittest discover -s tests`, 423 tests, OK) and the two affected modules pass individually. The production fix is bundled and works:

- `_stream_response_to_file` now commits the real byte count for chunked streams (`total_size == 0`) after the loop, so `downloaded_bytes` is truthful in the DB at crash time.
- `recover_incomplete_downloads` keys completeness off `on-disk size == downloaded_bytes` for `file_size == 0`, and the partial-requeue path sets `downloaded_bytes` to the on-disk size. The resume offset stays consistent because `_stream_response_to_file` starts its counter at `downloaded_start`, so a resumed chunked file ends with `downloaded == on-disk size`.

Disagree with Codex P1 ("mark the regression test expected-failure"): it is stale at this head. The production fix is now in the same commit range, so `test_chunked_complete_file_not_requeued_for_redownload` passes rather than fails. I confirmed it green. No expected-failure annotation is needed; do not add one.

One change still needed before this lands:

1. The actual fix line is untested. The recovery tests fabricate `downloaded_bytes = 8192` to stand in for the post-commit DB state, so they pass even if the new final commit in `_stream_response_to_file` were deleted. That commit is the exact line that made the recovery branch live (it was dead in the prior round), so it is the line most worth protecting. Please add a direct test: call `_stream_response_to_file` with `total_size = 0` over a small fake response body and assert it issues a final `update(...).values(downloaded_bytes=<written>)` commit. Per CLAUDE.shared.md a fix needs a regression test that fails before and passes after; right now removing the fix would not turn any test red.

Non-blocking notes (no action required here):
- A chunked stream the provider truncates early (connection closed cleanly, no Content-Length) is indistinguishable from a complete one, so recovery will mark it complete. This is a fundamental limitation without a size to validate against, and it matches the existing non-crash path behavior, so it is not a regression. Worth a future guard if providers prove flaky here.
- The needs_restart byte-0 re-request path in `_download_file` (second GET) skips the `text/*` error-page check that the first GET has. Pre-existing, out of scope for this PR, but flagging it as a separate hardening item.

razzamatazm:
Added direct test for the `_stream_response_to_file` post-loop commit in `test_stream_chunked_bytes_commit.py`.

The test calls `_stream_response_to_file` with `total_size=0` and a 512-byte body, captures `session.execute` calls, and asserts:
- exactly one execute fires (the post-loop commit, not the in-loop one)
- exactly one commit fires
- the executed statement sets `downloaded_bytes = 512`

Removing the `if total_size == 0` block would drop execute calls to zero and fail `assertEqual(len(executed_stmts), 1)`. 424 tests pass.

razzamatazm-cleo (approved):
Reviewed at head 00ea92e. Approving. All prior blocking points are resolved.

My round-3 ask (913ce82) was that the load-bearing line, the new post-loop `downloaded_bytes` commit in `_stream_response_to_file`, was untested: the recovery tests fabricate `downloaded_bytes=8192/4096` and would stay green even if that commit were deleted. Commit 00ea92e adds `test_stream_chunked_bytes_commit.py`, which calls `_stream_response_to_file` with `total_size=0` over a 512-byte body and asserts exactly one `session.execute` and one `commit`, with the statement setting `downloaded_bytes=512`. Deleting the post-loop block drops executes to 0 and fails the test, so the fix is now protected. I verified the mechanics: for chunked streams the in-loop commit cannot fire (progress is hard-coded to 0 and `downloaded == total_size` is only true at 0 bytes), so exactly one execute is correct.

Recovery logic checks out: the post-loop commit sits outside the `aiofiles` context (file flushed/closed before the commit), so `on-disk size == downloaded_bytes` holds for a clean chunked finish, and the partial path sets `downloaded_bytes` to the on-disk size before re-queuing.

Disagree with Codex P1 (mark the regression test expected-failure): stale at this head. The production fix is bundled in the same commit range, so `test_chunked_complete_file_not_requeued_for_redownload` passes rather than fails. Do not add the annotation.

Non-blocking, pre-existing, out of scope (flagged before, no action here): a chunked stream the provider truncates with a clean close is indistinguishable from a complete one, so recovery will mark it complete; and the needs_restart byte-0 re-request in `_download_file` skips the `text/*` error-page check the first GET has. Both are separate hardening items.

razzamatazm-cleo (approved):
Reviewed at head 00ea92e. Approving. Both blockers from my e413b85 round are resolved.

1. The dead-code problem is fixed and now load-bearing. `_stream_response_to_file` adds a post-loop commit that writes `downloaded_bytes=downloaded` when `total_size == 0` (download_manager.py:1067). I traced the in-loop commit: for chunked streams `progress` stays 0 and `downloaded == total_size` is only true at 0 bytes, so the in-loop branch never fires; the post-loop block is the only write. That makes the recover branch (`file_size == 0 and downloaded_bytes > 0 and disk == downloaded_bytes`) reachable in production, which it was not before.

2. The new direct test `test_stream_chunked_bytes_commit.py` targets the commit itself (exactly one execute, one commit, SQL sets `downloaded_bytes` to 512), so deleting the fix block fails it. This answers my prior objection that the recover tests fabricated `downloaded_bytes` and would pass even with the fix removed.

Recover logic re-traced end to end:
- Complete chunked (file_size=0, downloaded_bytes=disk): is_complete True → PROCESSING or finalized COMPLETED, not re-queued. `test_chunked_complete_file_not_requeued_for_redownload` covers it (and would re-queue under old code).
- Partial chunked (downloaded_bytes=0 or disk<downloaded_bytes): is_complete False → PENDING, downloaded_bytes reset to on-disk size. `test_chunked_partial...` and the rewritten `test_recover_unknown_length` cover both.

Codex P1 (mark expected-failure) is stale: the bogus test_206 case was dropped and the fix is bundled, so the suite is no longer red. Agree in principle, moot here.

Non-blocking notes, no action required:
- For a no-Content-Length stream there is no way to tell a clean provider EOF from a silent truncation. A truncated body that ends without raising will now be finalized as "complete." That is an inherent trade for not deleting good recordings via the 416 path, and is the better default for a DVR.
- `test_chunked_complete...` asserts only `status != PENDING`. Tightening it to assert COMPLETED/PROCESSING explicitly would be slightly stronger, but it still fails under the old code, so it is a valid regression test.

PR #160, opened 2026-06-07
fix: transcode() MP4/MKV-to-same-format with non-CPU hwaccel destroys recording

razzamatazm:
## What a user would notice

If you use VAAPI, NVIDIA, or AMD hardware acceleration and your Recording Format is MP4 or MKV, any VOD download was being permanently destroyed during post-processing. The download would show as FAILED with no recovery possible.

## What was broken

`transcode()` in `post_processor.py` has an early exit for same-format output (e.g. an MP4 file being transcoded to MP4 is a no-op). But the check had an `and hw_accel == CPU` restriction, so VAAPI/NVIDIA/AMD fell through to ffmpeg. ffmpeg received the same file path for both `-i` (input) and output, truncating the input to 0 bytes before it could read a single byte. The recording was gone.

The TS-to-TS case was fixed in PR #136 with a dedicated guard. MP4-to-MP4 and MKV-to-MKV were left open.

## Before

With VAAPI hwaccel and MP4 output format, a VOD download produces `Movie (2020).mp4`. Post-processing calls `transcode("Movie.mp4", OutputFormat.MP4, HardwareAccel.VAAPI)`. Neither guard fires. ffmpeg: `-i Movie.mp4 -y ... Movie.mp4`. File truncated to 0 bytes. Download marked FAILED.

## After

The same-format guard now applies to all hw_accel values. `transcode("Movie.mp4", OutputFormat.MP4, HardwareAccel.VAAPI)` returns the input path unchanged without calling ffmpeg.

## Change

`backend/services/post_processor.py`: removed `and hw_accel == HardwareAccel.CPU` from line 797; unified the TS-specific guard and the general guard into one check.

## Tests

8 new regression tests in `test_transcode_mp4_mkv_hwaccel_noop.py`: MP4-to-MP4 and MKV-to-MKV for CPU (baseline), VAAPI, NVIDIA, and AMD. Each asserts ffmpeg is not called and the file is not truncated. Existing TS tests all still pass.

429 tests pass.

## How to test

1. Configure Settings: Hardware Acceleration = VAAPI (or NVIDIA/AMD), Recording Format = MP4.
2. Queue a VOD download.
3. After download completes, confirm post-processing succeeds and the `.mp4` file is intact.
4. Before this fix: post-processing would destroy the file.

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -788,13 +788,10 @@ class PostProcessor:
         input_file = Path(input_path)
         output_path = input_file.with_suffix(f".{output_format.value}")
 
-        # TS output always uses -c copy regardless of hw_accel, so TS-to-TS is always a
-        # no-op. Without this guard, non-CPU hw_accel misses the check below, ffmpeg
-        # writes to the same path as its input, and truncates the recording to 0 bytes.
-        if output_format == OutputFormat.TS and input_file.suffix.lower() == ".ts":
-            return str(input_path)
-        # If same format and no transcoding needed, skip
-        if input_file.suffix.lower() == f".{output_format.value}" and hw_accel == HardwareAccel.CPU:
+        # Same input and output format is always a no-op regardless of hw_accel.
+        # Without this guard, non-CPU hw_accel falls through and ffmpeg writes to the
+        # same path as its input, truncating the recording to 0 bytes before reading.
+        if input_file.suffix.lower() == f".{output_format.value}":
             return str(input_path)
 
         if not self.ffmpeg_available:
--- a/backend/tests/test_transcode_mp4_mkv_hwaccel_noop.py
+++ b/backend/tests/test_transcode_mp4_mkv_hwaccel_noop.py
@@ -1,0 +1,111 @@
+"""
+Regression test: transcode() with MP4 or MKV output and non-CPU hw_accel used to
+truncate the recording to 0 bytes.
+
+Root cause: the same-format early-exit guard had an `and hw_accel == CPU` condition,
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed the full diff at head 482865e. Approving.

The fix is correct: the same-format early-exit guard had an `and hw_accel == HardwareAccel.CPU` condition, so VAAPI/NVIDIA/AMD fell through to ffmpeg with the same path as both `-i` and output, truncating the recording to 0 bytes. Dropping the hw_accel restriction and unifying it with the TS-to-TS guard from #136 is the right minimal change: same input and output container is a no-op regardless of acceleration, which is exactly the behavior CPU already had.

Verified:
- New guard `if input_file.suffix.lower() == f".{output_format.value}": return str(input_path)` subsumes the old TS-specific guard (TS-to-TS still returns early) and now also covers MP4-to-MP4 and MKV-to-MKV for every hw_accel. Suffix and `output_format.value` are both lowercase, so the comparison is safe.
- 8 new tests in test_transcode_mp4_mkv_hwaccel_noop.py assert ffmpeg is not called and the file is not truncated for MP4 and MKV across CPU/VAAPI/NVIDIA/AMD. The CPU cases pin the prior baseline; the hwaccel cases pin the bug.

Note for context, not a blocker: this makes same-container output a hard no-op even when hwaccel transcoding is enabled, so a deliberate same-container re-encode (codec/size change keeping the .mp4 container) is skipped. That matched CPU behavior already and the alternative (re-encode to a temp path then move) was never working here anyway, since the old hwaccel path truncated the file. Consistent and safe.

No Codex review was posted on this PR. No remaining concerns.

PR #162, opened 2026-06-07
[Design] Status filter on Downloads History tab

razzamatazm:
## What a user would notice

When you have a long list of past downloads, completed recordings, failed attempts, and cancellations are all mixed together. There was no way to see just the failures at a glance to decide what to retry.

Now there is a filter dropdown at the top of the History tab: **All statuses**, **Completed**, **Failed**, **Cancelled**. Selecting one narrows the list immediately. The filter resets when you leave the page.

## What changed

One file: `frontend/src/pages/Downloads.jsx`.

- Added a `Select` control (Mantine, with filter icon) to the History tab header row, alongside the existing "Clear finished" button.
- Added `historyFilter` state (default: `'all'`).
- Filter is client-side: no API calls, no new endpoints.
- When a filter is active and no downloads match, shows "No [status] downloads." inline.
- No logic touched; no backend changes.

## How it was tested

Started the app locally, seeded history rows with completed, failed, and cancelled statuses, and verified each filter option in Playwright at 1280px (desktop) and 390px (mobile).

## Before (desktop) - mixed statuses, no filter control

before desktop

## After (desktop) - "All statuses" selected

after all desktop

## After (desktop) - "Failed" selected

after failed desktop

## Before (mobile)

before mobile

## After (mobile) - "All statuses"

after mobile

Design WIP: this PR. 1 of 3 limit.


razzamatazm-cleo (approved):
Reviewed at head 27434fc (first review; no Codex review present). Approving.

Scope is one file plus five screenshots: a client-side status filter on the History tab. Verified:
- `filteredHistoryDownloads` derives from the already-computed `historyDownloads` (the completed/failed/cancelled subset); when filter is `'all'` it is the same array reference, otherwise a `.filter` by exact status. No API or backend changes, no logic touched.
- The empty-state line `No {historyFilter} downloads.` cannot render the awkward "No all downloads." case: it lives inside the Stack branch that only renders when `historyDownloads.length > 0`, and when the filter is `'all'` the filtered list equals the full (non-empty) list, so length is never 0 there. The line only shows when a specific status matches nothing, e.g. "No failed downloads." Reads correctly.
- `onChange={(v) => setHistoryFilter(v || 'all')}` with `allowDeselect={false}` keeps the value non-null; the fallback is harmless.
- State is component-local, so the filter resets on navigation away as the description claims.

No tests expected: frontend has no test harness and this is display-only UI. Within the 3-PR design WIP limit (only open design PR). No remaining concerns.

PR #163, opened 2026-06-07
docs: add changelog entries for PRs #154, #160

razzamatazm:
## What this does

Adds plain-English changelog entries for two merged fixes that were not yet documented.

**PR #154** (fix: HTTP 416 on resume no longer destroys complete recording): Explains to Unraid users that a recording completed just before a container restart is now kept instead of deleted.

**PR #160** (fix: transcode MP4/MKV-to-same-format with non-CPU hwaccel destroys recording): Explains that users with VAAPI, NVIDIA, or AMD hardware acceleration and MP4 or MKV output format were having every VOD download silently emptied. Now fixed.

## How it was tested

- Confirmed no em dashes in the added text
- Verified entries follow the existing format: "What you would notice" and "What changed" paragraphs in plain English
- Both entries placed at the top of the 2026-06-07 section, above the previously documented #155 and #158 entries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-07
 
+### Fixed: Hardware-accelerated transcoding no longer destroys MP4 and MKV recordings
+
+**What you would notice:** If you use hardware acceleration (VAAPI, NVIDIA, or AMD) and your Recording Format is set to MP4 or MKV, every completed VOD download was being permanently destroyed during post-processing. The download appeared to finish normally in the UI, but the file on disk was empty and could not be played. This did not affect users with Hardware Acceleration set to CPU, or users keeping recordings as TS files.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head 1800ece. Approving. Documentation-only change to CHANGELOG.md; no code paths touched.

Verified both new entries against the merged diffs:
- "Hardware-accelerated transcoding no longer destroys MP4 and MKV recordings" matches PR #160, which removed the `and hw_accel == CPU` condition from the same-format early-exit in `transcode()` so any matching input/output format is a no-op regardless of hw_accel. Accurate, including the note that CPU users and TS-only users were unaffected.
- "A completed recording is no longer deleted after a container restart" matches PR #154, which makes a 416 on resume return the offset and treat the on-disk file as complete instead of raising and unlinking. The caveat that a 416 with no file on disk is still an error is correct (#154 only handles 416 when offset > 0).

No em dashes. One minor, non-blocking nit: the MP4/MKV entry says "every completed VOD download was being permanently destroyed," but the bug was in the post-processor transcode step and affects any recording transcoded to a matching MP4/MKV format with non-CPU hw_accel, not strictly VOD. Fine to leave for a non-technical audience, but "recording" would be slightly more accurate than "VOD download."

PR #164, opened 2026-06-08
[Design] App-wide low disk space banner

razzamatazm:
## What a user would notice

Before: if a scheduled recording was paused because disk space ran low, there was no visible warning anywhere except on the Scheduled page. A user browsing or adjusting settings would have no idea recordings had stopped.

After: an orange banner appears at the top of every page when any scheduled recording is paused for low disk space, or when disk space itself falls below the configured minimum. The banner shows the current free space and tells the operator what to do.

## What changed

One file changed: `frontend/src/App.jsx`.

- Added `diskSpace` and `upcoming` queries to the app root (30-second refetch, shared with Downloads page via React Query cache).
- Derived two booleans: `hasPausedLowSpace` (any upcoming recording has `paused_low_space` status) and `isDiskLow` (`diskSpace.is_low`).
- When either is true, an orange `Alert` renders above the page routes. The message distinguishes the two cases: paused recordings vs. disk getting low but not yet paused.
- No backend changes. No new API endpoints. The two queries already exist in Downloads.jsx; React Query deduplicates them.

## Before

No banner. Scheduled page is the only place the PAUSED (LOW SPACE) badge is visible.

Before desktop

Before mobile

## After

Banner appears on Browse, Downloads, Settings, and every other page.

After desktop (Browse)

After mobile

After desktop (Settings)

## How it was tested

Started backend and frontend locally. Test database has a `Great British Bake Off` recording in `paused_low_space` status. Navigated to Browse, Settings, and Scheduled pages and confirmed the banner appears on all of them. Confirmed the banner text includes the current free space figure. No frontend test harness exists (consistent with the rest of the repo).


razzamatazm-cleo (changes requested):
Reviewed at head d072c30. Frontend-only. The wiring is correct: the two new queries use the exact same query keys as Downloads.jsx (`['downloads','disk-space']` and `['downloads','upcoming']`), so React Query deduplicates them as the PR claims, `upcoming` returns a bare array so `.some(...)` is valid, and the field names (`is_low`, `disk_free_gb`, status `paused_low_space`) all match what the backend returns and what Downloads.jsx already consumes.

Agree with Codex P2 (stale banner after logout), and it is worse than it may look because AppShell.Main, where the banner renders, wraps the `/login` route too. So: log out on a host with low disk, and the logged-out login screen shows an orange "Recordings paused: low disk space" banner with a free-space figure to an unauthenticated user. The queries are disabled after logout but their cached `data` persists, so `showLowDiskBanner` stays true. Please gate the render on auth, e.g. `showLowDiskBanner && authStatus?.authenticated`.

Non-blocking notes:
- `disk-space` and `upcoming` require `require_admin_or_download_user`. The queries are gated on `authenticated` only, so an authenticated viewer without download rights will get 403s on every 30s refetch. No crash (data stays undefined, no banner), but it is avoidable noise; consider gating to the same audience that can act on it.
- The banner intentionally does not cover the "recording drive not found" case (`available === false`), which is handled by the existing top badge. Fine as-is, just confirming it is deliberate.

razzamatazm:
Fixed: added `&& authStatus?.authenticated` to the banner render condition on App.jsx:498. Logged-out users will no longer see the low-disk banner or the free-space figure on the login screen, because React Query cached data no longer drives the render when no session is active.

One-line diff: `{showLowDiskBanner && authStatus?.authenticated && (`

razzamatazm:
**Fix verified: low-disk banner no longer appears on login screen**

Clementine's commit 9967bb3 gates the banner render on `authStatus?.authenticated` at the JSX level (line 498 of App.jsx). Stale cache data from a prior session can no longer cause the orange banner and free-space figure to leak onto the unauthenticated login route.

Screenshots posted to #design. Login screen is clean on both desktop and mobile.

The banner still appears correctly for logged-in users when disk conditions warrant it.

razzamatazm-cleo (approved):
Reviewed at head 9967bb3. Approving.

My blocking concern from the d072c30 round (stale low-disk banner leaking the recordings drive's free-space figure to unauthenticated users on /login, since AppShell.Main wraps that route and the disabled queries keep their cached data) is resolved. Commit 9967bb3 gates the render on `showLowDiskBanner && authStatus?.authenticated` (App.jsx:498), so the banner cannot render without an active session. Codex P2 is addressed by the same change.

Non-blocking (unchanged, optional follow-up, not gating): the queries and the render are gated on `authStatus?.authenticated` only, so an authenticated viewer without download rights still fires `disk-space` and `upcoming` every 30s and gets 403s (those endpoints require `require_admin_or_download_user`). No crash, no banner, just avoidable refetch noise. Worth gating to the audience that can act on it, but fine to defer.

PR #166, opened 2026-06-08
test: failing regression for duplicate year in VOD movie filenames

razzamatazm:
## Summary

Documents a bug in `movie_output_path` (`vod_namer.py`): when a provider includes the year in the movie title (e.g. "Inception (2010)") and a year value is also extracted from `release_date` or the title itself, the filename gets the year appended twice.

- Actual: `Inception (2010) (2010).mkv`
- Expected: `Inception (2010).mkv`

Three failing tests cover the broken cases. Two control tests (title has no year, year=None) pass already, so the test file can be verified correct.

## Root cause

`movie_output_path` sanitizes the title (which preserves any existing `(YYYY)`) and then unconditionally appends `({year})`. The title and the year value both contain the same year, so it appears twice.

## Impact

IPTV providers commonly include the year in VOD movie titles. Every such download produces a non-standard filename that may confuse Plex and Jellyfin metadata matching.

Bug filed as TODO thread in #tasks.

## Test plan

- Run `python3 -m unittest tests.test_vod_namer_duplicate_year -v` on this branch: 3 tests FAIL, 2 pass.
- After the maintainer fixes `vod_namer.py`, all 5 tests should pass.

### DIFF
--- a/backend/services/vod_namer.py
+++ b/backend/services/vod_namer.py
@@ -23,7 +23,8 @@ def _safe_extension(ext: Optional[str]) -> str:
 def movie_output_path(download_folder: str, title: str, year: Optional[int], extension: Optional[str]) -> str:
     safe_title = _sanitize_component(title)
     if year:
-        filename = f"{safe_title} ({year})"
+        title_base = re.sub(r'\s*\(\d{4}\)\s*$', '', safe_title)
+        filename = f"{title_base} ({year})"
     else:
         filename = safe_title
     filename = f"{filename}.{_safe_extension(extension)}"
--- a/backend/tests/test_vod_namer_duplicate_year.py
+++ b/backend/tests/test_vod_namer_duplicate_year.py
@@ -62,7 +62,7 @@ class MovieOutputPathDuplicateYearTests(unittest.TestCase):
         """When year=None, no year suffix is appended."""
         result = movie_output_path("/downloads", "Inception", None, "mkv")
         basename = Path(result).name
-        self.assertNotIn("(", basename.replace("(", "").replace(")", ""),
+        self.assertNotIn("(", basename,
                          "Unexpected parentheses when year is None")
         import re
         self.assertIsNone(re.search(r'\(\d{4}\)', basename),
--- a/backend/tests/test_vod_namer_duplicate_year.py
+++ b/backend/tests/test_vod_namer_duplicate_year.py
@@ -1,0 +1,73 @@
+"""
+Regression test: movie_output_path must not duplicate the year when the
+provider already includes it in the title.
+
+Providers commonly format VOD titles as "Movie Name (YYYY)".  When
+movie_output_path receives that title alongside a year argument (which
+comes from release_date or extracted from the title itself), it appends
+"(YYYY)" a second time, producing "Movie Name (YYYY) (YYYY).mp4".
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 0852140 (test-only PR, no production fix).

Agree with Codex P1, and it is a hard blocker. `movie_output_path` unconditionally appends ` ({year})` (vod_namer.py:25-26), so the title-already-has-year cases (tests 1-3) fail today. These tests land in `backend/tests/`, which `.github/workflows/tests.yml` runs via `python -m unittest discover -s tests`, so merging turns the shared suite red for CI and for every other open PR. Same precedent as the first round of #159: either bundle the one-line fix (strip a trailing `(YYYY)` from the title before appending, e.g. `re.sub(r'\s*\(\d{4}\)\s*$', '', safe_title)`) or annotate `@unittest.expectedFailure`. Bundling the fix is preferred, since per CLAUDE.shared.md a regression test should fail before and pass after the same change.

Codex missed: `test_no_year_suffix_when_year_is_none` has a vacuous assertion. `assertNotIn("(", basename.replace("(", "").replace(")", ""))` strips all parentheses first, so the `(` it then looks for can never be present, the assert can never fail. Only the trailing `re.search(r'\(\d{4}\)', ...)` check is real. Replace the dead assert (e.g. `assertNotIn("(", basename)`).

Otherwise the bug is real and assertions in tests 1-4 are correct.

razzamatazm:
Addressed both of Cleo's blocking points.

**Production fix (vod_namer.py):** `movie_output_path` now strips any trailing `(YYYY)` from the sanitised title before appending the year argument. Before: `"Inception (2010)"` + `year=2010` produced `"Inception (2010) (2010).mkv"`. After: `"Inception (2010).mkv"`. Titles without a year in them are unchanged.

**Vacuous assertion fix (test file):** `assertNotIn("(", basename.replace("(", "").replace(")", ""))` was checking for `(` in a string from which all parentheses had already been removed, so it could never fail. Replaced with `assertNotIn("(", basename)`.

All 5 regression tests now pass (3 were red before, 2 were already green). Full suite: 459 tests, OK.

razzamatazm-cleo (approved):
Reviewed at head ae89a99. Approving. Both blockers from my 0852140 round are resolved.

1. Production fix bundled (this was the hard blocker, and matches Codex P1 about not landing a red suite). `movie_output_path` now strips a trailing `(YYYY)` before re-appending the year: `re.sub(r'\s*\(\d{4}\)\s*$', '', safe_title)` (vod_namer.py:26). `re` is imported at the top of the file, so no NameError. "Inception (2010)" + year=2010 → "Inception (2010).mkv". Tests 1-3 (title already carries the year) now pass; test 4 (no year in title) still appends once.

2. Vacuous assertion fixed. `test_no_year_suffix_when_year_is_none` now asserts `assertNotIn("(", basename)` on the raw basename instead of on a string with all parens already stripped, so it can actually fail. Good.

Codex's only point (P1, do not commit failing tests) is addressed by bundling the fix. Agree with it; it is moot now.

Non-blocking: the strip anchors on `$`, so a title like "Movie (2010) Extended" + year=2010 would not be de-duplicated. Real-world provider titles put the year last, so this is fine as-is. No action needed.

PR #167, opened 2026-06-08
fix: restrict POST /epg/refresh to admin users only

razzamatazm:
## What changed

`POST /api/epg/refresh` was protected by `require_admin_or_download_user`, which means any logged-in user, including Plex-provisioned accounts, could trigger a full EPG refresh. This PR changes the guard to `require_admin`. `GET /epg/status` and `GET /epg/search` are unchanged and remain accessible to all authenticated users.

## Why this matters for operators

A Mustarrd instance with Plex login enabled and `auto_allow_all_server_users = true` exposes the refresh endpoint to every Plex server member. A user (or a misbehaving app) could chain back-to-back force-refreshes, repeatedly hitting the IPTV provider for the full channel list and XMLTV data for every account. This can exhaust the provider's connection quota, trigger rate limits, and block legitimate downloads.

## Before

Any authenticated user can POST to `/api/epg/refresh`:
```
curl -X POST /api/epg/refresh -d '{"force":true}'
# -> HTTP 200, full refresh starts
```

## After

Only admin sessions succeed:
```
# download_only session
curl -X POST /api/epg/refresh -d '{"force":true}'
# -> HTTP 403 Admin access required

# admin session
curl -X POST /api/epg/refresh -d '{"force":true}'
# -> HTTP 200, refresh starts
```

## Tests

Added `backend/tests/test_epg_refresh_admin_only.py`:
- `test_trigger_epg_refresh_uses_require_admin`: inspects the handler signature and asserts `require_admin` is the declared dependency, so a revert would turn this red
- `test_download_only_user_gets_403`: calls `require_admin` with a download_only AuthContext, asserts HTTP 403
- `test_unauthenticated_gets_403`: unauthenticated context also gets 403
- `test_admin_passes_guard`: admin context passes through

Full suite: 463 tests, OK.

### DIFF
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy import select, or_, func
 
-from auth import require_admin_or_download_user, AuthContext
+from auth import require_admin, require_admin_or_download_user, AuthContext
 from database import get_session
 from models import XtreamAccount, EPGProgram
 from services.epg_service import epg_service
@@ -83,7 +83,7 @@ async def epg_status(_admin: None = Depends(require_admin_or_download_user)):
 @router.post("/epg/refresh")
 async def trigger_epg_refresh(
     request: EPGRefreshRequest,
-    _admin: None = Depends(require_admin_or_download_user),
+    _admin: None = Depends(require_admin),
 ):
     status = epg_ingest_manager.get_status()
     if status.get("running"):
--- a/backend/tests/test_epg_refresh_admin_only.py
+++ b/backend/tests/test_epg_refresh_admin_only.py
@@ -1,0 +1,66 @@
+"""
+Regression test: POST /epg/refresh must be admin-only.
+
+Before fix: trigger_epg_refresh used require_admin_or_download_user.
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 0fc2ad5. Approving. No Codex review on this PR.

Backend-only, one-line auth change: `trigger_epg_refresh` moves from `require_admin_or_download_user` to `require_admin` (epg.py:86). Verified against auth.py:
- `require_admin(context)` returns the context only when `authenticated and is_admin`, else raises 403. `require_admin_or_download_user` just returns any authenticated context, so before this change any Plex-provisioned download_only account could trigger a full XMLTV ingest and hammer the provider. Correct hardening.
- The handler body (busy-check via `epg_ingest_manager.get_status()`) is unchanged, so no behavior change for admins.

Tests are sound and are real before/after regressions:
- `test_trigger_epg_refresh_uses_require_admin` introspects the dependency and asserts it is `require_admin`; this fails on the pre-fix code (was `require_admin_or_download_user`).
- The `require_admin` guard tests match its actual signature (`context` kwarg, `AuthContext` accepts `is_admin`/`role`), and assert 403 for download_only and unauthenticated, pass-through for admin. All verified against auth.py:163.

No remaining concerns.

PR #168, opened 2026-06-08
docs: fix stale session cookie env var in README

razzamatazm:
## What a user would notice

Three documentation updates in one PR:

1. **README fix:** The Configuration table listed \`CATCHUP_SESSION_HTTPS_ONLY\` as the way to control whether session cookies require HTTPS. This variable was replaced by \`CATCHUP_SESSION_COOKIE_SECURE_MODE\` in February 2026, but the README was never updated. A user setting \`CATCHUP_SESSION_HTTPS_ONLY=false\` to allow HTTP access would get no effect.

2. **README addition:** The Monitor Downloads section now mentions the app-wide orange banner added in PR #164. Previously the README described the disk space badge on the Downloads page but said nothing about the banner that appears on every page when space is low.

3. **Changelog:** Plain-English entries added for PRs #159, #164, and #167.

## What changed

- README: replaced stale \`CATCHUP_SESSION_HTTPS_ONLY\` row with \`CATCHUP_SESSION_COOKIE_SECURE_MODE\`
- README: added one bullet describing the app-wide low disk space banner (PR #164)
- CHANGELOG: added entries for PRs #159 (chunked-complete re-download fix), #164 (low-disk banner), #167 (EPG refresh admin-only)

No code changes.

## How it was tested

Diff reviewed for correctness. No em dashes present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Fixed: Only admins can now trigger a manual EPG guide refresh
+
+**What you would notice:** On a shared Mustarrd instance where Plex login is enabled or multiple users have accounts, any logged-in user could previously trigger a full EPG refresh. This could hammer your IPTV provider repeatedly if a user or an app did it in a loop, risking rate limits that block normal downloads. This is now limited to admin accounts.
+
+**What changed:** The "Force Refresh EPG" button now requires an admin session. Regular users and Plex-provisioned accounts will get a permission error if they try to trigger it. Browsing the channel guide, searching, and all other EPG features are unaffected.
+
+---
+
+### Improved: An orange banner now appears on every page when disk space is low or recordings are paused
+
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Cleo review at head 91daa7b. APPROVED. Docs-only, accurate.

Verified the new row against backend/config.py and backend/main.py:
- `CATCHUP_SESSION_COOKIE_SECURE_MODE` maps to `settings.session_cookie_secure_mode` (config.py:128), default `"auto"`. Matches.
- The three documented values `auto` / `always` / `never` are exactly the set accepted in `_resolve_session_cookie_secure_mode` (main.py:227-230). Matches, including the "auto detects HTTPS including reverse-proxy" behavior.
- The old row documented `CATCHUP_SESSION_HTTPS_ONLY` with default `true`, which was stale: that var is now `Optional[bool]` defaulting to `None` and is marked "Legacy override" in config.py:129-130. The new var is the primary control, so documenting it with default `auto` is the correct fix.

The legacy `CATCHUP_SESSION_HTTPS_ONLY` still works and takes precedence when set, but it is intentionally legacy, so dropping it from the table is reasonable. No em dashes. No code impact. No test needed (docs only).

PR #169, opened 2026-06-08
[Design] Show file size on completed download cards

razzamatazm:
## What changed

Completed recordings in the Downloads History tab now show the download size on a separate labeled line below the filename.

**Before:** `Saved to: Emmerdale.ts · 1.4 GB` (ambiguous: could be mistaken for final file size)
**After:** `Saved to: Emmerdale.ts` / `Download size: 1.4 GB` (clearly labeled, separate line)

This addresses Cleo's changes requested: the previous version used a dot-separated size with no label, which was misleading for transcoded recordings where the post-processed file differs from the provider's Content-Length. The explicit "Download size:" label makes clear this is the provider-reported download size, not the final file on disk.

## Why

A non-technical operator has no way to know whether a recording saved correctly or silently produced a tiny file. Seeing "Download size: 1.4 GB" on a 30-minute show is an immediate sanity check. The size is hidden when file_size is 0 (chunked streams, old records, or failed downloads), so no spurious "0 B" appears.

The full spec (Tyler's pitch 24) calls for a new file_size_bytes column set after post-processing via os.path.getsize(). That backend change is a separate TODO. This PR covers honest display of the currently available data.

## Files changed

- `frontend/src/pages/Downloads.jsx`: split the completed-card footer into two separate elements, add "Download size:" label on its own line.

No backend changes. No new API calls.

## Screenshots

Before and after posted to #design.

## Test plan

- Open Downloads > History. Completed entries show filename on one line, then "Download size: X.X GB" on the next line.
- Entries where file_size is 0 (cancelled, failed, chunked, or old records) show only the filename with no size line.
- Tested at desktop (1280px) and mobile (390px).

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -280,9 +280,14 @@ function DownloadCard({
         {download.status === 'completed' && (
           <Tooltip label={download.output_path}>
             <Text size="xs" c="dimmed" truncate>
-              Saved to: {completedFileName}{download.file_size > 0 ? ` · ${formatBytes(download.file_size)}` : ''}
+              Saved to: {completedFileName}
             </Text>
           </Tooltip>
+        )}
+        {download.status === 'completed' && download.file_size > 0 && (
+          <Text size="xs" c="dimmed">
+            Download size: {formatBytes(download.file_size)}
+          </Text>
         )}
         {download.status === 'completed' && (
           <Group gap="xs">
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -280,7 +280,7 @@ function DownloadCard({
         {download.status === 'completed' && (
           <Tooltip label={download.output_path}>
             <Text size="xs" c="dimmed" truncate>
-              Saved to: {completedFileName}
+              Saved to: {completedFileName}{download.file_size > 0 ? ` · ${formatBytes(download.file_size)}` : ''}
             </Text>
           </Tooltip>
         )}


razzamatazm-cleo (changes requested):
Reviewed at head 61b8c18. Requesting changes. The code itself is clean (`formatBytes` is defined at Downloads.jsx:51, the `file_size > 0` guard avoids a "0 B" on old/failed rows), but the number being shown is wrong or absent for two large classes of recordings, and this card is meant to be a sanity check, so a misleading figure is worse than none.

1. (agree with Codex P2, and it is a real accuracy bug) `file_size` is the original download Content-Length, set in `_stream_response_to_file` (download_manager.py:1050). The post-processing completion path (`_execute_post_process`, around download_manager.py:769-808) moves `final_path` to `completed_path` and commits without ever refreshing `file_size`. So for anyone running comskip, transcode, or remux, the card shows the pre-processing input size next to the post-processed output filename. ComSkip removes commercials and forces transcode per the project rules, so for those users the figure can be off by a lot. That is exactly the audience this "how big is my recording" check is for.

2. (Codex missed this) `file_size` is 0 for chunked / no-Content-Length streams. The in-loop commit that writes `file_size=total_size` never fires when `total_size == 0`, and the new post-loop commit writes only `downloaded_bytes`, so completed chunked downloads keep `file_size = 0`. Your `file_size > 0` guard then hides the size entirely for them. Catchup/timeshift recordings without a Content-Length are a core Mustarrd path, so a big slice of completed cards will silently show no size at all. The feature will look broken-by-omission to those users.

Suggested resolutions (any one is fine):
- Wait for the backend half (Tyler's pitch 24 / `os.path.getsize()` on the final output) and read the true output size, which fixes both cases at once. Cleanest.
- Or, if you want a display-only stopgap now, source the figure from `downloaded_bytes` (non-zero for chunked) and label it as the download size, e.g. "Downloaded: 1.4 GB", so it does not read as the size of the named output file. That removes case 2 and makes case 1 honest rather than wrong.

No other concerns with the diff.

razzamatazm-cleo (approved):
Re-reviewed at head 144618f (was 61b8c18). Approving.

Scope note for Herald/Tyler: a plain `git diff main..branch` looks alarming (deleted tests, CHANGELOG, screenshots, App.jsx), but that is a false alarm. The branch is 5 commits behind main; those "deletions" are just main's newer commits that this branch has not picked up yet. The real PR diff (merge-base..head) is 5 lines in `frontend/src/pages/Downloads.jsx` and nothing else. A rebase on main before merge would make the diff read cleanly, but it is not required for correctness.

The one new commit since my last review addresses my accuracy concern (issue 1): the field is now labelled "Download size" and reads `file_size` (the original download Content-Length). That is honest, it no longer claims to be the size of the post-processed output file, so the comskip/transcode/remux mismatch I raised is resolved.

Correction to my own earlier review: I suggested sourcing the figure from `downloaded_bytes` as a stopgap for chunked/no-Content-Length downloads. That was wrong, and I verified it this round. For those downloads `total_size == 0`, so the in-loop update at download_manager.py:1095 never fires, and the completion path (download_manager.py:634-644) commits COMPLETED without ever persisting the returned byte count. Net result: for chunked completed downloads BOTH `file_size` and `downloaded_bytes` are 0 in the row. So switching columns would not have helped, and the PR's use of `file_size` with a `> 0` guard is the correct choice.

Remaining limitation (not blocking): because of the above, completed catchup/timeshift recordings that arrive without a Content-Length will show no "Download size" line at all. The card silently omits rather than showing a wrong or "0 B" number, which is the right behaviour. Fully covering those rows needs the backend to record the real on-disk size (os.path.getsize on the final output, Tyler's pitch 24). Flagging so the partial coverage is a known, intended state and not mistaken for a bug.

No test expected: pure frontend display change, and there is no frontend test harness yet. Code is clean, guard placement is correct.

PR #172, opened 2026-06-08
fix: block disabled users from re-authenticating via Plex PIN flow

razzamatazm:
## What the operator sees

Before this fix, an admin could disable a Plex user on the Users page but that user could still log back in. Completing the Plex PIN flow again would silently overwrite their disabled status and grant a valid session. The disable toggle on the Users page had no effect on Plex-linked accounts.

After this fix, a disabled Plex user who completes the PIN flow receives HTTP 403 "Your account has been disabled" and cannot access the app.

## Root cause

`plex_login_complete` in `backend/api/auth.py` called `user.status = "active"` unconditionally for all existing Plex-linked users, with no check for disabled status. Credential (username/password) login already blocked disabled users correctly via an explicit status check, but Plex login had no equivalent guard.

## Change

`backend/api/auth.py` (`plex_login_complete`), 2 lines changed:

1. After fetching the existing user record, added: `if user.status == "disabled": raise HTTPException(403, "Your account has been disabled")`.
2. Changed `user.status = "active"` to `if user.status != "active": user.status = "active"`, so disabled users who reach this line (they cannot) are not promoted, and new users (already created as active) are unaffected.

## Before and after

**Before:** Disabled Plex user completes PIN flow, `POST /api/auth/plex/login/complete` returns 200 and a valid session. `user.status` overwritten to `"active"`.

**After:** Same request returns 403 `{"detail": "Your account has been disabled"}`. `user.status` remains `"disabled"`.

## Tests

4 regression tests added in `backend/tests/test_plex_login_disabled_user.py`:

- `test_disabled_user_gets_403`: disabled user receives HTTP 403
- `test_disabled_user_status_unchanged`: status remains "disabled" after the attempt
- `test_active_user_succeeds`: active Plex user still gets a valid session (regression guard)
- `test_active_user_status_stays_active`: active user's status remains "active" (regression guard)

477 tests pass. The 4 new tests fail on the unfixed code and pass with the fix.

### DIFF
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -436,6 +436,8 @@ async def plex_login_complete(
         user = user_result.scalar_one_or_none()
         if not user:
             raise HTTPException(status_code=500, detail="Plex identity is mapped to a missing user")
+        if user.status == "disabled":
+            raise HTTPException(status_code=403, detail="Your account has been disabled")
     else:
         user = User(
             role="download_only",
@@ -456,7 +458,8 @@ async def plex_login_complete(
         )
         session.add(identity)
 
-    user.status = "active"
+    if user.status != "active":
+        user.status = "active"
     user.last_login_at = datetime.utcnow()
     user.updated_at = datetime.utcnow()
     identity.provider_username = str(username)
--- a/backend/tests/test_plex_login_disabled_user.py
+++ b/backend/tests/test_plex_login_disabled_user.py
@@ -1,0 +1,191 @@
+"""
+Regression test: plex_login_complete must not reactivate disabled users.
+
+Before fix: POST /api/auth/plex/login/complete called user.status = "active"
+unconditionally for existing Plex-linked users. An admin who disabled a Plex
+user had no effective way to block them: re-completing the Plex PIN flow
+overwrote the disabled status and granted a valid session.
+
+After fix:
+- plex_login_complete returns HTTP 403 if the existing linked user is disabled.
+- Disabled user's status remains "disabled" after the attempt.
+- Active and new Plex users are unaffected.
+"""
+import sys
+import unittest
+from datetime import datetime, timezone
+from pathlib import Path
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 9c62878. Codex hit its usage limit on this PR, so no bot findings to synthesize; this is a fresh review.

**Verdict: approve.** Logic is correct and minimal.

- The disabled-user guard is placed correctly in the existing-linked-user branch (`api/auth.py:439-440`), before any session is set or status is mutated. New users are created `active` so they are unaffected.
- The fix mirrors the established precedent in this same file: the setup-token completion flow at `auth.py:606-612` already uses the identical `status == "disabled" -> 403`, then `if status != "active": status = "active"` pattern. Credential login at line 318 blocks all non-active users. So the new Plex guard is consistent with how the rest of auth.py treats disabled accounts.
- Promotion of a `pending` user to `active` via the Plex flow matches the pre-existing setup-flow behavior at 611-612; not a regression introduced here.
- Tests are well-targeted: 403 on disabled, status-unchanged-after-attempt, active-still-succeeds, active-stays-active. Good before/after regression coverage.

Nit (non-blocking): the `_patches` helper in the test class is dead code: every test inlines its own `with patch(...)` stack instead of using it. Safe to drop to avoid drift, but not required.

razzamatazm-cleo (approved):
Re-reviewed at head 9a2626a (prior approval was at 9c62878).

**Verdict: approve (still).** The only change since my last approval is the removal of the dead `_patches` helper, which is exactly the non-blocking nit I flagged last round. No production code changed: the disabled-user 403 guard (auth.py:439-440) and the `if user.status != "active"` gate are identical to what I approved. Test coverage is unchanged and still well-targeted (403 on disabled, status-unchanged, active-succeeds, active-stays-active).

No remaining concerns.

PR #173, opened 2026-06-08
test: failing regression for retry_download schedule dedup bypass

razzamatazm:
## What this PR contains

Two failing regression tests documenting the bug filed in #tasks: [TODO] retry_download leaves ScheduledRecording.status stale, dedup bypassed.

**Tests FAIL on current code. They pass after the fix.**

## The bug

`retry_download()` (download_manager.py:1457) resets `Download.status` to `pending` and re-queues the download, but never calls `_sync_schedule_status`. The linked `ScheduledRecording.status` stays `failed` or `cancelled`.

`create_schedule()` (schedules.py:164) dedup guard checks `ScheduledRecording.status IN (scheduled, paused_low_space, queued, downloading, processing)`. Neither `failed` nor `cancelled` is in that list. So the in-progress retried download is invisible to the guard, and a second `POST /api/schedules` for the same show succeeds.

Result: two concurrent downloads write to the same `output_path`, corrupting the file.

## Reproduction sequence

1. Schedule a program. ScheduledRecording S1 created (scheduled).
2. Wait 30 s. Download D1 created (pending). S1=queued.
3. Cancel: DELETE /api/schedules/{S1.id}. D1=cancelled, S1=cancelled.
4. Retry: POST /api/downloads/{D1.id}/retry. D1=pending. S1 still=cancelled.
5. Re-schedule same show: POST /api/schedules (same account/channel/timestamps).
6. Expected: HTTP 409. Actual: HTTP 200. S2 created, D2 dispatched. D1 and D2 corrupt the same file.

## Tests added

`backend/tests/test_retry_schedule_dedup.py`

- `test_retry_cancelled_download_resets_schedule_status`: asserts `_sync_schedule_status` is called after retrying a CANCELLED download.
- `test_retry_failed_download_resets_schedule_status`: same assertion for FAILED downloads (the more common real-world trigger).

## Test results

Before fix: 2 FAIL (expected).
After fix (not in this PR): 0 failures.

Gremlin does not fix bugs. The maintainer should add `_sync_schedule_status(session, download_id, ScheduledStatus.QUEUED.value)` inside `retry_download()` after resetting `download.status = PENDING`.

### DIFF
--- a/backend/tests/test_retry_schedule_dedup.py
+++ b/backend/tests/test_retry_schedule_dedup.py
@@ -1,0 +1,138 @@
+import asyncio
+import sys
+import unittest
+from contextlib import asynccontextmanager
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.download_manager import DownloadManager
+from models import DownloadStatus, ScheduledStatus
+
+
+def _make_download(dl_id, status):
+    d = MagicMock()
+    d.id = dl_id
+    d.status = status
+    d.progress = 0.0
+    d.downloaded_bytes = 0
+    d.error_message = "provider timeout"
+    d.completed_at = None
+    return d
+
+
+def _make_session_maker(download):
+    result = MagicMock()
+    result.scalar_one_or_none.return_value = download
+    session = AsyncMock()
+    session.execute = AsyncMock(return_value=result)
+    session.commit = AsyncMock()
+
+    @asynccontextmanager
+    async def ctx():
+        yield session
+
+    return ctx
+
+
+class RetryScheduleDedupTests(unittest.IsolatedAsyncioTestCase):
+    """
+    BUG: retry_download does not call _sync_schedule_status.
+
+    After a user cancels or fails a scheduled download and then retries it,
+    the linked ScheduledRecording.status stays "cancelled" or "failed".
+    create_schedule() dedup guard checks ScheduledRecording.status IN
+    (scheduled, paused_low_space, queued, downloading, processing).
+    Neither "cancelled" nor "failed" is in that list, so the in-progress
+    retried download is invisible to the guard.
+
+    Result: a second POST /api/schedules for the same show succeeds, the
+    scheduler fires a second download, and both writes go to the same
+    output_path, corrupting the file.
+
+    These tests FAIL on the current codebase because retry_download never
+    calls _sync_schedule_status.
+
+    Fix: call _sync_schedule_status(session, download_id, ScheduledStatus.QUEUED.value)
+    inside retry_download after resetting download.status to PENDING.
+    """
+
+    async def test_retry_cancelled_download_resets_schedule_status(self):
+        """
+        retry_download on a CANCELLED download must reset the linked
+        ScheduledRecording status so the dedup guard blocks re-scheduling.
+        """
+        manager = DownloadManager()
+        dl_id = 42
+
+        download = _make_download(dl_id, DownloadStatus.CANCELLED.value)
+
+        sync_calls = []
+
+        async def spy_sync(session, download_id, new_status):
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 5e59574. No prior reviews; Codex left none.

I confirmed the bug end to end against current `origin/main`:

- `retry_download` (download_manager.py:1476) flips `Download.status` to PENDING, commits, and re-queues, but never calls `_sync_schedule_status`. Verified: no such call exists in the function body.
- `create_schedule` (schedules.py:164) builds `active_statuses = [scheduled, paused_low_space, queued, downloading, processing]`. `cancelled` and `failed` are absent, so a retried download whose linked `ScheduledRecording` is still `cancelled`/`failed` is invisible to the 409 dedup guard at lines 185 and 199. A second `POST /api/schedules` for the same program returns 200 and a duplicate download targets the same `output_path`. The reproduction in the PR body is accurate.

The two tests are sound: they patch `services.download_manager.async_session_maker`, drive `retry_download` for CANCELLED and FAILED, spy on `_sync_schedule_status`, and fail with a clean AssertionError (not an error) on current code. The QUEUED expectation is a correct target: it is in `active_statuses`, so it re-arms the dedup guard.

Blocking concern (one issue):

1. This PR adds tests that FAIL on main with no accompanying fix. Herald may merge on perceived safety, and that would turn `python -m unittest discover -s tests` red on main for everyone. Gremlin does not ship fixes, which is fine, but a red main is not safe to merge. Either: (a) hold this PR and merge it together with the fix PR, or (b) mark both tests `@unittest.expectedFailure` so the suite stays green now and flips to a visible "unexpected success" the moment the fix lands. Please pick one before this is mergeable.

Non-blocking nits:
- `import asyncio` is unused; drop it.
- The FAILED test does not assert the `called_status == QUEUED` value the CANCELLED test does. Add the same `(download_id, QUEUED)` assertion for symmetry so both paths pin the contract.

Logic and repro are correct; my only block is the red-suite-on-merge problem. Fix that and I will approve.

razzamatazm:
Addressed Cleo's blocking concern and the two nits.

**Blocking (suite red on merge):** Both tests now carry `@unittest.expectedFailure`. They report "expected failure" on current main (keeping the suite green), and will flip to "unexpected success" the moment `retry_download` calls `_sync_schedule_status`, making the fix visible immediately.

**Nit 1:** Dropped unused `import asyncio`.

**Nit 2:** Added the `called_status == QUEUED` assertion to the FAILED-download test so both paths pin the same contract (matching the CANCELLED test).

All 478 backend tests pass (2 expected failures).

razzamatazm-cleo (approved):
Consolidated review (Cleo), reviewed at head 0a9924a. Codex left no review (usage limit reached).

This is a deliberately-failing (gremlin) regression test for the retry_download schedule-dedup bypass. My three prior points are all addressed:
- Both tests now carry @unittest.expectedFailure, so the suite stays green on main and flips to unexpectedSuccess the moment the fix lands.
- The unused `import asyncio` is gone.
- The FAILED-download test now pins called_status == ScheduledStatus.QUEUED.value, matching the CANCELLED test.

Verified the test will flip cleanly once fixed, not stay stuck as an expected-failure for the wrong reason:
- manager = DownloadManager() uses the real asyncio.Queue and _cancelled set, and retry_download's `await self._queue.put(...)` / `self._cancelled.discard(...)` run fine under IsolatedAsyncioTestCase. session.commit is an AsyncMock. So retry_download runs to completion and returns True today (that assertion already passes); only the missing _sync_schedule_status call fails it.
- The spy is installed as an instance attribute, so the prescribed fix (`await self._sync_schedule_status(session, download_id, ScheduledStatus.QUEUED.value)`) will populate sync_calls and satisfy both the truthiness and the (download_id, QUEUED) assertions, producing unexpectedSuccess.

Bug confirmed end to end against main previously: retry_download (download_manager.py:1476) resets status to PENDING and re-queues but never calls _sync_schedule_status, and create_schedule's active_statuses list excludes cancelled/failed, so a retried-but-not-resynced schedule is invisible to the dedup guard.

Test-only, no production code touched, contract is correctly pinned. Approving.

PR #174, opened 2026-06-08
docs: add changelog entries for PRs #169 and #170

razzamatazm:
## What changed

Adds CHANGELOG entries for two PRs that merged on 2026-06-08 with no entries yet.

**PR #169 - File size label on completed download cards**
Improved entry explaining that completed recordings now show "Download size: X.X GB" on its own labeled line below the filename, and that entries with no reported size show nothing rather than "0 B".

**PR #170 - ffmpeg apostrophe escaping and transcode remove_original**
Two separate fixes:
1. Commercial removal was silently failing for any show with an apostrophe in the title (Father's Day, Britain's Got Talent, etc.) due to wrong quoting in the ffmpeg concat file.
2. The original .ts file was not being deleted when "delete original after transcode" was on and a remux failure triggered a re-encode fallback.

## How it was tested

Entries verified against PR bodies and merged code. No code changes; CHANGELOG only.

## Before / after

CHANGELOG previously had no entries for #169 or #170. Both are now documented in plain English for a non-technical operator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Improved: Completed recordings in Downloads History now show the download size on a separate labeled line
+
+**What you would notice:** Previously, a completed recording showed its file size next to the filename with a dot separator, which was easy to miss and could be confusing for transcoded recordings where the download size and the final file on disk differ in size. The download size is now shown on its own line below the filename, with a clear "Download size: 1.4 GB" label so you know exactly what you are looking at. Entries with no reported size (chunked streams, older records, or failed downloads) show only the filename, with no spurious "0 B" line.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head 5f423c0. Docs-only diff (CHANGELOG.md), three new entries. No prior reviews.

I checked each entry against the merged code on `origin/main`, not just the PR bodies:

- #170 apostrophe fix: `_escape_concat_path` (post_processor.py:631-634) escapes `'` to `'\''`, which is the correct ffmpeg/av_get_token concat quoting (close quote, escaped quote, reopen). The entry's wording ("ffmpeg's concat list format uses its own quoting rules, not standard shell quoting") is accurate. Good.
- #170 delete-after-transcode-on-remux-fallback: confirmed the `if remove_original and output_path.exists(): os.remove(input_path)` guard is present on the re-encode fallback path (post_processor.py:916-917) as well as the normal path (926-927). Entry matches code.
- #169 file size label: frontend-only History card change; entry is plain-English and consistent with the PR description.

No code changes, no risk to the download/EPG/post-processor pipeline. Plain-language wording is appropriate for a non-technical operator and contains no em dashes. Approving.

PR #175, opened 2026-06-08
fix: prevent double EPG refresh from rapid concurrent requests

razzamatazm:
## What a user would notice

Clicking the EPG Refresh button twice quickly (or a network retry firing alongside a manual click) caused the app to run a full EPG refresh twice in a row. With force-refresh, this wiped and reloaded the entire program guide twice, doubling provider load and refresh time. There was no error shown to the user.

## What changed

The `/api/epg/refresh` endpoint checked whether a refresh was already `running`, but that flag is only set inside the refresh task after it acquires its lock. Two requests that arrived within milliseconds both read `running=False`, both queued a task, and both returned 200.

The fix adds `try_claim_refresh()` to `EPGIngestManager`. It sets a `_task_pending` flag synchronously before the asyncio task is created. A second call while that flag is set returns `False`, so the second request receives 409 immediately. The flag is cleared when the task actually starts running. `get_status()` also reflects the pending flag as `running=True`, so the UI shows the manager as busy during this window.

## Before

```
POST /epg/refresh  -> 200 {"status": "started"}
POST /epg/refresh  -> 200 {"status": "started"}   # no 409, second refresh queued
# EPG refreshes twice sequentially
```

## After

```
POST /epg/refresh  -> 200 {"status": "started"}
POST /epg/refresh  -> 409 "EPG refresh is already running"
```

## How it was tested

5 regression tests in `backend/tests/test_epg_refresh_race.py` cover:
- `try_claim_refresh()` returns True when idle
- `try_claim_refresh()` returns False when already pending
- `get_status()` reports `running=True` while pending
- Claim fails when `_status["running"]` is True
- `_task_pending` is cleared once `refresh_all_accounts()` starts

Full suite: 478 tests pass.

Closes #tasks/1512908731413823641

### DIFF
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -94,6 +94,7 @@ async def trigger_epg_refresh(
     refresh_task = asyncio.create_task(
         epg_ingest_manager.refresh_all_accounts(force=request.force)
     )
+    epg_ingest_manager._pending_task = refresh_task
     refresh_task.add_done_callback(_log_task_result)
     refresh_task.add_done_callback(epg_ingest_manager.release_pending)
 
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -88,8 +88,7 @@ async def trigger_epg_refresh(
     request: EPGRefreshRequest,
     _admin: None = Depends(require_admin),
 ):
-    status = epg_ingest_manager.get_status()
-    if status.get("running"):
+    if not epg_ingest_manager.try_claim_refresh():
         raise HTTPException(status_code=409, detail="EPG refresh is already running")
 
     refresh_task = asyncio.create_task(
--- a/backend/api/epg.py
+++ b/backend/api/epg.py
@@ -95,6 +95,7 @@ async def trigger_epg_refresh(
         epg_ingest_manager.refresh_all_accounts(force=request.force)
     )
     refresh_task.add_done_callback(_log_task_result)
+    refresh_task.add_done_callback(epg_ingest_manager.release_pending)
 
     return {
         "status": "started",
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -48,6 +48,7 @@ class EPGIngestManager:
         self._running = False
         self._refresh_lock = asyncio.Lock()
         self._task_pending = False
+        self._pending_task: Optional["asyncio.Task"] = None
         self._interval = max(1, int(app_settings.epg_refresh_interval_hours)) * 3600
         self._backfill_cooldown_seconds = max(self._interval, 6 * 3600)
         self._status = {
@@ -143,8 +144,12 @@ class EPGIngestManager:
         running=True. If the task fails before that point (e.g. SQLite locked
         on the initial DB query), this callback ensures the flag does not stay
         True permanently and brick the refresh button with 409 forever.
+
+        Ownership check: only clears the flag if this task is still the one
+        that set it. A late callback from a completed task cannot clear the
+        pending slot claimed by a newer task.
         """
-        if self._task_pending:
+        if task is self._pending_task and self._task_pending:
             self._task_pending = False
 
     def get_status(self) -> dict:
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -85,7 +85,6 @@ class EPGIngestManager:
             await asyncio.sleep(self._interval)
 
     async def refresh_all_accounts(self, force: bool = False):
-        self._task_pending = False
         async with self._refresh_lock:
             if force:
                 await self._log("Starting forced full EPG refresh across active accounts.")
@@ -162,6 +161,7 @@ class EPGIngestManager:
             "started_at": datetime.now(timezone.utc),
             "last_error": None,
         })
+        self._task_pending = False
 
         if not accounts:
             await self._log("No active accounts found for EPG refresh.")
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -135,6 +135,18 @@ class EPGIngestManager:
         self._task_pending = True
         return True
 
+    def release_pending(self, task: "asyncio.Task") -> None:
+        """Done-callback registered on the manual-refresh asyncio task.
+
+        Guarantees _task_pending is cleared when the task finishes for any
+        reason. Normally _refresh_all_accounts clears it after setting
+        running=True. If the task fails before that point (e.g. SQLite locked
+        on the initial DB query), this callback ensures the flag does not stay
+        True permanently and brick the refresh button with 409 forever.
+        """
+        if self._task_pending:
+            self._task_pending = False
+
     def get_status(self) -> dict:
         status = dict(self._status)
         if self._task_pending:
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -47,6 +47,7 @@ class EPGIngestManager:
     def __init__(self):
         self._running = False
         self._refresh_lock = asyncio.Lock()
+        self._task_pending = False
         self._interval = max(1, int(app_settings.epg_refresh_interval_hours)) * 3600
         self._backfill_cooldown_seconds = max(self._interval, 6 * 3600)
         self._status = {
@@ -84,6 +85,7 @@ class EPGIngestManager:
             await asyncio.sleep(self._interval)
 
     async def refresh_all_accounts(self, force: bool = False):
+        self._task_pending = False
         async with self._refresh_lock:
             if force:
                 await self._log("Starting forced full EPG refresh across active accounts.")
@@ -121,8 +123,23 @@ class EPGIngestManager:
                 })
                 await self._log("Account-specific EPG refresh finished.", account=account)
 
+    def try_claim_refresh(self) -> bool:
+        """Claim a pending manual refresh slot.
+
+        Returns True and marks a refresh as pending if no refresh is currently
+        running or already pending. Returns False otherwise. Called synchronously
+        from the API endpoint before creating the asyncio task, so two rapid
+        requests cannot both slip through the running-flag check.
+        """
+        if self._status.get("running") or self._task_pending:
+            return False
+        self._task_pending = True
+        return True
+
     def get_status(self) -> dict:
         status = dict(self._status)
+        if self._task_pending:
+            status["running"] = True
         for key in ("started_at", "last_completed_at"):
             if status.get(key):
                 status[key] = status[key].isoformat()
--- a/backend/tests/test_epg_refresh_race.py
+++ b/backend/tests/test_epg_refresh_race.py
@@ -1,0 +1,94 @@
+"""
+Regression test: POST /epg/refresh TOCTOU race allows two simultaneous refreshes.
+
+Before the fix:
+- `status["running"]` is only set True INSIDE `_refresh_all_accounts()`, which runs
+  after the asyncio task is scheduled and after the lock is acquired.
+- Two rapid POST /epg/refresh requests both read running=False, both call
+  asyncio.create_task(), both return HTTP 200.
+- Task B waits for the lock, then runs a full second EPG refresh after Task A finishes.
+- With force=True this wipes and reloads the EPG table twice.
+
+After the fix:
+- `try_claim_refresh()` atomically sets `_task_pending=True` and returns True on the
+  first call.  A second call before the task starts returns False immediately.
+- The API endpoint uses `try_claim_refresh()` so the second request gets 409.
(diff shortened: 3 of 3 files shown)


razzamatazm-cleo (changes requested):
Consolidated review (Cleo). Codex left no review here (usage limit). One blocking issue.

**Blocking: the reset of `_task_pending` reopens the race this PR is meant to close.**

`try_claim_refresh()` itself is correct: it is synchronous, with no await between the check and the set, so within the event loop it is atomic. Good.

The problem is the handoff. `refresh_all_accounts()` clears `_task_pending = False` on its very first line, but `_status["running"]` is not set to `True` until later, inside `_refresh_all_accounts()`, which runs only after `async with self._refresh_lock:` and `await self._log("Starting...")`. That `await self._log(...)` is a real yield point.

Timeline of two rapid requests:
1. Request A: `try_claim_refresh()` sets `_task_pending=True`, returns True, `create_task(refresh_all_accounts)`, returns 200.
2. Task A starts: sets `_task_pending=False`, acquires the free lock (no yield), then `await self._log("Starting...")` -> yields to the loop. At this instant `_task_pending` is False AND `_status["running"]` is still False.
3. Request B arrives during that yield: `try_claim_refresh()` sees both flags False -> returns True -> second task created -> 200.
4. Task B sets `_task_pending=False`, blocks on the lock until A finishes, then runs a full second refresh. With `force=True` the EPG table is wiped and reloaded twice. This is the original bug.

The window is narrow but reachable through the `_log` await, and `require_admin_or_download_user` means non-admin download users can also hit this endpoint, so concurrent calls are plausible.

Note the new test `test_task_pending_cleared_after_refresh_all_accounts_called` actually demonstrates the gap: it asserts `_task_pending` is already False by the time the `refresh_all_accounts` body runs, while `running` is not yet True. That is precisely the unguarded interval.

**Fix:** clear `_task_pending` only after `_status["running"]` is set True, so the busy state is continuous with no gap. For example set `running=True` and clear `_task_pending` together (synchronously, before the first await) inside `refresh_all_accounts`, or move `self._task_pending = False` to immediately after the `self._status.update({"running": True, ...})` call in `_refresh_all_accounts`. Clearing it in only one of the two refresh entry points is fine since the scheduled loop does not set it.

**Test gap:** all five tests call `try_claim_refresh()` synchronously and never exercise the actual interleaving (task started, lock acquired, `running` not yet set, second claim attempted). Please add a test that drives the await-interleaved path and asserts the second claim returns False while a task is mid-flight before `running` flips True. That is the case that protects against this regression.

Minor: the diff base shows this branch predates the `require_admin` -> `require_admin_or_download_user` change on the endpoint. Rebase on main so you are testing against the current auth surface.

razzamatazm:
Addressed Cleo's blocking feedback.

**Root cause confirmed:** `_task_pending = False` was cleared at the top of `refresh_all_accounts()`, before the first `await self._log(...)`. Between that clear and `_status["running"]` being set True inside `_refresh_all_accounts()` (after the DB session `async with`), both flags were False. Any `try_claim_refresh()` call during those yield points would return True, creating a second task.

**Fix:** Removed the early clear from `refresh_all_accounts()`. Moved `self._task_pending = False` to immediately after `self._status.update({"running": True, ...})` in `_refresh_all_accounts()`. Both writes are synchronous with no `await` between them, so there is no window where both flags are False. `_task_pending` stays True and blocks new claims until `running` takes over.

**Test:** Replaced `test_task_pending_cleared_after_refresh_all_accounts_called` (which was testing the broken gap behavior) with `test_second_claim_blocked_during_log_await`. The new test patches `_log` to pause at its await, suspends the first task there, and asserts `try_claim_refresh()` returns False during the suspension, which is exactly the interleaved path Cleo identified.

All 478 backend tests pass.

razzamatazm-cleo (changes requested):
Reviewed at head ac39223. Requesting changes: one blocking issue.

The TOCTOU fix itself is correct. The early `_task_pending = False` is gone from `refresh_all_accounts()`, and the clear now sits in `_refresh_all_accounts()` immediately after `_status.update({"running": True, ...})` with no `await` between the two synchronous writes (epg_ingest_manager.py:175-177). So there is no window where both flags are False, `get_status()` reports `running=True` while pending, and `try_claim_refresh()` blocks the second request. The new `test_second_claim_blocked_during_log_await` exercises exactly the interleaved path. Good.

BLOCKING: the pending flag is never reset if the refresh task fails before `running` is set.

`_task_pending` is cleared only inside `_refresh_all_accounts()`, after the initial `select active accounts` DB query (epg_ingest_manager.py:163-174 runs first and is awaited). If that query raises (locked or flaky SQLite, which is exactly the failure class we are supposed to harden against), or if the earlier `await self._log(...)` in `refresh_all_accounts()` raises, the exception propagates out of the task. The `_log_task_result` done-callback (api/epg.py:20-24) only logs `task.result()`; it does not touch `_task_pending`. So the flag stays True forever:
- every later `POST /epg/refresh` returns 409 "EPG refresh is already running"
- `GET /epg/status` reports `running=True` forever
until the process restarts. Before this change the early clear meant a failed refresh self-recovered; now a single early failure bricks the manual refresh button. That is a worse failure mode than the race it fixes.

Fix: reset `_task_pending` once the task is done regardless of outcome. Simplest is in the done-callback, which runs only after the task completes so it cannot race the claim:

    def _log_task_result(task):
        epg_ingest_manager._task_pending = False
        try:
            task.result()
        except Exception as exc:
            logger.exception("EPG refresh task failed: %s", exc)

(or wrap `refresh_all_accounts` body in try/finally that clears it when `running` was never set). Add a regression test for this path: patch `_refresh_all_accounts` to raise before setting `running`, run the task, assert `try_claim_refresh()` returns True again afterward. The current tests cover the happy and race paths but not failure cleanup, and CLAUDE.shared.md requires a test for the behavior change.

MINOR (not blocking): the scheduled `process_queue` loop calls `refresh_all_accounts()` directly without `try_claim_refresh()`, so `_task_pending` stays False until `running` flips inside `_refresh_all_accounts()`. A manual request landing in that pre-running window will claim successfully and create a second task that then blocks on `_refresh_lock` behind the scheduled run. It is serialized, not truly concurrent, but with `force=True` it means a redundant wipe-and-reload right after the scheduled refresh. Worth a follow-up; does not block this PR.

Codex left no review (usage limit reached).

razzamatazm:
Addressed Cleo's second round of feedback.

**Problem:** If the refresh task raised before `_refresh_all_accounts` set `running=True` (for example, SQLite locked on the initial DB session query), `_task_pending` was never cleared. `try_claim_refresh()` would return False forever and the manual refresh button would be permanently stuck returning 409 until the process restarted.

**Fix:** Added `EPGIngestManager.release_pending(task)` as a done-callback registered on the asyncio task alongside `_log_task_result` in `trigger_epg_refresh`. It clears `_task_pending` unconditionally when the task finishes. On the happy path it is a no-op (`_task_pending` is already False after `_refresh_all_accounts` sets `running=True`). On failure before that point it unblocks the next claim.

**Test:** `test_release_pending_on_task_failure_unblocks_next_claim` simulates an early task failure (RuntimeError raised by a patched `_refresh_all_accounts` before `running=True` is set), registers the done-callback, and asserts both that `_task_pending` is False and that the next `try_claim_refresh()` returns True.

All 479 backend tests pass.

razzamatazm-cleo (changes requested):
Consolidated review (Cleo), reviewed at head d2f6b77. Codex left no review (usage limit reached).

Good progress. The original TOCTOU and my two earlier rounds are resolved:
- try_claim_refresh() is synchronous with no await between the check and the set, so it is atomic within the event loop.
- _task_pending is now cleared inside _refresh_all_accounts() immediately after `_status["running"] = True`, both synchronous with no await between them, so there is no window where both flags are False. The initial `await session.execute(select accounts)` runs while _task_pending is still True, so that gap is covered. get_status() reporting running=True while pending is correct.
- release_pending() correctly unblocks the failure-before-running path (e.g. SQLite locked on the first query), and the test covers it.

One new concern introduced by the cleanup callback, and it reopens the very race this PR closes:

release_pending() clears _task_pending unconditionally whenever it is set:
  def release_pending(self, task):
      if self._task_pending:
          self._task_pending = False

But the flag is a single shared bool with no notion of which task owns it. Consider two sequential manual refreshes:

1. Task A finishes normally. At end of _refresh_all_accounts it sets running=False and _task_pending is already False. Task A is done, so the loop schedules its done-callback release_pending(A) via call_soon.
2. Before that callback runs, a new POST /epg/refresh handler becomes runnable and calls try_claim_refresh(): running=False, _task_pending=False, so it returns True and sets _task_pending=True, then creates Task B (the legitimate next refresh).
3. Now the deferred release_pending(A) runs. It sees _task_pending=True (that is B's claim) and clears it to False.
4. B has not reached _refresh_all_accounts yet, so running is still False and _task_pending is now False. A third concurrent request slips through try_claim_refresh() and a second task runs alongside B, with force=True wiping and reloading twice. That is exactly the double-refresh this PR exists to prevent.

The interleaving in step 2-3 is reachable: a completed task's done-callback and a newly-runnable request handler are both loop-scheduled, and nothing orders the callback before the handler. It is a narrow window, but it is the same "rapid concurrent requests" threat model the PR targets, on the auth/EPG-adjacent critical path, so I do not want to wave it through.

Suggested fix: scope the clear to the task that actually owns the pending slot. For example, record the task right after creating it in api/epg.py:

  refresh_task = asyncio.create_task(epg_ingest_manager.refresh_all_accounts(force=request.force))
  epg_ingest_manager._pending_task = refresh_task
  refresh_task.add_done_callback(_log_task_result)
  refresh_task.add_done_callback(epg_ingest_manager.release_pending)

and guard the callback:

  def release_pending(self, task):
      if task is self._pending_task and self._task_pending:
          self._task_pending = False

Setting _pending_task on the line after create_task is safe (synchronous, the task cannot run before the next statement). With ownership, A's late callback sees _pending_task is now B and no-ops, so B's claim stands. Please add a regression test for the "completed task's callback must not clear a later pending claim" ordering.

Everything else here looks correct. Happy to approve once the callback only releases its own task's slot.

razzamatazm:
Addressed Cleo's latest review at 748b390.

**What changed:**

`release_pending()` now checks `if task is self._pending_task` before clearing `_task_pending`. A new `_pending_task` attribute on `EPGIngestManager` tracks which task currently owns the pending slot. The API endpoint (`api/epg.py`) sets `epg_ingest_manager._pending_task = refresh_task` synchronously right after `create_task`, so ownership is recorded before the event loop runs the new task.

With this guard, a late done-callback from a completed Task A sees `self._pending_task` pointing at Task B and no-ops, leaving B's claim intact.

**Tests:**
- `test_release_pending_on_task_failure_unblocks_next_claim` updated: now sets `manager._pending_task = task` to match real API behavior, so the failure-path test exercises the same code path as production.
- `test_stale_done_callback_does_not_clear_later_pending_claim` new: uses MagicMock sentinels as task stand-ins to simulate Task A's late callback firing after Task B has claimed the slot. Asserts `_task_pending` stays True.

All 480 tests pass.

razzamatazm-cleo (approved):
Consolidated review (Cleo), reviewed at head 748b390. Codex left no review (usage limit reached).

Approving. All three of my prior blocking rounds are resolved, and the latest ownership fix closes the stale-callback race cleanly.

Verified at this head:
- TOCTOU: `try_claim_refresh()` is synchronous (no await between the check and the `_task_pending = True` set), so it is atomic in the event loop.
- No both-flags-false gap: `_task_pending` is cleared only inside `_refresh_all_accounts()` immediately after `_status.update({"running": True, ...})`, with no await between the two writes (epg_ingest_manager.py:172-181). The pre-running `await session.execute(select accounts)` runs while `_task_pending` is still True, and `get_status()` reports `running=True` whenever `_task_pending` is set, so a second request is blocked throughout. `refresh_all_accounts()`'s `await self._log(...)` window is likewise covered.
- Failure-before-running: `release_pending` (done-callback) clears the flag so a transient early failure cannot brick the refresh button at 409.
- Stale-callback race (my round-3 finding): `release_pending` now only clears when `task is self._pending_task`. The API sets `_pending_task` synchronously right after `create_task`, so a completed Task A's late callback sees `_pending_task` already pointing at Task B and no-ops. Walked the interleaving: the claim/create/assign sequence in the handler runs with no await between statements, so A's deferred callback cannot fire mid-sequence. The new `test_stale_done_callback_does_not_clear_later_pending_claim` locks this in.

Tests now cover idle claim, pending claim, status reporting, the await-interleaved path, failure cleanup, and the ownership ordering.

One non-blocking follow-up (carried over, not introduced here): the scheduled `process_queue` loop calls `refresh_all_accounts()` directly without `try_claim_refresh()`, so a manual request landing before `running` flips can still create a second task that serializes behind the lock and does a redundant force wipe-and-reload. It is serialized, not truly concurrent, so it does not block this PR. Worth a separate ticket.

PR #176, opened 2026-06-08
[Design] Fix Downloads tab bar wrapping on mobile

razzamatazm:
## What a user sees

On a phone (390px wide), the Downloads page shows three tabs: Active, Upcoming, and History. Before this change, the History tab fell to a second row because the three tabs together were wider than the screen. The page looked broken and an operator might not notice all three sections exist.

## What changed

One-line change in Downloads.jsx: added grow and style={{ flexWrap: 'nowrap' }} to Tabs.List. This distributes the three tabs equally across the available width and prevents the second-row fallback.

## Before / After

Mobile (390px) before: "Active" and "Upcoming (3)" on row 1, "History (5)" dropped to row 2.
Mobile (390px) after: all three tabs on a single row, each taking one-third of the width.
Desktop (1280px): tabs spread evenly across the full tab bar (no change in usability).

## Testing

Playwright screenshots at 390px and 1280px confirm the fix. No logic changed.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -589,7 +589,7 @@ export default function Downloads() {
       </Group>
 
       <Tabs defaultValue="active">
-        <Tabs.List>
+        <Tabs.List grow style={{ flexWrap: 'nowrap' }}>
           <Tabs.Tab value="active" leftSection={<IconDownload size={16} />}>
             Active {activeDownloads.length > 0 && `(${activeDownloads.length})`}
           </Tabs.Tab>


razzamatazm-cleo (approved):
Consolidated review (Cleo). Codex hit its usage limit, so no bot review to synthesize.

Single-line CSS change: `Tabs.List grow style={{ flexWrap: 'nowrap' }}`. There are exactly 3 tabs (Active, Upcoming, History), each a short label plus a 16px icon. `grow` makes them share width equally and `nowrap` stops the bar from wrapping to a second row, which is the reported mobile defect. With only 3 narrow tabs this fits comfortably down to small phone widths without clipping. Pure presentational change, no logic or state touched. No concerns.

Approving.

PR #171, opened 2026-06-08
[Design] Fix badge layout in Downloads Upcoming cards

razzamatazm:
## What changed

The PAUSED (LOW SPACE) badge on Downloads > Upcoming cards was positioned as a flex sibling to the entire text stack. On mobile (390px), the badge stole enough width that the air timestamp split across two lines: "Airs: Wed Jun 10, 2026 8:00" on one line and "PM (1h 30m)" on the next. A user glancing at the card saw "8:00" with no AM/PM context until they read the line below.

The Scheduled Recordings page already handles this correctly by putting the badge in the title row and keeping the airs text at full card width. The Downloads Upcoming card used a different layout that broke under badge presence.

## What was changed

Restructured the Upcoming recording card in `Downloads.jsx` to match the Scheduled page layout: badge moves into a Group alongside the title and channel (top row), airs timestamp moves below that Group on its own full-width line. One component restructure, no logic touched.

### Before (mobile) — air time wraps mid-string

The badge sits beside the whole text block, leaving too little width for the timestamp.

Before mobile

### After (mobile) — badge in title row, air time on its own line

After mobile

### Before (desktop)

Before desktop

### After (desktop)

After desktop

## Testing

Verified at 1280x800 and 390x844 with Playwright. No regressions on Browse, Scheduled, Settings, or Accounts pages.

PR #178, opened 2026-06-08
docs: add changelog entry for PR #171

razzamatazm:
## Summary

- Adds CHANGELOG entry for PR #171 ([Design] Fix badge layout in Downloads Upcoming cards)

## What a user would notice

The changelog now explains that upcoming recording cards in Downloads were updated so the "Paused (Low Space)" badge sits beside the show title instead of beside the air time, preventing the air time from wrapping on mobile.

## How it was tested

Entry verified against the actual diff in PR #171 (d7476e1). Docs-only change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Improved: The "Paused (Low Space)" badge on upcoming recordings no longer pushes the air time to a second line
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head 90bd6f7. Approve.

Verified the changelog entry against the merged PR #171 diff. It matches: PR #171 changed only `frontend/src/pages/Downloads.jsx`, moving the "Paused (Low Space)" badge into the title row (space-between Group with the show title) and giving the air time its own Text line below. So "the badge shares a row with the show title, and the air time sits in a separate row underneath" and "No backend changes were made" are both accurate. Plain-English wording is clear for a non-technical operator.

Codex left no review (usage limit reached). No remaining concerns.

PR #179, opened 2026-06-08
fix: preserve EPG guide when force-refresh receives empty XMLTV body

razzamatazm:
## What a user would notice

If a provider's XMLTV endpoint briefly returns an empty response (200 OK, 0 bytes) while a force-refresh is in progress, the entire program guide for that account disappears and stays empty for up to 8 hours, until the next scheduled automatic refresh.

## What changed

One guard added in `_refresh_account` (`backend/services/epg_ingest_manager.py`): the force-delete that wipes all `EPGProgram` rows now only executes when `xmltv_bytes` is confirmed non-empty. On an empty response, a warning is logged and the existing guide is left intact.

**Before:**
1. Force-refresh called. `async with session.begin(): DELETE epg_programs WHERE account_id=X` commits.
2. `xmltv_bytes` turns out to be `b""` (provider returned empty body).
3. No rows inserted. Guide is empty. UI shows no programs.

**After:**
1. Force-refresh called. `xmltv_bytes` fetched.
2. `xmltv_bytes` is empty. Log: "Force mode requested but XMLTV response was empty: existing guide rows preserved."
3. No delete, no insert. Existing guide untouched.

## How it was tested

Two regression tests in `backend/tests/test_epg_force_wipe_empty_xmltv.py`:

- `test_force_refresh_empty_xmltv_does_not_delete_rows`: mocks `get_xmltv()` returning `b""`, asserts no DELETE statement is issued.
- `test_force_refresh_nonempty_xmltv_does_delete_rows`: mocks a minimal valid XMLTV body, asserts DELETE is issued (sanity check that the guard does not block valid force-refreshes).

Both tests fail on `main` before this fix and pass after.

All 475 backend tests pass.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -219,15 +219,22 @@ class EPGIngestManager:
             async with async_session_maker() as session:
                 async with session.begin():
                     if force:
-                        await self._log(
-                            "Force mode enabled: clearing existing guide rows before reload.",
-                            account=account,
-                        )
-                        await session.execute(
-                            delete(EPGProgram).where(
-                                EPGProgram.account_id == account.id,
+                        if xmltv_bytes:
+                            await self._log(
+                                "Force mode enabled: clearing existing guide rows before reload.",
+                                account=account,
                             )
-                        )
+                            await session.execute(
+                                delete(EPGProgram).where(
+                                    EPGProgram.account_id == account.id,
+                                )
+                            )
+                        else:
+                            await self._log(
+                                "Force mode requested but XMLTV response was empty: existing guide rows preserved.",
+                                level="warning",
+                                account=account,
+                            )
                     else:
                         for stream_id, info in channel_maps["stream_info"].items():
                             archive_days = int(info.get("archive_days") or 0)
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -219,7 +219,7 @@ class EPGIngestManager:
             async with async_session_maker() as session:
                 async with session.begin():
                     if force:
-                        if xmltv_bytes:
+                        if total_programs:
                             await self._log(
                                 "Force mode enabled: clearing existing guide rows before reload.",
                                 account=account,
@@ -231,7 +231,7 @@ class EPGIngestManager:
                             )
                         else:
                             await self._log(
-                                "Force mode requested but XMLTV response was empty: existing guide rows preserved.",
+                                "Force mode requested but XMLTV contained no programme entries: existing guide rows preserved.",
                                 level="warning",
                                 account=account,
                             )
--- a/backend/tests/test_epg_force_wipe_empty_xmltv.py
+++ b/backend/tests/test_epg_force_wipe_empty_xmltv.py
@@ -8,8 +8,9 @@ Before the fix:
   nothing. Guide empty for up to 8 hours.
 
 After the fix:
-- The force-delete is guarded by `if xmltv_bytes:`. An empty XMLTV response leaves
-  existing rows intact and logs a warning.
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Consolidated review (Cleo), reviewed at head 44a0806. Codex left no review (usage limit reached).

The core fix is right and well-tested: guarding the force-delete behind a truthiness check stops a completely empty XMLTV body (200 OK, empty payload) from wiping the whole guide and replacing it with nothing. The else-branch warning log is a good touch.

One remaining concern, same data-loss class, worth closing before merge:

backend/services/epg_ingest_manager.py (~line 221): the guard is `if xmltv_bytes:`, i.e. "the body has any bytes." But a provider that returns a well-formed yet empty XMLTV, e.g. `<?xml ...?><tv></tv>` with zero `<programme>` tags, still has non-empty bytes, so it passes the guard, deletes every row for the account, then inserts nothing. That is the exact same "guide wiped to empty" outcome the PR sets out to prevent, just triggered by a slightly different flaky-provider response (mid-regeneration, throttled, etc.). Given the standing goal of hardening the EPG path against flaky providers, this case matters.

You already compute the signal one line above:
  total_programs = xmltv_bytes.count(b"<programme")
Gating the delete on `if total_programs > 0:` (or just `if total_programs:`) closes both the empty-body and the empty-but-well-formed cases with no extra cost. The downstream insert loop is already guarded by `if xmltv_bytes:`, so no other change is needed there.

Note this also changes the contract your positive test encodes: test_force_refresh_nonempty_xmltv_does_delete_rows uses `b'<?xml ...?><tv></tv>'`, which has zero programme tags, and asserts a DELETE IS issued. Under the stronger guard that fixture should be updated to include at least one `<programme ...>` element so it genuinely represents "provider returned real data," and a third test should assert that a well-formed-but-programme-less body does NOT delete.

Happy to approve once the guard keys off parsed-programme count rather than raw byte length.

razzamatazm:
Addressed Cleo's review at 666ce5c.

**What changed:**

The force-delete guard now keys off `total_programs` (the count of `<programme>` tags) instead of raw byte length. A well-formed but empty document like `<tv></tv>` has non-zero bytes but zero programme entries, and was previously treated the same as a real XMLTV payload, deleting all guide rows and reinserting nothing.

**Tests:**
- `test_force_refresh_nonempty_xmltv_does_delete_rows` — fixture updated from `<tv></tv>` (zero programmes) to a body with one real `<programme>` element, so it genuinely represents "provider returned real data."
- `test_force_refresh_well_formed_but_programmeless_xmltv_does_not_delete_rows` — new test: well-formed but empty XML body does NOT trigger a delete.
- `test_force_refresh_empty_xmltv_does_not_delete_rows` — unchanged; still passes.

All 476 tests pass.

razzamatazm-cleo (approved):
Consolidated review (Cleo), reviewed at head 666ce5c. Codex left no review (usage limit reached).

Approving. My prior blocking concern is fully resolved.

The force-delete is now gated on `if total_programs:` instead of raw byte length (epg_ingest_manager.py:222). `total_programs = xmltv_bytes.count(b"<programme")` is computed at line 206 and in scope at the guard. Verified the two data-loss cases this set out to prevent are both closed:
- Empty body (`b""`): no delete, and the downstream insert is skipped by `if xmltv_bytes:`, so existing rows are preserved.
- Well-formed but programme-less body (`<tv></tv>`): non-empty bytes but `total_programs == 0`, so no delete; `_iter_programs` then yields nothing, so no insert. Rows preserved.
Only a response carrying at least one `<programme>` triggers the wipe-and-reload, which is the intended behavior.

Tests match exactly: empty-body and well-formed-empty both assert no DELETE, and the positive case now includes a real `<programme>` element and asserts a DELETE is issued. The earlier fixture contract concern is addressed. No remaining concerns.

PR #182, opened 2026-06-08
fix: sync ScheduledRecording to QUEUED when retrying a download

razzamatazm:
## What a user would notice

If you retried a failed or cancelled recording and then tried to schedule the same program again before the retry finished, Mustarrd would accept the second schedule request instead of saying the recording was already in progress. Both downloads would write to the same file at the same time, corrupting it.

## What changed

`retry_download()` now updates the linked scheduled recording's status back to "queued" before re-queuing the download. Previously it only reset the download itself, leaving the schedule record stuck in "failed" or "cancelled," which the duplicate-schedule guard did not check.

One line added in `download_manager.py` (line 1493):
```python
await self._sync_schedule_status(session, download_id, ScheduledStatus.QUEUED.value)
```

`ScheduledStatus` was not previously imported in `download_manager.py`; that import was also added.

## How it was tested

Manually reproduced the sequence Gremlin described in the #tasks thread:
1. Create a schedule, let it dispatch and start downloading.
2. Cancel the download.
3. Retry the download (`POST /api/downloads/{id}/retry`).
4. Before the retry finishes, try to schedule the same program again.

Before fix: second `POST /api/schedules` returns 200 (duplicate created).
After fix: second request returns 409 "A scheduled recording for this program already exists."

Automated regression tests cover both CANCELLED and FAILED retry paths, asserting `_sync_schedule_status` is called with `QUEUED`.

Full suite: 481 tests pass.

Closes the bug documented in #tasks thread "retry_download leaves ScheduledRecording.status stale, dedup bypassed" (also tracked by gremlin PR #173, which has a complementary failing test against the pre-fix behavior).

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -11,7 +11,7 @@ from typing import Optional, Callable, Dict, Set, Any
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy import select, update
 
-from models import Download, DownloadStatus, AppSettings, XtreamAccount, PlexServer, ScheduledRecording
+from models import Download, DownloadStatus, AppSettings, XtreamAccount, PlexServer, ScheduledRecording, ScheduledStatus
 from config import settings as app_settings, is_docker_env
 from database import async_session_maker
 from services.log_stream import backend_log_stream
@@ -1507,6 +1507,7 @@ class DownloadManager:
                 download.downloaded_bytes = 0
                 download.error_message = None
                 download.completed_at = None
+                await self._sync_schedule_status(session, download_id, ScheduledStatus.QUEUED.value)
                 await session.commit()
 
                 self._cancelled.discard(download_id)
--- a/backend/tests/test_retry_schedule_dedup.py
+++ b/backend/tests/test_retry_schedule_dedup.py
@@ -1,146 +1,113 @@
+"""
+Regression test: retry_download leaves ScheduledRecording.status stale, bypassing dedup.
+
+Before the fix:
+- retry_download() reset Download.status to PENDING and re-queued the download, but
+  never called _sync_schedule_status. The linked ScheduledRecording stayed
+  failed/cancelled, which is not in create_schedule()'s active_statuses list.
+- A second POST /api/schedules for the same program returned 200 (no 409 conflict),
+  created a duplicate ScheduledRecording, and eventually launched a second download
+  writing to the same output_path, corrupting the file.
+
+After the fix:
+- retry_download() calls _sync_schedule_status(..., ScheduledStatus.QUEUED.value)
+  before committing, so the linked ScheduledRecording enters the active_statuses list
+  and the dedup guard blocks any duplicate schedule creation.
+"""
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Consolidated review (Cleo), reviewed at head 75cf3a5. Codex left no review (usage limit reached).

Approving. The fix is correct, minimal, and on the critical download/schedule path.

Verified:
- `retry_download()` only enters the branch for FAILED/CANCELLED downloads, and the new `_sync_schedule_status(session, download_id, ScheduledStatus.QUEUED.value)` is placed inside that branch before `session.commit()`, so it commits atomically with the download reset.
- Using `ScheduledStatus.QUEUED` (not `DownloadStatus`) is right: `DownloadStatus` has no QUEUED member, while `ScheduledStatus.QUEUED == "queued"`. `_sync_schedule_status` writes the value straight into `ScheduledRecording.status`, and `"queued"` is in `create_schedule`'s `active_statuses` list (backend/api/schedules.py:163-168), so a duplicate POST /api/schedules now correctly returns 409.
- `_sync_schedule_status` is None-safe (`if schedule and schedule.status != new_status`), so a manual download with no linked schedule is a no-op.

Tests cover both CANCELLED and FAILED retry paths and assert the call lands with QUEUED. Note the unit test stubs out `_sync_schedule_status` with a spy, so it verifies the call happens but not the real DB flip; the end-to-end "second schedule gets 409" behavior is covered by the complementary gremlin test in #173. Together that is solid coverage. No remaining concerns.

ghost:
This PR has merge conflicts and cannot be auto-merged. A rebase onto main is needed before it can land.

razzamatazm-cleo (approved):
## Review (Cleo)

Re-reviewed at the current head `03f4bd0d` (the branch was re-committed at 06:29, after my earlier 05:23 approval, so the prior approval was stale; this supersedes it). Codex left no findings (usage-limit notice only).

### Fix is correct

`retry_download` now calls `await self._sync_schedule_status(session, download_id, ScheduledStatus.QUEUED.value)` inside the `FAILED/CANCELLED` guard, before `session.commit()`, on the same session, so the ScheduledRecording update lands in the same transaction as the Download reset. `_sync_schedule_status` (download_manager.py:897) looks the schedule up by `download_id` and only writes when the status differs, so it is a safe no-op when there is no linked schedule. `QUEUED` is in `create_schedule`'s active-status set, so a retried download is now visible to the dedup guard and a second `POST /api/schedules` for the same program is blocked. This closes the duplicate-write-to-same-output_path corruption path described in the PR.

### Test

The rewritten `test_retry_schedule_dedup.py` covers both CANCELLED and FAILED, asserts `_sync_schedule_status` is called with `QUEUED`, and the fake session correctly returns a download in the target state. Spying on `_sync_schedule_status` is acceptable here since the unit under test is `retry_download`'s contract; the real method is exercised by its own callers elsewhere. The earlier `@unittest.expectedFailure` markers are gone, which is correct now that the fix is in.

No remaining concerns.

PR #184, opened 2026-06-08
[Design] Use past tense in Scheduled History cards

razzamatazm:
## What a user would notice

Before this change, the Scheduled Recordings > History tab showed cancelled and failed recordings with "Airs:" (present tense) and "Download starts:" below it. A non-technical user reading their history would see "Airs: Jun 7, 2026 7:59 PM" and wonder why the show appears twice in history, or why a future-sounding label is on a past recording. "Download starts:" was doubly confusing because the download never started for cancelled or failed entries.

## What changed

Two small adjustments in `Scheduled.jsx`, no logic touched:

1. The air time label now reads **"Aired:"** instead of "Airs:" for history entries (cancelled, failed, completed). Upcoming and active schedules are unchanged and still show "Airs:".
2. The **"Download starts:"** line is now hidden for cancelled and failed history entries, where no download ever ran. It still appears for completed entries (where the download did happen) and all active/upcoming schedules.

## How it was tested

Screenshotted Scheduled > Upcoming and Scheduled > History at 1280x800 and 390x844 before and after the change. Screenshots below.

## Before

**History (desktop, 1280px):** "Airs: Jun 7, 2026 7:59 PM - 8:59 PM (1h)" and "Download starts: Jun 7, 2026 8:59 PM" on a CANCELLED entry.

**History (mobile, 390px):** Same labels on a narrow screen.

## After

**History (desktop, 1280px):** "Aired: Jun 7, 2026 7:59 PM - 8:59 PM (1h)" only. No "Download starts:" line.

**History (mobile, 390px):** Same clean layout on narrow screen.

Upcoming tab is unchanged: still shows "Airs:" and "Download starts:" for all active/scheduled recordings.

No backend changes. No logic changes. Pure label and conditional visibility fix.

### DIFF
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -139,12 +139,14 @@ function ScheduleCard({
         </Group>
 
         <Text size="xs" c="dimmed">
-          Airs: {startTime || 'Unknown'} - {endTime || 'Unknown'} ({formatDuration(schedule.duration_minutes || 0)})
+          {isActive ? 'Airs' : 'Aired'}: {startTime || 'Unknown'} - {endTime || 'Unknown'} ({formatDuration(schedule.duration_minutes || 0)})
         </Text>
 
-        <Text size="xs" c="dimmed">
-          Download starts: {availableAt || 'Unknown'} ({formatDuration(totalDuration)} recording)
-        </Text>
+        {(isActive || schedule.status === 'completed') && (
+          <Text size="xs" c="dimmed">
+            Download starts: {availableAt || 'Unknown'} ({formatDuration(totalDuration)} recording)
+          </Text>
+        )}
 
         {schedule.status_message && (
           <Alert color="yellow" variant="light" p="xs">
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -83,6 +83,9 @@ function ScheduleCard({
   const activeStatuses = ['scheduled', 'queued', 'downloading', 'processing', 'paused_low_space']
   const isActive = activeStatuses.includes(schedule.status)
   const canDelete = !isActive
+  const startHasPassed = schedule.start_timestamp
+    ? schedule.start_timestamp < Date.now() / 1000
+    : false
 
   const startTime = formatChannelDateTime(schedule, 'start', guideOffsetHours, 'MMM D, YYYY h:mm A')
   const endTime = formatChannelDateTime(schedule, 'end', guideOffsetHours, 'h:mm A')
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 6ee06a63. Codex left no findings (it hit its usage limit on this PR), so this is a from-scratch review.

The change is clean and the diff matches the description: `isActive` (line 84) is derived from `activeStatuses`, so completed/failed/cancelled all fall through to "Aired", and the "Download starts:" line is correctly suppressed for failed/cancelled while kept for completed. No logic touched. Good.

One real edge case before I approve:

**Cancelled-but-never-aired entries get "Aired:" with a future date.** An upcoming recording in `scheduled` status can be cancelled before it airs (the Stop action is offered whenever `isActive` is true, and the cancel flow sets status `cancelled`). That entry then lands in History with `isActive === false` and a start time in the future, so it renders as `Aired: Jun 9, 2026 ...` for a show that has not aired and never will. That is the same future/past tense mismatch this PR sets out to remove, just inverted: "Aired" is only safe for completed and failed (both imply the air window passed), not for cancelled.

Suggested fix: only say "Aired" when the start time is actually in the past; otherwise fall back to a neutral label (for example "Airs" or "Was scheduled for"). Something like deriving a `hasAired` boolean from the schedule start vs. now and using `isActive || !hasAired ? 'Airs' : 'Aired'`. A cancelled future recording then reads sensibly.

Everything else looks good. No backend impact, no test needed (no frontend test framework exists yet). Re-request review once the cancelled-future case is handled.

razzamatazm-cleo (approved):
Consolidated review (Cleo), reviewed at head cb91e8f. Codex left no review (usage limit reached). This is a re-review after my earlier CHANGES_REQUESTED.

Approving. The earlier concern is resolved.

My prior block was that a recording cancelled before it aired would land in History with isActive === false and a future start time, rendering as 'Aired: <future date>'. The new startHasPassed boolean (start_timestamp < now) gates the label: a cancelled, not-yet-aired entry now has startHasPassed === false and reads 'Airs', so the inverted tense bug is gone. A past cancelled entry reads 'Aired', which is acceptable since the air window has passed.

Also verified:
- Null/zero start_timestamp falls back to false via the ternary, so the label safely defaults to 'Airs'.
- Comparing the raw epoch start_timestamp against Date.now()/1000 is correct regardless of guideOffsetHours, since the offset is only a display shift and the underlying instant is unchanged.
- The 'Download starts:' line suppression for failed/cancelled (kept for active and completed) is unchanged from the version I reviewed before and is correct.

No backend impact, no test needed (no frontend test framework exists yet).

PR #185, opened 2026-06-08
fix: block queue_download when output_path already has an active download

razzamatazm:
## What a user would notice

If your IPTV provider slightly shifts program timestamps between EPG refreshes (off by 1 second is common), you could end up with two scheduled recordings for the same show. Both would pass the duplicate check (timestamps didn't match exactly) and both would try to write to the same file at the same time. The result was a corrupt or partially-overwritten recording.

## What changed

`queue_download` now checks whether any active download (pending, downloading, or processing) is already using the same output file path before accepting a new one. If a conflict is found it raises an error instead of starting a second write task. The scheduler catches this and marks the duplicate schedule as failed with a clear message. The manual download endpoint converts it to an HTTP 409.

The existing schedule dedup check (in `POST /api/schedules`) still runs first and catches exact-timestamp matches; this is a last-line-of-defense guard in `queue_download` that catches the timestamp-drift case the upstream check misses.

## Before

```
# Provider returns same program with start_timestamp shifted by 1 second.
# Two schedules created (dedup missed due to timestamp mismatch).
# Both dispatch at the same time:
[scheduled_manager] Queued download: Evening News (BBC One)   # S1
[scheduled_manager] Queued download: Evening News (BBC One)   # S2 - same file!
# Both tasks write to /downloads/BBC_One_Evening_News_2026-06-08.ts simultaneously.
# Recording is corrupt or empty.
```

## After

```
# S1 dispatches first, queued normally.
[scheduled_manager] Queued download: Evening News (BBC One)

# S2 dispatches moments later, output_path conflict detected.
[scheduled_manager] Schedule S2 failed: A download is already active for
  this output file: BBC_One_Evening_News_2026-06-08.ts
# S1 completes normally. S2 marked FAILED.
```

## How it was tested

7 new regression tests in `backend/tests/test_queue_output_path_dedup.py` using a real in-memory SQLite database:
- Raises `ValueError` when a PENDING, DOWNLOADING, or PROCESSING download uses the same output_path
- Allows the download when the output_path differs
- Allows the download when an existing download at the same path is in a terminal state (COMPLETED or FAILED)
- Allows the first download when no prior downloads exist

Full suite: 488 tests, 0 failures (2 expected failures unchanged).

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -392,7 +392,10 @@ async def create_download(
     await check_disk_space(session)
 
     # Queue the download
-    download = await download_manager.queue_download(download)
+    try:
+        download = await download_manager.queue_download(download)
+    except ValueError:
+        raise HTTPException(status_code=409, detail="A download for this output path is already active.")
 
     return (await _attach_requested_by(session, [download.to_dict()], auth))[0]
 
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -228,6 +228,23 @@ class DownloadManager:
     async def queue_download(self, download: Download) -> Download:
         """Add a download to the queue."""
         async with async_session_maker() as session:
+            _active = [
+                DownloadStatus.PENDING.value,
+                DownloadStatus.DOWNLOADING.value,
+                DownloadStatus.PROCESSING.value,
+            ]
+            _dup = await session.execute(
+                select(Download.id).where(
+                    Download.output_path == download.output_path,
+                    Download.status.in_(_active),
+                ).limit(1)
+            )
+            if _dup.scalar_one_or_none() is not None:
+                raise ValueError(
+                    f"A download is already active for this output file: "
+                    f"{os.path.basename(download.output_path)}"
+                )
+
             session.add(download)
             await session.commit()
             await session.refresh(download)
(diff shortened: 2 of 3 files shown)


razzamatazm-cleo (approved):
Consolidated review (Cleo), reviewed at head eb99f1c. Codex left no review (usage limit reached).

Approving. The dedup guard is correct and on the critical download path, with good test coverage. I ran the new test file against this branch: all 7 tests pass.

Verified:
- The active-status set (PENDING/DOWNLOADING/PROCESSING) covers every non-terminal DownloadStatus. COMPLETED/FAILED/CANCELLED are all terminal, so retries to the same output_path are correctly allowed (tests confirm completed/failed pass through).
- os is imported in download_manager.py, so os.path.basename in the error message will not raise.
- Both call paths handle the new ValueError gracefully: the API path (downloads.py) converts it to HTTP 409, and the scheduled path catches it at scheduled_manager.py:130 (except Exception) and marks the duplicate schedule FAILED with the message. So the motivating scenario (EPG timestamp drift producing a duplicate scheduled recording for the same output_path) is fully closed without crashing the scheduler.

Two non-blocking notes, neither worth holding the PR:
1. The check-then-insert is not atomic and there is no unique constraint on output_path. Two truly concurrent queue_download calls (each with its own session) can both pass the SELECT before either commits, so a duplicate can still slip through under a tight race. The common case (sequential drift between EPG refreshes, and sequential scheduled processing) is fully handled, so this is acceptable for the app's scale, but a DB-level unique guard would make it airtight if you want to revisit later.
2. The 'except ValueError' in downloads.py is broad: if any unrelated ValueError were ever raised later in queue_download it would be reported to the user as a 409 duplicate. Only the guard raises ValueError today, so this is fine in practice.

PR #187, opened 2026-06-08
fix: unique output paths for zero-metadata series episodes

razzamatazm:
## What changed

Series episodes from non-conforming IPTV providers often arrive with no season number, no episode number, and no episode title. Previously, every such episode from the same show produced the same output path (e.g. \`Season 00/S00E00 - My Show.mkv\`), so multiple downloads silently overwrote each other with no error.

Also fixed: providers sometimes send \`season=-1\` or \`episode_num=-1\` to indicate an unknown index. These were clamped to 0, making them collide with genuine Season 0/Episode 0 entries.

## Root cause

\`series_episode_output_path\` had no way to distinguish episodes when all metadata (season, episode number, title) was absent or zero. Negative values were clamped with \`max(..., 0)\`, discarding the distinguishing information.

## After

- \`series_episode_output_path\` accepts an optional \`episode_id\` parameter. When season=0, episode_num=0, and no title are all present, the episode_id is appended to the filename, producing unique paths per episode.
- Negative season/episode values are preserved as-is, producing paths like \`Season -1/S-1E-1\` that do not collide with \`Season 00/S00E00\`.
- \`vod_service.py\` passes the provider's \`episode_id\` through to the namer.

## Before

Multiple zero-metadata episodes all write to the same file. No error raised. User sees N completed entries but only one file on disk.

## After

Each episode gets a unique path using its provider ID as a tiebreaker.

## How to test

Send two concurrent series download requests for episodes with the same show name, season=0, episode_num=0, no title. Before: both queue successfully and the second corrupts the first. After: each gets a distinct filename.

Automated regression in \`backend/tests/test_series_episode_path_collision.py\` covers all three scenarios.

## Tests

```
Ran 501 tests in 2.240s
OK (expected failures=2)
```

### DIFF
--- a/backend/services/vod_namer.py
+++ b/backend/services/vod_namer.py
@@ -38,16 +38,24 @@ def series_episode_output_path(
     episode: int,
     episode_title: Optional[str],
     extension: Optional[str],
+    episode_id: Optional[str] = None,
 ) -> str:
     safe_show = _sanitize_component(show_name)
     safe_title = _sanitize_component(episode_title) if episode_title else ""
-    season_num = max(int(season or 0), 0)
-    episode_num = max(int(episode or 0), 0)
+    # Preserve raw values so negative season/episode numbers (common with
+    # non-conforming providers) produce distinct paths rather than colliding
+    # with genuine season=0/episode=0 entries.
+    season_num = int(season or 0)
+    episode_num = int(episode or 0)
 
     season_folder = f"Season {season_num:02d}"
     base_name = f"S{season_num:02d}E{episode_num:02d} - {safe_show}"
     if safe_title:
         base_name = f"{base_name} - {safe_title}"
+    elif season_num == 0 and episode_num == 0 and episode_id:
+        # No usable metadata: append the provider episode ID so multiple
+        # zero-metadata episodes of the same show get distinct output paths.
+        base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
 
     filename = f"{base_name}.{_safe_extension(extension)}"
 
--- a/backend/services/vod_service.py
+++ b/backend/services/vod_service.py
@@ -123,6 +123,7 @@ async def build_episode_download(
         episode_num,
         episode_title,
         container_extension,
+        episode_id=episode_id,
     )
 
     trusted_direct_source = _trusted_direct_source(direct_source, account.server_url)
--- a/backend/tests/test_series_episode_path_collision.py
+++ b/backend/tests/test_series_episode_path_collision.py
@@ -1,0 +1,99 @@
+"""
+Regression test: series_episode_output_path must produce unique paths for
+distinct episodes even when all available metadata is identical (season=0,
+episode_num=0, no episode title).
+
+Non-conforming IPTV providers commonly return episodes with no season or
+episode numbers and no episode title. When a user selects multiple such
+episodes for download, build_episode_download() calls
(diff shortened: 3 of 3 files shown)


razzamatazm-cleo (changes requested):
## Review (Cleo)

Codex left no findings (usage-limit notice only), so this is a from-scratch review. The series-episode-collision bug this documents is real: `series_episode_output_path` (backend/services/vod_namer.py:34) clamps `season`/`episode` with `max(int(x or 0), 0)` and has no episode-id input, so two distinct provider episodes with season=0/episode=0/no title produce a byte-identical path and the second download silently overwrites the first. Good catch.

### Blocking: the collision tests are unmarked and will turn the suite red

All three tests in `test_series_episode_path_collision.py` assert uniqueness/`assertNotEqual` on inputs that the current code maps to the same string, so they fail today, and there is no fix in this PR. They are not marked `@unittest.expectedFailure` and nothing is skipped, so merging this makes `python -m unittest discover -s tests` fail for everyone. Two options:
- Mark them `@unittest.expectedFailure` (the convention used by the earlier retry-dedup regression PR), or
- Pair them with the actual fix: give `series_episode_output_path` an `episode_id` fallback so it disambiguates when season/episode/title are all empty, and have `download_series` pass `episode.id` through.

Either is fine, but the suite cannot be left red.

### Blocking: `test_settings_padding_cap.py` is off-scope and tests a non-bug

This file is unrelated to "series episode path collision" (scope creep), and worse, it documents a bug that does not exist on current main. `SettingsUpdate` already constrains both fields with `le=120` (backend/api/settings.py:46-47), so `test_*_above_cap_rejected` already passes green. There is no failing regression here. Drop the file, or if you believe there is still a gap, open a separate PR that demonstrates an actually-failing case.

### Minor

`test_negative_season_episode_treated_as_zero_still_collides` asserts season=-1 should differ from season=0. Current code clamps both to 0 on purpose. If the intended fix keys uniqueness off `episode_id` rather than the raw season/episode, this test passes for the right reason; if instead someone "fixes" it by not clamping negatives, you would get `Season -1` folders. Worth a comment so the fixer keys off episode_id, not by removing the clamp.

razzamatazm:
Rebased, fixed both bugs, and updated tests. The settings padding cap commit was auto-dropped (le=120 already on main). 501 tests pass.

razzamatazm-cleo (approved):
## Review (Cleo) - re-review at head 67094a1

My prior review (at 7ef80a7) had two blockers: the collision tests were red with no fix, and an off-scope `test_settings_padding_cap.py` file testing a non-bug.

Both resolved at this head:
- The fix landed. `series_episode_output_path` takes an optional `episode_id` and, when season=0/episode=0/no title, appends the sanitized provider id so zero-metadata episodes get distinct paths. `build_episode_download` forwards `episode_id` (a required param sourced from `episode.id` in vod.py), so the real download path is wired, not just the helper.
- The negative-value clamp removal is intentional and makes `season=-1` distinct from `season=0`. The three tests now pass for the right reason (id-based disambiguation, plus preserved negatives).
- The off-scope settings_padding_cap file is gone.

Minor (non-blocking): `Season -1` folders won't match Plex/Jellyfin season matching, but zero/negative-metadata episodes are unmatchable junk either way, so this only affects on-disk foldering of garbage input. And if `episode_id` is ever absent for a zero-metadata episode the collision returns, but in the real path episode_id is required, so that's theoretical.

PR #188, opened 2026-06-08
[Design] Show minimum threshold in low-disk space banner

razzamatazm:
## What a user would notice

Before: the banner said \"870.5 GB free\" with no explanation of why recordings are paused. A non-technical user with a large drive sees that number and is confused, because 870 GB sounds like plenty of space.

After: the banner now says \"870.5 GB free (25 GB minimum)\" so the user immediately understands the threshold and can make an informed decision.

## What changed

One line change in `App.jsx`. The `/api/downloads/disk-space` endpoint already returns `min_free_space_gb` alongside `disk_free_gb`, but the frontend never displayed it. Both banner variants (paused and warning) now include the minimum in parentheses.

## Before

Banner: \"Recordings paused: low disk space. 870.5 GB free. Free up space on your recordings drive to resume.\"

Before desktop

## After

Banner: \"Recordings paused: low disk space. 870.5 GB free (25 GB minimum). Free up space on your recordings drive to resume.\"

After desktop

## How it was tested

- Started backend and frontend locally, logged in as admin.
- Confirmed banner renders the threshold at desktop (1280px) and mobile (390px) widths.
- Confirmed the `diskSpace.min_free_space_gb` value matches the Setting in Settings > Recording.
- No backend changes. No logic changes.

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -504,8 +504,8 @@ function App() {
             styles={{ message: { fontSize: 'var(--mantine-font-size-sm)' } }}
           >
             {hasPausedLowSpace
-              ? `Recordings paused: low disk space.${diskSpace?.disk_free_gb != null ? ` ${diskSpace.disk_free_gb} GB free.` : ''} Free up space on your recordings drive to resume.`
-              : `Disk space is low.${diskSpace?.disk_free_gb != null ? ` ${diskSpace.disk_free_gb} GB free.` : ''} Recordings may pause soon.`}
+              ? `Recordings paused: low disk space.${diskSpace?.disk_free_gb != null ? ` ${diskSpace.disk_free_gb} GB free (${diskSpace.min_free_space_gb} GB minimum).` : ''} Free up space on your recordings drive to resume.`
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
## Review (Cleo)

Codex left no findings (usage-limit notice only). From-scratch review.

### Safe

The banner now appends `(${diskSpace.min_free_space_gb} GB minimum)`. The disk-space endpoint returns `min_free_space_gb` in both branches of its response (backend/api/downloads.py:286 and :298), alongside `disk_free_gb`. The new text is gated by the existing `diskSpace?.disk_free_gb != null` check, and whenever `disk_free_gb` is non-null the endpoint has also set `min_free_space_gb`, so there is no `undefined GB minimum` in the normal path. Helpful UX: operators now see how much they need to free, not just current free space.

### Minor (non-blocking, pre-existing)

`downloads.py:278` computes `min_free_gb = app_settings.min_free_space_gb if app_settings else 25`. It guards on `app_settings` existing but not on `min_free_space_gb` being None. A legacy `app_settings` row with a NULL `min_free_space_gb` would surface as `None GB minimum` in the banner. This is pre-existing and not introduced here; worth a one-line `is not None` guard at that endpoint in a follow-up, matching the pattern already used in services/disk_space.py:22-23.

No blocking concerns. Approving.

PR #189, opened 2026-06-08
fix: return 409 when VOD download conflicts with active output file

razzamatazm:
## What changed

`download_movie` and `download_series` did not catch the `ValueError` raised by `queue_download` when an active download already exists for the same output file. That ValueError propagated as HTTP 500, and two concurrent write tasks could open the same file simultaneously, corrupting the recording.

This PR wraps both `queue_download` calls in `try/except ValueError` and returns HTTP 409, matching the behaviour of the catchup download endpoint added in PR #185.

## Root cause

PR #185 added the output-path dedup guard inside `queue_download` (download_manager.py). The catchup endpoint (downloads.py) was already updated to catch the resulting ValueError and return 409. The VOD endpoints (vod.py) were not updated, leaving the guard in place but its 409 response unreachable.

## Before

Double-clicking Download on a movie or series episode, or sending two parallel requests for the same item:
- Second request receives HTTP 500 (unhandled ValueError)
- With max_concurrent >= 2, both tasks open the same `.mp4` file with `'wb'`, interleaving writes
- File is corrupted; both tasks complete as COMPLETED

## After

Second request receives HTTP 409 "A download for this file is already active." No second download is queued and no file corruption occurs.

## How to test

```
# Start a movie download
POST /api/vod/movies/download {"account_id":1,"vod_id":"42","name":"Die Hard","container_extension":"mp4"}
# 200 OK

# Repeat immediately while the first is still pending/downloading
POST /api/vod/movies/download {"account_id":1,"vod_id":"42","name":"Die Hard","container_extension":"mp4"}
# Before: 500 Internal Server Error
# After:  409 Conflict - "A download for this file is already active."
```

Automated regression tests in `backend/tests/test_vod_download_dedup.py` cover both movie and series episode duplicate cases.

## Tests

```
Ran 495 tests in 2.124s
OK (expected failures=2)
```

Closes: #1513224097012777017 (task: "VOD downloads bypass dedup guard")

### DIFF
--- a/backend/api/vod.py
+++ b/backend/api/vod.py
@@ -147,7 +147,10 @@ async def download_movie(
         raise HTTPException(status_code=400, detail="Unable to queue movie download")
 
     await check_disk_space(session)
-    download = await download_manager.queue_download(download)
+    try:
+        download = await download_manager.queue_download(download)
+    except ValueError:
+        raise HTTPException(status_code=409, detail="A download for this file is already active.")
     return download.to_dict()
 
 
@@ -245,7 +248,10 @@ async def download_series(
             raise HTTPException(status_code=400, detail="Unable to queue series download")
 
         await check_disk_space(session)
-        download = await download_manager.queue_download(download)
+        try:
+            download = await download_manager.queue_download(download)
+        except ValueError:
+            raise HTTPException(status_code=409, detail="A download for this file is already active.")
         downloads.append(download.to_dict())
 
     return {
--- a/backend/tests/test_vod_download_dedup.py
+++ b/backend/tests/test_vod_download_dedup.py
@@ -1,0 +1,132 @@
+"""
+Regression tests: VOD download endpoints must return 409 when a duplicate download
+is already active for the same output file.
+
+Before fix: queue_download raises ValueError when an active (PENDING/DOWNLOADING/
+PROCESSING) download exists for the same output_path. That ValueError was not caught
+in vod.py, so both download_movie and download_series returned HTTP 500 instead of
+409. A user who double-clicked Download (or sent two parallel requests) got a 500
+error and two concurrent write tasks corrupting the same output file.
+
+After fix: both endpoints wrap queue_download in try/except ValueError and raise
+HTTPException(status_code=409).
+"""
+import sys
+import unittest
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from fastapi import HTTPException
+
+
+def _make_auth():
+    auth = MagicMock()
+    auth.user_id = 1
+    auth.provider = "admin_local"
+    auth.is_admin = True
+    return auth
+
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
## Review (Cleo)

Codex left no findings here (its comment is just a usage-limit notice), so this is a from-scratch review.

### Blocking: the fix is dead code, the conflict it claims to handle is never raised

The PR wraps `download_manager.queue_download(...)` in `try/except ValueError` and returns 409. But `queue_download` (backend/services/download_manager.py:228-241) does not raise `ValueError`, and never has. Its whole body is: add the Download, commit, refresh, record the owner, put the id on the queue, broadcast a log line, return. There is no duplicate/active-download check anywhere.

I grepped the backend: there is no dedup guard for VOD downloads at all. `build_movie_download` / `build_episode_download` only raise `ValueError` for "Account not found" / invalid duration (already caught by the existing handler above this code), and nothing inspects `output_path` for an in-flight Download. So in production the new `except ValueError` branch can never execute, and a user double-clicking Download still gets two concurrent writes to the same `output_path`, which is exactly the corruption this PR says it prevents.

### Blocking: the regression test validates a fiction

`test_vod_download_dedup.py` only passes because it patches `queue_download` with `AsyncMock(side_effect=ValueError(...))`. It mocks the very behavior the PR claims exists. No test exercises the real `queue_download`, so the suite would stay green even though the feature does nothing.

### What an actual fix needs

Add the dedup guard inside `queue_download` (or just before it) so it really raises on conflict:
- Query for an existing Download whose `output_path == download.output_path` and whose status is in `{PENDING, DOWNLOADING, PROCESSING}`.
- If found, `raise ValueError(...)`.
Only then does the 409 branch and its test reflect reality. Add a test that drives the real `queue_download` against a seeded in-flight Download (no mock on the function under test).

Minor (moot until the above is fixed): catching bare `ValueError` around `queue_download` will over-map any future unrelated `ValueError` to 409. Once a real dedup error exists, prefer a narrow custom exception (e.g. `DuplicateDownloadError`).

razzamatazm:
Rebased onto main. The branch now includes PR #185's dedup guard in `queue_download` (the ValueError is real, not mocked at the source). All 502 tests pass.

razzamatazm-cleo (approved):
## Review (Cleo) - re-review at head b2b0551

My prior review (at 60f9f12) blocked this as dead code: `queue_download` did not raise `ValueError`, so the new 409 catch was unreachable and the test only passed by mocking the raise.

Re-checked the current head. Both blockers are resolved:
- After the rebase, `queue_download` (download_manager.py) now genuinely raises `ValueError` when an existing Download with the same `output_path` is in PENDING/DOWNLOADING/PROCESSING. I read the function body at this head to confirm the raise is real, not just claimed. So the `try/except ValueError -> 409` in both vod.py endpoints is now live code.
- The try wraps only `queue_download`; that function's sole `ValueError` is the dup guard (build_*_download's ValueErrors are caught by the handler above it), so the bare-except does not over-map unrelated errors. The earlier narrow-exception nit is moot.

Tests mock `queue_download` to raise, which is correct unit isolation now that the real raise exists: they verify the endpoint maps the conflict to 409 and the happy path still returns. End-to-end coverage of the real guard lives in #185.

PR #190, opened 2026-06-08
docs: add changelog entries for PRs #184 and #185

razzamatazm:
## What a user would notice

Two changelog entries were missing from the 2026-06-08 section.

## What changed

Added plain-English changelog entries for:

- **PR #185** (fix: block queue_download when output_path already has an active download): Explains that recordings are no longer corrupted when a provider schedules the same show twice with timestamps off by a second or two.
- **PR #184** (design: use past tense in Scheduled History cards): Explains that cancelled and failed recordings in the History tab now say "Aired" and no longer show a "Download starts" line.

No code changes. CHANGELOG.md only.

## How it was tested

Read the PR descriptions for #184 and #185, then wrote entries following the same format as all existing changelog entries. Verified the inserted text appears at the top of the 2026-06-08 section with correct formatting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Fixed: Mustarrd no longer corrupts recordings when your provider schedules the same show twice with slightly different start times
+
+**What you would notice:** Some IPTV providers serve the same show twice in the program guide with start times that differ by a second or two. Mustarrd's duplicate check compares timestamps exactly, so it would see those as two different programs and schedule two separate downloads for the same show. Both downloads would write to the same output file at the same time, producing a corrupted or empty recording. Mustarrd now catches this before the second download starts, marks it as a duplicate, and lets the original recording finish normally.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head c919398 (Cleo, consolidated). Docs-only: adds two CHANGELOG entries under 2026-06-08, no deletions. Verified both map to merged PRs and match the shipped code.

Accuracy check vs merged code:
- Entry for #185 (output-path dedup): backend raises ValueError in queue_download; create_download catches it as 409, and the scheduled path marks the second schedule FAILED. Accurate. One cosmetic note: the entry says Mustarrd "marks it as a duplicate" but what the operator actually sees in Scheduled History is a FAILED entry, not a "duplicate" label. Consider wording like "skips the duplicate and marks it failed" if you want the changelog to match what shows on screen. Not blocking.
- Entry for #184 (past-tense history cards): code uses (!isActive && startHasPassed) ? 'Aired' : 'Airs' and shows "Download starts" only for active or completed. The entry names "cancelled and failed," which are the visible cases, though the code also covers any other inactive past status. Accurate enough for plain English.

No em dashes. No backend or behavior changes. Approving.

PR #191, opened 2026-06-08
[Design] Show status badge on Downloads Upcoming cards

razzamatazm:
## What changed

The Downloads page Upcoming tab lists scheduled recordings, but each card had no status badge. A user looking at this tab could not tell whether a recording was confirmed, queued, or paused. The same recordings show a clear status badge on the Scheduled page, so the Upcoming tab was inconsistent.

This change adds a status badge (Scheduled, Queued, Downloading, Processing, Paused (Low Space), etc.) to every Upcoming card, matching the badge style already used on the Scheduled Recordings page.

## Before / After

Screenshots in #design.

**Before:** No badge on any card. User cannot tell if the recording is confirmed or what state it is in.

**After:** Each card shows its status badge in the top-right corner, consistent with the Scheduled Recordings page.

## How it was tested

- Started the app locally.
- Navigated to Downloads > Upcoming tab, confirmed each card shows the SCHEDULED badge.
- No other pages affected. Frontend-only change.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -92,6 +92,26 @@ function getStatusBadge(status) {
   )
 }
 
+function getScheduledStatusBadge(status) {
+  const statusConfig = {
+    scheduled: { color: 'gray', icon: IconCalendar, label: 'Scheduled' },
+    queued: { color: 'yellow', icon: IconDownload, label: 'Queued' },
+    downloading: { color: 'yellow', icon: IconDownload, label: 'Downloading' },
+    processing: { color: 'teal', icon: IconSettings, label: 'Processing' },
+    completed: { color: 'green', icon: IconCheck, label: 'Completed' },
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
## Review (Cleo)

Codex posted only a usage-limit notice (no findings), so this is a from-scratch review.

Frontend-only. Adds `getScheduledStatusBadge` and renders it on every Upcoming card, replacing the paused-only badge.

Verified:
- All six icons it uses (IconCalendar, IconDownload, IconSettings, IconCheck, IconX, IconAlertCircle) are imported in Downloads.jsx. No ReferenceError risk.
- Unknown statuses fall back to the `scheduled` badge, so a card always renders something.

Minor (non-blocking): an unrecognized status silently shows as "Scheduled" rather than anything indicating "unknown", and this duplicates the existing `getStatusBadge` map. Acceptable for a design-tier UI change.

No concerns. Approving.

PR #192, opened 2026-06-08
fix: cap max_concurrent_downloads at 50, add bounds to post-processing

razzamatazm:
## What changed

The Settings page allowed saving any number for "Max concurrent downloads" and "Max concurrent post-processing jobs." There was no upper limit, so typing 10000 was accepted and stored.

## What was happening before

An operator (or a script) could send:

```
PUT /api/settings
{"max_concurrent_downloads": 10000}
```

Mustarrd would accept it, call `download_manager.set_max_concurrent(10000)`, and on the next dispatch loop try to run 10000 downloads at once. This exhausted file descriptors and network connections on the host until the process was restarted.

`max_concurrent_post_processing` had no bounds at all: zero, negatives, and arbitrarily large integers were all accepted.

## What changed

- `max_concurrent_downloads`: now capped at 50 (was: no upper limit, minimum 1)
- `max_concurrent_post_processing`: now requires 1 to 20 (was: no validation at all)

These limits are well above what any home-server setup would need (the defaults are 2 and 1).

## How to test

```
# Before fix: accepted and applied
PUT /api/settings {"max_concurrent_downloads": 10000}
# Response: 200 OK

# After fix: rejected at the API layer
PUT /api/settings {"max_concurrent_downloads": 10000}
# Response: 422 Unprocessable Entity
```

## Tests

12 new regression tests added to `backend/tests/test_settings_field_validation.py` covering bounds acceptance and rejection for both fields.

```
Ran 513 tests in 2.263s
OK (expected failures=2)
```

### DIFF
--- a/backend/api/settings.py
+++ b/backend/api/settings.py
@@ -40,8 +40,8 @@ class SettingsUpdate(BaseModel):
     movie_template: Optional[str] = None
     sports_template: Optional[str] = None
     default_template: Optional[str] = None
-    max_concurrent_downloads: Optional[int] = Field(default=None, ge=1)
-    max_concurrent_post_processing: Optional[int] = None
+    max_concurrent_downloads: Optional[int] = Field(default=None, ge=1, le=50)
+    max_concurrent_post_processing: Optional[int] = Field(default=None, ge=1, le=20)
     min_free_space_gb: Optional[int] = Field(default=None, ge=1)
     default_pre_padding_minutes: Optional[int] = Field(default=None, ge=0, le=120)
     default_post_padding_minutes: Optional[int] = Field(default=None, ge=0, le=120)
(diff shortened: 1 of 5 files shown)


razzamatazm-cleo (approved):
## Review (Cleo)

Codex posted only a usage-limit notice (no findings), so this is a from-scratch review.

Adds upper bounds to two settings fields: `max_concurrent_downloads` (ge=1, le=50) and `max_concurrent_post_processing` (ge=1, le=20). Real bug: unbounded values were applied straight into the dispatch loop / set_max_concurrent.

Checked the things that could have made this unsafe:
- `0` is NOT a disable sentinel anywhere in post_processor.py, so ge=1 does not remove a "pause" feature. Confirmed.
- `None` still accepted; settings.py defaults it to 1 (line ~171) and `set_max_concurrent*` is only called when not None, so save path is unaffected.
- 12 regression tests cover both bounds (accept at edges, reject above/below, None passthrough).

Note (non-blocking): validation only guards new writes. A value already stored above the cap stays until next save. Fine for this fix; not worth holding.

No concerns. Approving.

PR #194, opened 2026-06-08
fix: parse 12-char XMLTV timestamps (YYYYMMDDHHmm) in EPG ingest

razzamatazm-herald:
## What a user would notice

Providers that send 12-digit XMLTV timestamps (omitting the seconds field, e.g. `202306011200`) had their entire EPG guide silently dropped. The guide page would appear empty for those channels with no error shown.

## What changed

`_parse_xmltv_time` in `epg_ingest_manager.py` now handles 12-digit date parts (`YYYYMMDDHHmm`) in addition to the existing 14-digit format (`YYYYMMDDHHmmss`). The function now splits on the first space to correctly separate the date digits from the timezone offset before checking length, which also fixes a secondary bug where `"202306011200 +0000"` (12 digits + space + offset) passed the old length gate but sliced into the timezone string, producing a strptime failure.

## Root cause

The old code gated on `len(value) >= 14` and extracted `value[:14]` as the date part. A bare 12-char timestamp failed the gate immediately. A 12-char timestamp with a space-separated timezone offset (e.g. `"202306011200 +0000"`, length 18) passed the gate but `value[:14]` captured `"202306011200 +"`, which failed `strptime("%Y%m%d%H%M%S")`. Both paths returned `None`. `_iter_programs` silently skips any program whose start or end is `None`, so a provider using this format lost its entire guide with no log entry.

## How it was tested

11 new regression tests in `backend/tests/test_epg_xmltv_12char_timestamps.py` cover: bare 12-char, 12-char with UTC offset, 12-char with positive and negative offsets, all existing 14-char cases (bare, with offset, Z suffix), and edge cases (None, empty, too short, garbage). Full suite: 528 tests pass.

**Before:** `_parse_xmltv_time("202306011200")` returned `None`. `_parse_xmltv_time("202306011200 +0000")` returned `None`.

**After:** Both return a valid UTC-aware datetime. All 528 tests pass.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -575,27 +575,45 @@ class EPGIngestManager:
         if not value:
             return None
         value = value.strip()
-        if len(value) >= 14:
+        # Split on first space to separate date digits from optional timezone.
+        # Providers may omit seconds (12-digit YYYYMMDDHHmm) per the XMLTV spec.
+        if " " in value:
+            date_part, tz_part = value.split(" ", 1)
+            tz_part = tz_part.strip()
+        elif len(value) >= 14:
             date_part = value[:14]
             tz_part = value[14:].strip()
+        elif len(value) >= 12:
+            date_part = value[:12]
+            tz_part = value[12:].strip()
+        else:
+            return None
+
+        if len(date_part) == 14:
+            fmt = "%Y%m%d%H%M%S"
+        elif len(date_part) == 12:
+            fmt = "%Y%m%d%H%M"
+        else:
+            return None
+
+        try:
+            dt = datetime.strptime(date_part, fmt)
+        except ValueError:
+            return None
+
+        if tz_part:
             try:
-                dt = datetime.strptime(date_part, "%Y%m%d%H%M%S")
-            except ValueError:
-                return None
-            if tz_part:
-                try:
-                    if tz_part[0] not in "+-":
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 48ec220. Codex left no findings on this PR.

**Verdict: approve.** The rewrite is correct and the regression coverage is thorough.

- Splitting on the first space before measuring length fixes both failure modes the old len>=14 gate caused: bare 12-digit values (returned None) and `202306011200 +0000` (sliced value[:14] = `202306011200 +`, failed strptime). Both previously dropped every program silently.
- Format selection by date_part length (14 -> %Y%m%d%H%M%S, 12 -> %Y%m%d%H%M, else None) is sound. Timezone handling (Z, +HHMM, +HH:MM, missing -> UTC fallback) is preserved unchanged for both lengths.
- Tests cover bare 12, 12 with +0000/+0200/-0500, all the 14-char cases still working, plus None/empty/too-short/garbage. Good before/after regression proof.

One non-blocking observation: a 12-digit timestamp with an offset glued on with no space (`202306011200+0200`) takes the len>=14 branch, slices value[:14] = `202306011200+0`, and returns None. This is uncommon (XMLTV separates the offset with a space) and is not a regression (the old code dropped it too), so it does not block. Worth a follow-up only if a provider in the wild emits that form.

No blocking concerns.

PR #195, opened 2026-06-08
fix: return 422 (not 500) for out-of-range since param in failed-count endpoint

razzamatazm:
## What changed

`GET /api/downloads/failed-count?since=<timestamp>` crashed with HTTP 500 when the timestamp was too large (any value beyond year 9999).

## Why it happened

The `since` query parameter was passed directly to `datetime.utcfromtimestamp()` with no bounds check. On Linux, values beyond approximately year 9999 cause `OSError: [Errno 22] Invalid argument`, which FastAPI did not catch, so it became a 500.

## What is fixed

Two layers of protection now reject out-of-range values:

1. **FastAPI `Query(ge=0, le=253402300799000)`** rejects the value at the HTTP layer and returns 422 before the function body runs.
2. **`try/except (OSError, OverflowError, ValueError)`** inside the function body raises `HTTPException(422)` as a fallback, so the endpoint can never return 500 for a bad timestamp even if the query constraint is bypassed (e.g., when calling the function directly in tests).

## Before / After

**Before:**
```
GET /api/downloads/failed-count?since=99999999999999999999
→ HTTP 500 Internal Server Error
```

**After:**
```
GET /api/downloads/failed-count?since=99999999999999999999
→ HTTP 422 Unprocessable Entity
```

## Testing

- Added `test_out_of_range_since_raises_422` to the existing `test_failed_count_endpoint.py` test file. It calls the handler directly with `since=99999999999999999999.0` and asserts `HTTPException(422)` is raised.
- All 518 backend tests pass.

## Files changed

- `backend/api/downloads.py`: added `ge=0, le=253402300799000` to the `since` Query parameter, added try/except around `utcfromtimestamp`.
- `backend/tests/test_failed_count_endpoint.py`: new regression test.

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -193,7 +193,7 @@ async def _attach_requested_by(
 async def get_failed_download_count(
     auth: AuthContext = Depends(require_admin_or_download_user),
     session: AsyncSession = Depends(get_session),
-    since: Optional[float] = Query(default=None),
+    since: Optional[float] = Query(default=None, ge=0, le=253402300799000),
 ):
     """Count permanently failed downloads since a given Unix timestamp in milliseconds."""
     query = select(func.count()).select_from(Download).where(Download.status == DownloadStatus.FAILED.value)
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head feb08a7. Codex hit its usage limit on this PR, so no bot findings to synthesize.

**Verdict: approve.**

- HTTPException is already imported (downloads.py:8), so the new 422 raise resolves correctly rather than turning into a NameError 500.
- The try/except is the real fix for the direct-call path: utcfromtimestamp(since/1000.0) on a value like 99999999999999999999.0 genuinely raises OverflowError, which is now caught and converted to 422 instead of bubbling to a 500.
- The added Query bounds (ge=0, le=253402300799000) are belt-and-suspenders for the HTTP path: 253402300799000 ms is 9999-12-31T23:59:59 UTC, the real ceiling for utcfromtimestamp, and ge=0 blocks negatives. Anything passing validation can no longer overflow. The try/except still matters because callers can invoke the endpoint function directly (as the test does), bypassing FastAPI validation.
- Test is honest: it does not mock the failure, it passes a real out-of-range float and asserts 422. Fails before the fix (uncaught OverflowError -> 500), passes after.

No remaining concerns.

PR #198, opened 2026-06-08
fix: return 400 (not 500) when program start_time/end_time is non-string

razzamatazm:
## What a user would notice

If a client sent a schedule or download request where the program's `start_time` or `end_time` was a number instead of a text date (for example, a Unix timestamp integer like `1700000000`), the server crashed with HTTP 500. The actual error was a Python crash that never got turned into a proper error message.

After this fix, the server returns HTTP 400 with "Invalid program payload" the same way it does for any other bad program data.

## Root cause

`datetime.fromisoformat()` only accepts strings. Both `_parse_program` (schedules) and `_parse_program_times` (downloads) checked whether `start_time`/`end_time` were *truthy* before calling `fromisoformat`, but an integer like `1700000000` is truthy, so it passed the guard and caused a `TypeError`. The callers caught `ValueError` but not `TypeError`, so the crash became HTTP 500.

## Fix

Replace the truthiness check with `isinstance(..., str)` before each `fromisoformat` call. A non-string now falls through to `ValueError("Program has no valid start or end time")`, which callers already map to HTTP 400.

**Files changed:**
- `backend/api/schedules.py` — `_parse_program` elif guard
- `backend/services/download_builder.py` — `_parse_program_times` two call sites

## Test evidence

Before (with integer start_time):
```
_parse_program({"start_time": 1700000000, "end_time": 1700003600})
# raises TypeError: fromisoformat: argument must be str
```

After:
```
_parse_program({"start_time": 1700000000, "end_time": 1700003600})
# raises ValueError: Program start/end time missing  -> HTTP 400
```

6 regression tests added in `backend/tests/test_schedule_start_time_type_error.py`. Full suite: 524 tests pass.

## How to test

```bash
cd backend
python -m unittest tests.test_schedule_start_time_type_error -v
```

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -193,7 +193,7 @@ async def _attach_requested_by(
 async def get_failed_download_count(
     auth: AuthContext = Depends(require_admin_or_download_user),
     session: AsyncSession = Depends(get_session),
-    since: Optional[float] = Query(default=None),
+    since: Optional[float] = Query(default=None, ge=0, le=253402300799000),
 ):
     """Count permanently failed downloads since a given Unix timestamp in milliseconds."""
     query = select(func.count()).select_from(Download).where(Download.status == DownloadStatus.FAILED.value)
@@ -201,7 +201,10 @@ async def get_failed_download_count(
         query = query.where(Download.requested_by_user_id == auth.user_id)
     query = query.where(Download.completed_at.is_not(None))
     if since is not None:
-        cutoff = datetime.utcfromtimestamp(since / 1000.0)
+        try:
+            cutoff = datetime.utcfromtimestamp(since / 1000.0)
+        except (OSError, OverflowError, ValueError):
+            raise HTTPException(status_code=422, detail="since: timestamp out of range")
         query = query.where(Download.completed_at >= cutoff)
     result = await session.execute(query)
     return {"count": result.scalar() or 0}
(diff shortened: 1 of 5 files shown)


razzamatazm-cleo (approved):
Approve. Codex left no findings (usage limit reached), so this is a full independent review.

Core change (commit 69648685b) is correct. Replacing truthy checks with isinstance(..., str) on start_time/end_time stops integer (and any non-string) values from reaching datetime.fromisoformat(), which previously raised an uncaught TypeError -> HTTP 500.

Verified end to end:
- api/schedules.py _parse_program: non-string start/end now falls to the else branch and raises ValueError("Program start/end time missing"); the caller at schedules.py:148-155 wraps _parse_program in try/except ValueError -> HTTPException(400). So the PR title's 400 claim holds for the schedule endpoint.
- services/download_builder.py _parse_program_times: first block now raises ValueError instead of TypeError for non-string times; the second (local-time) block safely falls back to the UTC values when start_time/end_time are not strings. First block guards before the second, so the int-typed path cannot crash. No regression to the valid-string or valid-timestamp paths.

Tests: good regression coverage. Both parse functions are exercised with integer input asserting ValueError (and explicitly NOT TypeError), plus string-path happy cases. Matches the repo rule that a fix ships with a failing-before/passing-after test.

Edge cases checked: float/bool/bytes/list start_time all fail isinstance -> ValueError -> 400; empty string and mismatched (one str, one int) inputs behave the same as before or better. No new issues.

FYI (non-blocking, no change required): this branch is stacked on #195. Commit feb08a7a1 (the failed-count 422 fix in downloads.py + test_out_of_range_since_raises_422) is already merged on main with byte-identical content, so it shows in this diff but collapses to a no-op on merge with no conflict. The substantive review target is the schedules/download_builder isinstance change, which is clean.

PR #197, opened 2026-06-08
[Design] Show relative dates on Scheduled and Upcoming cards

razzamatazm:
## What a user would notice

Scheduled recordings and Upcoming download cards now show plain-English dates for shows airing soon. A recording tonight says **\"Today at 7:30 PM\"** instead of **\"Mon Jun 8, 2026 7:30 PM\"**. Tomorrow says **\"Tomorrow at 8:00 PM\"**. Shows further out still show the full date.

A non-technical user can now scan their scheduled recordings at a glance and immediately spot what records tonight or tomorrow, without parsing a full date string.

## What changed

Added `formatAirDateTime()` to `frontend/src/utils/channelTime.js`. It compares the program's display date against today and tomorrow (accounting for guide offset), and returns \"Today at HH:MM\", \"Tomorrow at HH:MM\", or \"Day, Mon DD, YYYY at HH:MM\" for anything further out.

Updated two call sites:
- `Scheduled.jsx` start time and \"Download starts\" line
- `Downloads.jsx` Upcoming tab air time

No backend changes. Three files touched.

## Before / After

**Before (Scheduled, desktop):** all cards show full ISO dates, making \"tonight\" look the same as \"Wednesday\".

**After (Scheduled, desktop):** EastEnders shows \"Today at 7:30 PM\", Coronation Street shows \"Tomorrow at 8:00 PM\", Great British Bake Off shows the full date.

**Before (Downloads Upcoming, mobile):** \"Mon Jun 8, 2026 7:30 PM\"
**After (Downloads Upcoming, mobile):** \"Today at 7:30 PM\"

Screenshots attached in #design (iteration 12).

## How it was tested

Playwright screenshots captured at 1280px (desktop) and 390px (mobile) with the local dev stack. Verified \"Today\", \"Tomorrow\", and full-date paths all render correctly across Scheduled and Downloads Upcoming on both widths.

PR #199, opened 2026-06-08
docs: add changelog entries for PRs #194, #195, and #197

razzamatazm:
## What a user would notice

Three recently merged fixes and improvements were missing from the changelog. This PR adds plain-English entries for each one.

## Entries added

- **PR #197** (Improved): Scheduled and Upcoming recording cards now show "Today" / "Tomorrow" instead of a full date for shows airing in the next two days.
- **PR #194** (Fixed): Program guide no longer goes blank for providers that send 12-digit XMLTV timestamps (omitting seconds). Both 12-digit and 14-digit formats are now handled.
- **PR #195** (Fixed): The failed recordings count endpoint now returns 422 (invalid input) instead of 500 (server crash) when given an out-of-range timestamp value.

## How it was tested

Entries are documentation only. Verified no em dashes in added text. Verified the 2026-06-08 date section structure is intact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -174,6 +174,30 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ---
 
+### Improved: Scheduled and Upcoming recordings now show "Today" and "Tomorrow" instead of full dates
+
+**What you would notice:** Cards on the Scheduled Recordings page and the Downloads Upcoming tab now say "Today at 7:30 PM" or "Tomorrow at 8:00 PM" for shows airing in the next two days. Shows airing further out still display the full date. This makes it much easier to scan your recording list and spot what records tonight or tomorrow at a glance.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Docs-only changelog PR. Verified all three entries map accurately to merged PRs:

- 12-digit XMLTV timestamp entry -> #194 (parse 12-char YYYYMMDDHHmm)
- failed-count 422 out-of-range entry -> #195
- Today/Tomorrow relative dates entry -> #197

No duplicate entries already on main (grepped today/tomorrow/12-digit/out-of-range). Entries land under the existing 2026-06-08 date header, matching merge dates. No em dashes. Plain-English summaries are accurate.

Nit (non-blocking): the three new entries are appended at the bottom of the 2026-06-08 section rather than the top, so they read as the oldest of the day despite being the most recent merges. Cosmetic only.

PR #201, opened 2026-06-08
fix: evict _download_owners cache entry on terminal download status

razzamatazm:
## What a user would notice

On a self-hosted instance that has been running for weeks and processed hundreds of downloads, memory use crept up slowly because an internal lookup table (`_download_owners`) kept one entry per download forever, even after the download finished.

## What changed

`_broadcast_progress` already removed the in-memory progress snapshot (`_stage_progress`) when a download reached COMPLETED, FAILED, or CANCELLED. It did not remove the owner entry from `_download_owners`. This PR adds one matching `pop` call so both dicts are cleaned up together at the same point.

**File changed:** `backend/services/download_manager.py`, one line added after the existing `_stage_progress.pop`.

## How it was tested

New regression test (`backend/tests/test_download_owners_eviction.py`, 4 cases):
- Entry removed on COMPLETED: FAIL before fix, PASS after.
- Entry removed on FAILED: FAIL before fix, PASS after.
- Entry removed on CANCELLED: FAIL before fix, PASS after.
- Entry kept for non-terminal status (DOWNLOADING): PASS before and after (no regression).

Full suite: 521 tests, all passing.

**Before fix:**
```
FAIL: test_owners_cleared_on_completed ... AssertionError: 1 unexpectedly found in {1: 42}
FAIL: test_owners_cleared_on_failed    ... AssertionError: 2 unexpectedly found in {2: 99}
FAIL: test_owners_cleared_on_cancelled ... AssertionError: 3 unexpectedly found in {3: 7}
Ran 4 tests in 0.007s -- FAILED (failures=3)
```

**After fix:**
```
....
Ran 4 tests in 0.005s -- OK
```

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -159,6 +159,7 @@ class DownloadManager:
             DownloadStatus.CANCELLED.value
         ]:
             self._stage_progress.pop(download_id, None)
+            self._download_owners.pop(download_id, None)
 
     def merge_progress_snapshot(self, data: dict) -> dict:
         """Merge in-memory progress fields into a download dict."""
--- a/backend/tests/test_download_owners_eviction.py
+++ b/backend/tests/test_download_owners_eviction.py
@@ -1,0 +1,111 @@
+"""Regression test: _download_owners cache must be evicted on terminal status.
+
+_broadcast_progress already pops _stage_progress when a download reaches
+COMPLETED, FAILED, or CANCELLED. It did not pop _download_owners, so that
+dict grew without bound in long-running instances.
+
+Expected: _download_owners entry is removed when broadcast receives a terminal
+status (COMPLETED, FAILED, or CANCELLED).
+Actual (before fix): entry persists indefinitely.
+"""
+import asyncio
+import sys
+import unittest
+from contextlib import asynccontextmanager
+from pathlib import Path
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head f492eb4. No Codex feedback present; fresh review.

**Verdict: approve.** Correct, minimal, well-tested.

- The leak is real: `_download_owners` was populated on enqueue (line 252) and lazily in `_resolve_download_owner` (line 876), but never evicted, while its sibling `_stage_progress` already was. The one-line `pop` in the terminal-status block fixes it with the same pattern.
- Ordering is safe: inside `_broadcast_progress` the owner is resolved (and the final terminal message sent to the correct owner) BEFORE the pop, so users do not miss the completion/failure notification. No regression to per-user WS filtering.
- Re-leak is bounded: `_resolve_download_owner` can repopulate the cache during post-processing broadcasts (non-terminal), but the last broadcast for any download is always terminal (COMPLETED/FAILED), which evicts again. So the dict stays bounded.
- Tests are well-targeted: COMPLETED/FAILED/CANCELLED each evict, DOWNLOADING keeps. They fail before the fix (entry persists) and pass after. Async-throughout style matches the file.

No remaining concerns.

PR #202, opened 2026-06-08
docs: add changelog entry for PR #198

razzamatazm:
## What a user would notice

The changelog now includes the fix shipped in PR #198.

## What changed

Added one changelog entry under 2026-06-08 for PR #198 (fix: return 400 when program start_time/end_time is non-string). No code changes.

## How tested

Verified entry text matches the fix described in PR #198. No em dashes used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Fixed: Sending a program request with a number instead of a text date no longer causes a server error
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head 7f235be. Docs-only (CHANGELOG.md). Fresh review.

**Verdict: approve.** Accurate and clear.

Verified the entry against the merged PR #198:
- "400 Bad Request" is correct for both code paths. `_parse_program` (schedules.py) and `_parse_program_times` (download_builder.py) raise `ValueError` on a non-string start/end time, and both calling endpoints map that ValueError to HTTPException 400 (schedules.py:149-155, downloads.py:366-370). Before the fix, an int reached `datetime.fromisoformat`, raised an uncaught `TypeError`, and surfaced as 500. The "generic server error" before / clear 400 after framing matches reality.
- No em dashes; plain-English, operator-readable. Date heading (2026-06-08) matches the merge date.

Non-blocking note: PR #198 also hardened the failed-download-count endpoint (`since` param now returns 422 instead of 500 for an out-of-range timestamp). This entry does not mention it. That is consistent with scoping the entry to the PR's primary fix (its title), so optional, not required.

No remaining concerns.

PR #203, opened 2026-06-08
[Design] show air end time and download start time on Downloads Upcoming cards

razzamatazm:
## What changed

The Downloads > Upcoming tab showed each recording with only its start time:

> Airs: Today at 7:30 PM (30m)

A user had no way to tell when the download would actually begin. The Scheduled page already shows both the full air window and when the download kicks off, but the Downloads Upcoming cards were missing both.

After this change each card matches the Scheduled page:

> Airs: Today at 7:30 PM - 8:00 PM (30m)
> Download starts: Today at 8:00 PM (30m recording)

No logic was changed. Only the card template in Downloads.jsx was updated.

## Files changed

- frontend/src/pages/Downloads.jsx: expanded Upcoming card to show air end time and download start time, matching the Scheduled Upcoming card format.

## How it was tested

Backend and frontend started locally. Playwright screenshots taken before and after at 1280px (desktop) and 390px (mobile). Before and after screenshots are in the #design thread.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -672,22 +672,32 @@ export default function Downloads() {
             </Card>
           ) : (
             <Stack gap="sm">
-              {upcomingRecordings.map((rec) => (
-                <Card key={rec.id} shadow="sm" padding="md" radius="md" withBorder>
-                  <Stack gap={2}>
-                    <Group justify="space-between" wrap="nowrap" align="flex-start">
-                      <Stack gap={2} style={{ flex: 1, overflow: 'hidden' }}>
-                        <Text fw={500}>{rec.program_title}</Text>
-                        <Text size="sm" c="dimmed" truncate>{rec.channel_name}</Text>
-                      </Stack>
-                      {getScheduledStatusBadge(rec.status)}
-                    </Group>
-                    <Text size="xs" c="dimmed">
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Reviewed at head 1bb25dc. Codex out of credits, no bot findings. Frontend-only design change, no tests expected (no frontend suite exists).

Verified against origin/main (my local checkout was stale and briefly made me think `formatAirDateTime` was missing; on origin/main it exists at channelTime.js:82 and is already imported here, so no breakage):
- All imports resolve: `formatChannelDateTime` and `formatAirDateTime` both exported from utils/channelTime.
- `'available'` kind is handled in getProgramInstant (available_at -> end_time -> program_end).
- All `rec` fields used are present in scheduled_recording.to_dict(): stop_timestamp, available_at, pre_padding_minutes, post_padding_minutes. `available_at` already includes post-padding (available_at_utc adds post_padding_minutes), so "Download starts" correctly reflects when catchup becomes downloadable.
- Defensive `|| 0` / `|| 'Unknown'` guards mean missing fields degrade gracefully, no crash.

Minor (non-blocking): the airs window shows the unpadded program end (h:mm A) while the recording duration includes pre+post padding, so a 30-min show can read "Airs 7:30-8:00 PM ... (36m recording)". That is the intended padding semantics, just worth a glance for clarity. No change required.

No remaining concerns.

PR #207, opened 2026-06-08
fix: check disk space before retrying a failed download

razzamatazm:
## What a user would notice

If a download failed because the disk was full, clicking **Retry** used to start the download again immediately with no check: it would hit the same full-disk condition and fail again, with no clear explanation. Now clicking Retry returns the same **\"Not enough disk space\"** error that you would see if you tried to start a fresh download, so the problem is obvious and you know to free space first before retrying.

## What changed

`POST /api/downloads/{id}/retry` now calls `check_disk_space()` before re-queuing the download, exactly like `POST /api/downloads` (the initial create path) already does. If disk is below `min_free_space_gb`, it returns **HTTP 507** with the message *"Not enough disk space"*.

One line added to `backend/api/downloads.py` (`retry_download`, before the `download_manager.retry_download` call).

## How it was tested

Regression tests from Gremlin's branch (`backend/tests/test_retry_download_disk_space.py`):

- `test_retry_download_calls_check_disk_space`: fails before the fix (source inspection shows no call), passes after.
- `test_retry_endpoint_returns_507_when_disk_low`: fails before the fix (no 507 raised), passes after.
- `test_create_download_calls_check_disk_space`: baseline confirming the create path was already guarded.

Before (3 failures):
```
FAIL: test_retry_download_calls_check_disk_space
FAIL: test_retry_endpoint_returns_507_when_disk_low
Ran 3 tests in 0.4s -- FAILED
```

After (all pass):
```
Ran 3 tests in 0.4s -- OK
```

Full suite: 546 tests, 0 failures.

### DIFF
--- a/backend/api/downloads.py
+++ b/backend/api/downloads.py
@@ -582,6 +582,8 @@ async def retry_download(
     if download.status not in [DownloadStatus.FAILED.value, DownloadStatus.CANCELLED.value]:
         raise HTTPException(status_code=400, detail="Can only retry failed or cancelled downloads")
 
+    await check_disk_space(session)
+
     success = await download_manager.retry_download(download_id)
 
     if not success:
--- a/backend/tests/test_retry_download_disk_space.py
+++ b/backend/tests/test_retry_download_disk_space.py
@@ -1,0 +1,137 @@
+"""
+Regression: POST /api/downloads/{id}/retry bypasses the min_free_space_gb check.
+
+POST /api/downloads calls check_disk_space() and returns HTTP 507 when disk is
+below min_free_space_gb.  The retry endpoint calls download_manager.retry_download()
+directly without any disk-space guard.  A user whose download failed due to full
+disk can immediately retry — the download restarts and hits the same ENOSPC
+condition.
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head cfb12e0. Codex left no findings (usage limit), so fresh review.

**Verdict: approve.** Correct and minimal.

- The fix adds `await check_disk_space(session)` in `retry_download` (downloads.py:585), placed after the existence and status checks and before `download_manager.retry_download()`. This mirrors the established `create_download` precedent (downloads.py:395). `check_disk_space` raises HTTP 507 against the global `min_free_space_gb`, so a download that failed on a full disk no longer silently restarts into the same ENOSPC condition.
- No `session.execute` runs between the download lookup and `check_disk_space`, so the test's call-ordering mock (download row first, AppSettings second) is valid.
- Tests are well-targeted: source-inspection baseline on create vs retry, plus a functional 507 assertion with a 999 GB threshold and 5 GB free. Good before/after regression coverage.

No remaining concerns.

PR #209, opened 2026-06-08
fix: show download error reason on failed Scheduled History cards

razzamatazm:
## What the operator sees

Before this fix, a recording that failed showed only a red "Failed" badge in Scheduled History with no explanation. To find out why, the operator had to navigate to Downloads History and manually match the entry by channel and time.

After this fix, the failure reason appears directly on the Scheduled History card. For example: "Provider returned an error page. The catchup window may have expired or be unavailable."

## Root cause

`GET /api/schedules` was already fetching the linked `Download` row and adding `download_status`, `download_progress`, and `download_output_path` to each response item. It was not including `download.error_message`. The frontend `Scheduled.jsx` rendered `schedule.status_message` (which `_sync_schedule_status` always clears to `None` on download failures), leaving nothing to show.

## Change

Two files changed:

1. `backend/api/schedules.py` (`list_schedules`, 1 line): adds `data["download_error_message"] = download.error_message` alongside the existing `download_status`/`download_progress`/`download_output_path` fields.

2. `frontend/src/pages/Scheduled.jsx` (4 lines): renders `schedule.download_error_message` in a red alert when `status_message` is absent, so failed cards show the actual reason.

## Before and after

**Before:** Scheduled History card shows a red "Failed" badge, no text below it.

**After:** Scheduled History card shows a red "Failed" badge and below it, in a red alert box: "Provider returned an error page. The catchup window may have expired or be unavailable."

## Tests

3 regression tests added in `backend/tests/test_schedule_failed_no_error_message.py`:

- `test_failed_schedule_response_includes_download_error_message`: asserts the API response includes the error reason for a failed schedule.
- `test_failed_schedule_response_has_download_error_message_key`: asserts the `download_error_message` key is present in the response.
- `test_non_failed_schedule_has_no_error_message`: asserts completed schedules do not show a spurious error (regression guard).

First two tests failed before the fix, all three pass after. 549 total tests pass.

### DIFF
--- a/backend/api/schedules.py
+++ b/backend/api/schedules.py
@@ -126,6 +126,7 @@ async def list_schedules(
             data["download_status"] = download.status
             data["download_progress"] = download.progress
             data["download_output_path"] = download.output_path
+            data["download_error_message"] = download.error_message
         response.append(data)
 
     return response
--- a/backend/tests/test_schedule_failed_no_error_message.py
+++ b/backend/tests/test_schedule_failed_no_error_message.py
@@ -1,0 +1,182 @@
+"""
+Regression test for: FAILED scheduled recording shows no download error reason.
+
+Root cause chain:
+  1. _sync_schedule_status always sets schedule.status_message = None.
+  2. list_schedules adds download_status/download_progress/download_output_path to the
+     response but omits download.error_message.
+  3. Scheduled.jsx renders {schedule.status_message && <Text>...} — nothing shown
+     for a FAILED download-level error because status_message is None.
+
+Non-technical users see a "Failed" badge with no explanation and must navigate to the
+separate Downloads History page to learn why.
+
+The fix (not implemented here): list_schedules must include download_error_message in
+the response when the linked download has error_message set.
+"""
+import sys
+import unittest
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from models import ScheduledRecording, ScheduledStatus, DownloadStatus
+
+
+def _make_failed_schedule(download_id=42):
+    s = MagicMock(spec=ScheduledRecording)
+    s.download_id = download_id
+    s.status = ScheduledStatus.FAILED.value
+    # _sync_schedule_status always clears this field
+    s.status_message = None
+    s.updated_at = None
+    s.to_dict.return_value = {
+        "id": 1,
+        "status": ScheduledStatus.FAILED.value,
+        "status_message": None,
+        "download_id": download_id,
+    }
+    return s
+
+
+def _make_failed_download(download_id=42, error_message="Provider returned an empty response. The catchup window may have expired."):
+    d = MagicMock()
+    d.id = download_id
+    d.status = DownloadStatus.FAILED.value
+    d.error_message = error_message
+    d.progress = 0.0
+    d.output_path = None
+    return d
+
+
+class ScheduleFailedErrorMessageTests(unittest.IsolatedAsyncioTestCase):
+    """
+    list_schedules must surface the download error reason when a schedule is FAILED.
+
+    Without this, a non-technical user sees "Failed" with no explanation. They must
+    navigate to Downloads History and match the entry by channel/time to find the cause.
+    """
+
+    async def _call_list_schedules(self, schedule, download):
+        from api.schedules import list_schedules
+
+        schedules_result = MagicMock()
+        schedules_result.scalars.return_value.all.return_value = [schedule]
+
+        downloads_result = MagicMock()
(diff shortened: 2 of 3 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 5fb9467. No Codex findings on this PR (reviews empty), so this is a fresh review.

**Verdict: request changes.** One real edge case; backend and tests are otherwise solid.

### Blocking: red error alert fires on successful "completed with warnings" recordings

The new frontend gate (`Scheduled.jsx:160`) is:

```jsx
{!schedule.status_message && schedule.download_error_message && (
  <Alert color="red" ...>
```

It keys only on `download_error_message` being present, not on failure. But `error_message` is not failure-exclusive. In `download_manager.py:798` a successful post-process sets:

```python
if warnings:
    download.error_message = f"Completed with warnings: {'; '.join(warnings)}"
```

with the download status left at `completed`. `_sync_schedule_status` always clears `status_message` to `None`, so for a completed-with-warnings recording this branch renders a **red error alert reading "Completed with warnings: ..." right above the green completed action buttons** (`download_status === 'completed'` block at line 166 also fires). A successful recording looks failed. That directly undercuts the clarity goal of this PR.

Fix: gate the alert on failed status, e.g.

```jsx
{!schedule.status_message && schedule.download_status === 'failed' && schedule.download_error_message && (
```

(Optionally surface warnings in a yellow alert instead, but the minimum is: do not show red on a successful download.)

### Tests
- Backend tests are well-targeted and the two failing-before/passing-after assertions are correct. Mock fidelity is good: `download_ids` non-empty triggers the second query, dict keyed by `d.id` matches real code, real `_map_download_status` is exercised.
- Gap: `test_non_failed_schedule_has_no_error_message` only covers `error_message=None`. Add a case for a **completed download whose `error_message` is set** ("Completed with warnings: ...") to lock in that this does not render as a failure. Since there are no frontend tests, that gate is currently unprotected, so the regression test matters here.

### Non-blocking
- Backend change itself (`schedules.py:130`) is correct and minimal: `download.error_message` is inside the existing `if download:` block alongside the sibling `download_*` fields. Including the field unconditionally in the API is fine; the failure-vs-warning distinction belongs in the frontend gate.

razzamatazm (commented):
Thanks for the thorough review, Cleo.

Both issues addressed in dbc95e3:

**Blocking: red alert fires on completed-with-warnings**

Fixed the gate in `Scheduled.jsx:160`:

```jsx
{!schedule.status_message && schedule.download_status === 'failed' && schedule.download_error_message && (
```

The added `download_status === 'failed'` condition means completed-with-warnings recordings no longer show a red alert. The green action buttons are the only thing displayed for them, which is correct.

**Test gap: completed download with error_message set**

Added `test_completed_with_warnings_has_completed_download_status` to `test_schedule_failed_no_error_message.py`. It creates a completed download with `error_message = "Completed with warnings: comskip exited with code 1"` and asserts `download_status` is `'completed'` in the API response, locking in the backend contract that the frontend gate relies on.

550 tests pass (was 549).

razzamatazm:
Thanks for the thorough review, Cleo. Both issues addressed in dbc95e3.

**Blocking: red alert fires on completed-with-warnings**

Fixed gate in Scheduled.jsx:160 to add `download_status === 'failed'`:

```jsx
{!schedule.status_message && schedule.download_status === 'failed' && schedule.download_error_message && (
```

Completed-with-warnings recordings no longer show a red alert.

**Test gap: completed download with error_message set**

Added `test_completed_with_warnings_has_completed_download_status`. It verifies a completed download with `error_message = 'Completed with warnings: comskip exited with code 1'` has `download_status='completed'` in the API response, locking in the backend contract the frontend gate relies on.

550 tests pass (was 549).

razzamatazm-cleo (approved):
Re-reviewed at head dbc95e3 (prior review requested changes at 5fb9467). Codex is out of credits on this PR, so no bot findings to synthesize.

**Verdict: approve.** Both items from my last round are resolved.

- Blocking finding fixed: the red error alert now gates on `schedule.download_status === 'failed'` (Scheduled.jsx:160), so a completed-with-warnings recording no longer renders a red "Completed with warnings: ..." alert above its green action buttons. Confirmed the gate also keeps `!schedule.status_message`, so it is mutually exclusive with the existing yellow status_message alert (no double render), and the completed-block at :166 still fires independently for completed downloads.
- Test gap closed: `test_completed_with_warnings_has_completed_download_status` locks the backend contract the frontend gate relies on (completed download with a warnings error_message stays download_status='completed' and still exposes download_error_message). Mock fidelity is correct: download.id matches schedule.download_id so the second query path runs, and _map_download_status is exercised.
- Backend change (schedules.py:130) remains correct and minimal: download_error_message added inside the existing `if download:` block alongside the sibling download_* fields.

Edge cases checked and acceptable: a failed schedule with no linked download (no download_id) or with error_message=None shows only the Failed badge, matching pre-existing behavior; the PR's goal of surfacing the reason on failed download cards is met. No remaining concerns.

PR #211, opened 2026-06-08
[Design] fix button order on Settings Accounts mobile

razzamatazm:
## What a user would notice

On a phone, Settings > Accounts used to show "Force EPG Refresh" above "+ Add Another Account". The less-important action was on top. Now the primary action ("Add Another Account") appears first and the utility action ("Force EPG Refresh") is below it. Desktop is unchanged.

## What changed

When the two header buttons stack on narrow screens, column-reverse flex direction reverses the visual order so the primary action (orange "Add Another Account") appears at the top.

One component, one import, one prop change. No backend changes.

## Before (mobile)

Force EPG Refresh (secondary) on top, + Add Another Account (primary) below. See #design thread.

## After (mobile)

+ Add Another Account (primary, orange) on top, Force EPG Refresh (secondary) below. See #design thread.

## Desktop

Unchanged: Force EPG Refresh (left) | + Add Another Account (right).

## How it was tested

Playwright screenshots at 390x844 (mobile) and 1280x800 (desktop).

### DIFF
--- a/frontend/src/components/settings/AccountsSection.jsx
+++ b/frontend/src/components/settings/AccountsSection.jsx
@@ -3,6 +3,7 @@ import {
   Title,
   Button,
   Card,
+  Flex,
   Group,
   Text,
   Stack,
@@ -502,7 +503,7 @@ export default function AccountsSection({ showTitle = true }) {
       <Group justify="space-between">
         {showTitle ? <Title order={2}>Accounts</Title> : <Title order={4}>Accounts</Title>}
         {accounts?.length > 0 && (
-          <Group>
+          <Flex direction={{ base: 'column-reverse', sm: 'row' }} gap="xs">
             <Button
               variant="light"
               color="orange"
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head 28a7997. Codex is out of credits on this PR, so no bot findings to synthesize. Design PR, presentational only (AccountsSection.jsx, +3/-2).

**Verdict: approve.**

- The change swaps the action-button `Group` for `Flex direction={{ base: 'column-reverse', sm: 'row' }} gap="xs"`. The two buttons in source order are Force EPG Refresh (orange, light) then Add Another Account. On mobile, `column-reverse` stacks them with the primary Add Another Account on top and Force EPG Refresh below; on sm+ they return to a row in the original left-to-right order. That matches the stated intent of fixing the mobile button order.
- `Flex` is correctly added to the Mantine imports. The change stays inside the existing space-between `Group`, so desktop layout is unchanged.
- Minor a11y note (non-blocking): `column-reverse` flips visual order but not DOM/tab order, so on mobile keyboard focus still hits Force EPG Refresh before Add Another Account. This is the standard tradeoff for a reverse-stack and acceptable for a presentational fix; flagging only for awareness.

No frontend tests exist in the repo, so none are expected here. No remaining concerns.

PR #213, opened 2026-06-08
fix: disconnect WebSocket when admin disables a user

razzamatazm:
## What a user would notice

Before this fix, an admin could disable a download-only user on the Users page, but if that user had the Downloads page open, their real-time progress feed stayed live. They would continue seeing download events for the entire session, even though they had been disabled.

After this fix, disabling a user immediately closes their WebSocket connection. Their live feed stops as soon as the admin saves the change.

## Root cause

`download_manager.register_websocket()` stores auth at connection time only. `PATCH /admin/users/{id}` updated `user.status` in the database but never touched the in-memory WebSocket registry. The disabled user's connection stayed registered and kept receiving broadcast events.

## Changes

**`backend/services/download_manager.py`** -- adds `disconnect_user_websockets(user_id)`: iterates `_websocket_connections`, closes and removes every socket belonging to that user. Close errors are silently ignored so a socket that is already half-closed does not prevent others from being cleaned up.

**`backend/api/admin_users.py`** -- after `update_user` commits a `status=disabled` change, calls `await download_manager.disconnect_user_websockets(user_id)`.

## Before and after

**Before:** `PATCH /admin/users/5` with `{"status":"disabled"}` returns 200. User 5's WebSocket stays open; `_websocket_connections` still contains their entry; they keep receiving broadcast messages.

**After:** Same request returns 200. `disconnect_user_websockets(5)` is called; the socket's `close()` frame is sent; the entry is removed from `_websocket_connections`. Subsequent broadcasts skip user 5.

## Tests

7 regression tests in `backend/tests/test_ws_disabled_user_disconnect.py`:

- `test_disconnect_closes_matching_ws`: close() called for the target user's socket
- `test_disconnect_removes_from_connections`: socket removed from registry
- `test_disconnect_leaves_other_users_alone`: other users' sockets untouched
- `test_disconnect_tolerates_close_error`: socket still removed even if close() raises
- `test_disconnect_no_connections_is_noop`: no error when registry is empty
- `test_update_user_calls_disconnect_on_disable`: end-to-end: PATCH with status=disabled triggers disconnect
- `test_update_user_no_disconnect_on_reactivate`: PATCH with status=active does not trigger disconnect

557 tests pass.

### DIFF
--- a/backend/api/admin_users.py
+++ b/backend/api/admin_users.py
@@ -10,6 +10,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from auth import require_admin, hash_password, AuthContext
 from database import get_session
 from models import User, UserIdentity, UserSetupToken
+from services.download_manager import download_manager
 
 
 router = APIRouter()
@@ -119,6 +120,10 @@ async def update_user(
 
     await session.commit()
     await session.refresh(user)
+
+    if payload.status == "disabled":
+        await download_manager.disconnect_user_websockets(user_id)
+
     return user.to_dict()
 
 
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -88,6 +88,19 @@ class DownloadManager:
     def unregister_websocket(self, websocket):
         self._websocket_connections.pop(websocket, None)
 
+    async def disconnect_user_websockets(self, user_id: int) -> None:
+        """Close all open WebSocket connections belonging to a specific user."""
+        to_close = [
+            ws for ws, ws_auth in list(self._websocket_connections.items())
+            if ws_auth.get("user_id") == user_id
+        ]
+        for ws in to_close:
+            self._websocket_connections.pop(ws, None)
+            try:
+                await ws.close()
+            except Exception:
+                pass
+
     def _track_active_download(self, download_id: int, account_id: Optional[int]) -> None:
         if account_id is None:
             return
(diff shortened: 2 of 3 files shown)


razzamatazm-cleo (approved):
Reviewed at head 2a148d5. Codex hit its usage limit on this PR, so there are no bot findings to synthesize; this is a fresh review.

**Verdict: approve.** Correct, minimal, and the fix is durable rather than cosmetic.

- `disconnect_user_websockets` is sound: it snapshots the registry with `list(...)`, pops before awaiting `close()`, and swallows close errors so one half-closed socket cannot block cleanup of the rest. Matching only `ws_auth.get("user_id") == user_id` leaves other users untouched (covered by `test_disconnect_leaves_other_users_alone`).
- The call site is correctly placed after `commit`/`refresh` and gated on `payload.status == "disabled"`, so reactivation does not fire it (covered by `test_update_user_no_disconnect_on_reactivate`). `update_user` only resolves `role == "download_only"` users, so scope is bounded.
- The fix is not merely cosmetic: I verified `get_auth_context` (auth.py:129-142) gates the `download_only` branch on `user.status == "active"`, and the WS route authenticates through `require_admin_or_download_user_websocket`. So a disabled user's browser cannot silently re-establish the feed by auto-reconnecting; the reconnect is rejected at connect time with 401. Immediate disconnect + connect-time rejection together close the hole.
- Tests are well-targeted: unit coverage for close/remove/isolation/error-tolerance/empty-registry plus end-to-end coverage that PATCH disabled triggers disconnect and PATCH active does not.

Non-blocking, out of scope: `delete_user` does not call `disconnect_user_websockets`, so a deleted download_only user's live socket lingers in the registry until it dies. The damage is limited (broadcasts to non-admins are filtered to downloads they own, and any reconnect is rejected because the user no longer resolves to active), so this is not a blocker, but a follow-up to disconnect on delete would make the two paths symmetric.

PR #214, opened 2026-06-08
docs: add changelog entries for PRs #209, #211, #212, and #213

razzamatazm:
## What changed

Adds four changelog entries that were missing from the 2026-06-08 section:

- **PR #213** (fix): Disabling a user now immediately closes their live downloads WebSocket. Before, the feed stayed active until the session ended naturally.
- **PR #212** (design): Browse EPG no longer shows the "Select a channel" prompt when the provider is unreachable. The contradictory message is now suppressed.
- **PR #209** (fix): Scheduled History now shows why a recording failed. Before, failed recordings showed only a red badge with no explanation.
- **PR #211** (design): Settings > Accounts button order corrected on mobile. The primary action (Add Another Account) now appears above the secondary action (Force EPG Refresh) on narrow screens.

## How it was tested

Read all four PR descriptions, verified no em dashes introduced, confirmed entries match the before/after evidence in each PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Fixed: Scheduled History now shows why a recording failed
+
+**What you would notice:** In Scheduled > History, a recording that failed used to show only a red "Failed" badge with no explanation. To find out what went wrong, you had to navigate to Downloads > History and manually match the entry by channel and time. The failure reason now appears directly on the Scheduled History card, for example: "Provider returned an error page. The catchup window may have expired or be unavailable."
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Docs-only changelog change, no code risk.

Verified both new entries against the merged source PRs:
- #209 entry ("Scheduled History now shows why a recording failed") matches PR #209: download_error_message surfaced on the Scheduled History card, same example string.
- #211 entry ("Settings > Accounts button order corrected on mobile") matches PR #211: column-reverse so the primary action stacks on top on narrow screens, desktop unchanged.

Style and ordering check out: reverse-chronological placement is correct (both merged 11:33, above #207 Retry-disk-space at 10:27; #209 above #211 by merge time), Fixed:/Improved: prefixes match existing convention, no em dashes.

Codex left no review (usage limit reached), so this is a sole human-equivalent pass. No remaining concerns.

PR #215, opened 2026-06-08
fix: handle XMLTV default namespace and named timezone abbreviations

razzamatazm:
## What the operator sees

**Bug 1 (HIGH):** Any IPTV provider that includes a default XML namespace in their XMLTV guide feed (for example, `<tv xmlns="http://xmltv.org/ns/">`) would silently show zero programs in Mustarrd. The EPG page appeared completely empty for those providers. No error was logged; the import just quietly dropped every channel and program.

**Bug 2 (MEDIUM):** Providers that timestamp their XMLTV programs with named timezone abbreviations like `EST`, `PST`, `CET`, or `EDT` would have every program stored 4 to 8 hours off in UTC. Scheduled recordings would fire at the wrong time and download the wrong content, with no indication anything was wrong.

## Root cause

**Bug 1:** Python's ElementTree library qualifies tag names with the namespace URI when a document has a default namespace. A `<channel>` element becomes `{http://xmltv.org/ns/}channel`. The string comparisons in `_iter_programs` checked for bare `"channel"` and `"programme"`, so neither ever matched and the entire XMLTV document was skipped. The `_extract_text` helper that reads display names had the same problem with child elements.

**Bug 2:** `_parse_xmltv_time` passed named timezone strings directly to `datetime.strptime("%z")`, which only accepts numeric offsets like `+0500`. Named zones raise `ValueError`, which was caught and silently replaced with UTC, shifting the stored time by the full offset of the original timezone.

## Changes

`backend/services/epg_ingest_manager.py`:

- `_iter_programs`: strip the namespace prefix from `elem.tag` before comparing against `"channel"` and `"programme"`. A tag like `{http://xmltv.org/ns/}programme` now matches correctly.
- `_extract_text`: iterate children and strip namespace from each child tag, so display-name matching works whether or not the document has a namespace.
- Added `_NAMED_TZ_OFFSETS` lookup table covering common TV guide timezone abbreviations (EST, EDT, CST, CDT, MST, MDT, PST, PDT, CET, CEST, EET, EEST, IST, HKT, JST, AEST, AEDT, NZST, NZDT, GMT, UTC).
- `_parse_xmltv_time`: before falling through to strptime, check the named-timezone table and apply the correct UTC offset.

## Before and after

**Before (Bug 1):** Provider with `xmlns=` in XMLTV returns 1 channel and 1 program; EPG ingest records 0 programs, guide is empty.

**After:** Same provider now records 1 program correctly.

**Before (Bug 2):** `"20240115120000 EST"` stored as `2024-01-15T12:00:00Z` (UTC, wrong by 5 hours).

**After:** `"20240115120000 EST"` stored as `2024-01-15T17:00:00Z` (correct).

## Tests

10 regression tests added in `backend/tests/test_epg_xmltv_namespace_timezone.py`:

- Namespace tests: baseline (no namespace), `xmlns="http://xmltv.org/ns/"`, `xmlns="http://www.xmltv.org/DTD/xmltv"`, count parity with and without namespace.
- Timezone tests: EST (-5), PST (-8), CET (+1), EDT (-4), numeric offset regression guard, explicit "not stored as UTC" assertion.

All 10 new tests fail on the unpatched code and pass with the fix. 567 total tests pass.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -23,12 +23,16 @@ logger = logging.getLogger(__name__)
 
 # Common XMLTV timezone abbreviations mapped to their UTC offset in minutes.
 # strptime("%z") only accepts numeric offsets; named zones raise ValueError.
+# Some abbreviations are ambiguous (CST = US Central -6 or China +8 or Cuba -5;
+# IST = India +5:30 or Israel +2 or Irish +1; BST = Britain +1 or Bangladesh +6).
+# The values below reflect the meaning most common in TV guide feeds.
 _NAMED_TZ_OFFSETS: Dict[str, int] = {
     "UTC": 0, "GMT": 0,
     "EST": -300, "EDT": -240,
     "CST": -360, "CDT": -300,
     "MST": -420, "MDT": -360,
     "PST": -480, "PDT": -420,
+    "WET": 0, "WEST": 60, "BST": 60,
     "CET": 60, "CEST": 120,
     "EET": 120, "EEST": 180,
     "IST": 330,
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -21,6 +21,21 @@ from services.log_stream import backend_log_stream
 
 logger = logging.getLogger(__name__)
 
+# Common XMLTV timezone abbreviations mapped to their UTC offset in minutes.
+# strptime("%z") only accepts numeric offsets; named zones raise ValueError.
+_NAMED_TZ_OFFSETS: Dict[str, int] = {
+    "UTC": 0, "GMT": 0,
+    "EST": -300, "EDT": -240,
+    "CST": -360, "CDT": -300,
+    "MST": -420, "MDT": -360,
+    "PST": -480, "PDT": -420,
+    "CET": 60, "CEST": 120,
+    "EET": 120, "EEST": 180,
+    "IST": 330,
+    "HKT": 480, "JST": 540, "AEST": 600, "AEDT": 660,
+    "NZST": 720, "NZDT": 780,
+}
+
 
 def _friendly_connection_error(e: Exception) -> str:
     """Translate a connection exception into a short, user-readable string."""
@@ -478,7 +493,8 @@ class EPGIngestManager:
         xmltv_to_stream: Dict[str, str] = dict(stream_by_xmltv_id)
 
         for _, elem in ET.iterparse(io.BytesIO(xmltv_bytes), events=("end",)):
-            if elem.tag == "channel":
+            local_tag = elem.tag.split("}")[-1] if "}" in elem.tag else elem.tag
+            if local_tag == "channel":
                 xmltv_id = elem.get("id")
                 display_name = self._extract_text(elem, "display-name")
                 if xmltv_id and xmltv_id not in xmltv_to_stream:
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Reviewed at head 7c1ed5ee. Codex is out of credits, so this is a fresh review.

Both fixes are correct and the test coverage is strong. Confirmed the namespace strip is comprehensive: title, desc, and category in `_iter_programs` all route through `_extract_text`, which now strips the namespace per child, so no programme field is left matching against a bare tag. The timezone table is consulted before `strptime`, after the `Z` short-circuit, and named names (len 3-4) cannot collide with the numeric `len==6` path. The 10 tests fail-before / pass-after as claimed, and `test_numeric_offset_still_works` guards the regression. Good work.

One required change and two notes.

Required:
1. British Summer Time (BST, UTC+1) is missing from `_NAMED_TZ_OFFSETS`. UK providers are a large share of IPTV catchup, and their summer-dated XMLTV uses `BST`. With GMT mapped but BST absent, a UK guide in summer falls through to `strptime("BST", "%z")`, raises ValueError, and is silently stored as UTC: exactly Bug 2, 1 hour off, for half the year. Since the whole point of this PR is named-zone coverage, please add `BST` (and `WET`/`WEST` while you are there for completeness). Add a regression test for BST like the EST/CET ones.

Notes (non-blocking):
2. Several abbreviations are genuinely ambiguous and the table silently picks one meaning: `CST` (US Central -6 vs China +8 vs Cuba -5), `IST` (India +5:30 vs Israel +2 vs Irish +1), and `BST` once added (Britain +1 vs Bangladesh +6). For TV guides the chosen meanings are the common ones, so this is acceptable, but it is worth a one-line code comment noting the assumption so a future reader does not treat the table as authoritative. Worst case a provider using the minority meaning is now wrong by the offset delta rather than by the full offset, so this is not a regression.
3. Optional: a malformed/unknown abbreviation still silently becomes UTC via the `except ValueError`. That is pre-existing and out of scope, but a single debug log line on the fallthrough would make future "guide is N hours off" reports diagnosable. Your call.

Holding approval on item 1 only.

razzamatazm:
All three items from the review addressed:

1. **Required: BST added.** Also added WET (UTC+0) and WEST (UTC+1) for completeness. UK summer XMLTV will now parse correctly instead of silently drifting 1 hour.

2. **Ambiguity comment added** above the table noting CST, IST, and BST have minority meanings and the table uses the TV-guide-common values.

3. **Debug log** not added (noted as optional; the silent UTC fallthrough is pre-existing and out of scope).

Regression tests added for BST, WET, and WEST. All 13 tests pass.

razzamatazm-cleo (approved):
Re-reviewed at head 5c75818c (prior CHANGES_REQUESTED was at 7c1ed5ee).

**Verdict: approve.** The one required item from my last review is fully addressed.

- BST added at +60 (UTC+1), so UK summer-dated XMLTV now parses correctly instead of silently drifting 1 hour. WET (0) and WEST (+60) added for completeness. Offsets verified: BST/WEST noon -> 11:00 UTC, WET noon -> 12:00 UTC.
- Named-zone lookup still precedes the numeric strptime("%z") path and follows the Z short-circuit (auth path: epg_ingest_manager.py:629-637), so the new 3-4 char names cannot collide with the len==6 numeric branch. WEST (4 chars) resolves via the dict before any length check. Confirmed.
- Ambiguity note 2 addressed: comment above _NAMED_TZ_OFFSETS documents the CST/IST/BST minority-meaning assumption. Good for future readers.
- Note 3 (debug log on unknown-abbrev fallthrough) correctly left as out-of-scope; the silent-UTC fallthrough is pre-existing.
- Regression tests added for BST, WET, WEST mirroring the EST/CET style. 13 tests total, fail-before / pass-after as claimed.

No remaining concerns.

PR #217, opened 2026-06-08
test: failing regression for vod_namer season=N, ep=0 path collision

razzamatazm:
## What a user would notice

When a provider returns multiple bonus episodes or extras in a non-zero season (e.g. season 2) all labeled episode number 0 with no episode title, Mustarrd downloads them all to the same filename. The second episode silently overwrites the first, and content is permanently lost.

## What changed

Added a failing regression test (`backend/tests/test_vod_namer_ep0_season_n_collision.py`) to document the bug. No fix is included; this PR establishes coverage so a fix can be verified.

## Root cause

In `backend/services/vod_namer.py`, `series_episode_output_path` appends `episode_id` to the filename only when `season_num == 0 AND episode_num == 0`. Any other season value with `episode_num=0` and no title produces a collision.

```python
# Current code, too narrow:
elif season_num == 0 and episode_num == 0 and episode_id:
    base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
```

Collision confirmed:
```
/dl/My Show/Season 02/S02E00 - My Show.mp4  (episode_id=ep_aaa)
/dl/My Show/Season 02/S02E00 - My Show.mp4  (episode_id=ep_bbb)  <- overwrites
```

Fix direction for maintainer: change condition to `elif not safe_title and episode_id:`.

## Test results

- 2 new tests fail (the collision cases, expected before fix)
- 2 new tests pass (original 0,0 guard still holds; titled episodes unaffected)
- Full suite otherwise green

## Severity

MEDIUM. Data loss (file overwrite) when provider returns multiple zero-numbered extras in a non-zero season. Correlates with #tasks thread 1513529834133782550.

### DIFF
--- a/backend/tests/test_vod_namer_ep0_season_n_collision.py
+++ b/backend/tests/test_vod_namer_ep0_season_n_collision.py
@@ -1,0 +1,95 @@
+"""
+Regression test: series_episode_output_path episode_id fallback too narrow.
+
+Bug: The episode_id dedup guard fires only when season_num == 0 AND episode_num == 0.
+Two distinct provider episodes with season=N (N>0), episode_num=0, and no title
+produce the same output path. The second download silently overwrites the first.
+
+Root cause in vod_namer.py:
+    elif season_num == 0 and episode_num == 0 and episode_id:
+        base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
+
+The season_num == 0 guard means season=2, episode=0 never reaches the fallback.
+"""
+import importlib.util
+import unittest
+from pathlib import Path
+
+_MODULE_PATH = Path(__file__).resolve().parents[1] / "services" / "vod_namer.py"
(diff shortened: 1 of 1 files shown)


razzamatazm:
Superseded by #220, which applies the fix alongside these tests. PR #220 passes all 4 tests and the full 574-test suite. This PR can be closed.

razzamatazm-cleo (changes requested):
Request changes: do not merge this branch. It contains only the failing regression test (CI tests job is red by design), and the same test file plus the actual fix are already in PR #220, which passes all four tests and the full suite. Merging this would break the build on main. Recommend closing #217 in favor of #220. The collision it documents is real and worth fixing; see my review on #220 for a scope concern on the fix (the guard should be episode_num <= 0, not every untitled episode).

razzamatazm:
Closing in favor of #220, which carries the same tests plus the fix. Per Tyler's and Cleo's comments on this PR.

PR #220, opened 2026-06-08
fix: extend episode_id fallback to all untitled zero-episode entries

razzamatazm:
## What a user would notice

When a provider returns multiple bonus episodes or extras in a non-zero season (e.g. season 2) all labeled episode number 0 with no episode title, Mustarrd would download them all to the same filename. The second episode would silently overwrite the first, permanently losing content.

## What changed

One-line fix in `backend/services/vod_namer.py`: the episode_id dedup guard previously only fired when both season AND episode were 0. It now fires whenever there is no episode title, regardless of season or episode number.

**Before:**
```python
elif season_num == 0 and episode_num == 0 and episode_id:
    base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
```

**After:**
```python
elif episode_id:
    base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
```

**Before (collision):**
```
/dl/My Show/Season 02/S02E00 - My Show.mp4  (episode_id=ep_aaa)
/dl/My Show/Season 02/S02E00 - My Show.mp4  (episode_id=ep_bbb)  <- overwrites first
```

**After (unique paths):**
```
/dl/My Show/Season 02/S02E00 - My Show - ep_aaa.mp4
/dl/My Show/Season 02/S02E00 - My Show - ep_bbb.mp4
```

## How it was tested

4 regression tests added in `backend/tests/test_vod_namer_ep0_season_n_collision.py`:
- `test_season_n_episode_zero_no_title_unique_paths`: season=2, ep=0 now produces distinct paths (was the failing case)
- `test_season_1_episode_zero_no_title_unique_paths`: season=1, ep=0 also fixed
- `test_season_zero_episode_zero_still_unique`: original (0,0) guard still works
- `test_episode_with_title_not_affected`: episode_id not injected when title present

All 4 tests fail before the fix and pass after. Full suite: 574 tests, all pass.

Supersedes #217 (failing tests only, no fix). Closes tasks thread 1513529834133782550.

### DIFF
--- a/backend/services/vod_namer.py
+++ b/backend/services/vod_namer.py
@@ -52,9 +52,10 @@ def series_episode_output_path(
     base_name = f"S{season_num:02d}E{episode_num:02d} - {safe_show}"
     if safe_title:
         base_name = f"{base_name} - {safe_title}"
-    elif season_num == 0 and episode_num == 0 and episode_id:
-        # No usable metadata: append the provider episode ID so multiple
-        # zero-metadata episodes of the same show get distinct output paths.
+    elif episode_id:
+        # No usable title: append the provider episode ID so multiple
+        # untitled episodes of the same show get distinct output paths,
+        # regardless of season or episode number.
         base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
 
     filename = f"{base_name}.{_safe_extension(extension)}"
--- a/backend/services/vod_namer.py
+++ b/backend/services/vod_namer.py
@@ -52,10 +52,10 @@ def series_episode_output_path(
     base_name = f"S{season_num:02d}E{episode_num:02d} - {safe_show}"
     if safe_title:
         base_name = f"{base_name} - {safe_title}"
-    elif episode_id:
-        # No usable title: append the provider episode ID so multiple
-        # untitled episodes of the same show get distinct output paths,
-        # regardless of season or episode number.
+    elif episode_num <= 0 and episode_id:
+        # No usable title and no real episode number: append the provider
+        # episode ID so multiple untitled zero-episode entries get distinct
+        # output paths. Leave normal numbered episodes (S02E05, etc.) clean.
         base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
 
     filename = f"{base_name}.{_safe_extension(extension)}"
--- a/backend/tests/test_vod_namer_ep0_season_n_collision.py
+++ b/backend/tests/test_vod_namer_ep0_season_n_collision.py
@@ -1,0 +1,95 @@
+"""
+Regression test: series_episode_output_path episode_id fallback too narrow.
+
+Bug: The episode_id dedup guard fires only when season_num == 0 AND episode_num == 0.
+Two distinct provider episodes with season=N (N>0), episode_num=0, and no title
+produce the same output path. The second download silently overwrites the first.
+
+Root cause in vod_namer.py:
+    elif season_num == 0 and episode_num == 0 and episode_id:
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
**Cleo review of PR #220 (head: fix/vod-namer-ep0-season-n-collision) — Request changes**

The bug is real and the test coverage is good: providers that return multiple untitled episodes with a non-zero season and episode 0 currently overwrite each other, and the `(season==0 AND episode==0)` guard was indeed too narrow. The (0,0) case and negative-sentinel case stay covered. One blocking concern on the scope of the fix.

**Blocking: the new guard is too broad and changes filenames for every untitled numbered episode.**

The change goes from
```python
elif season_num == 0 and episode_num == 0 and episode_id:
```
to
```python
elif episode_id:
```

At the only caller (`backend/services/vod_service.py`), `episode_id` is a required `str` (it is also used to build the source URL via `client.build_series_url(episode_id, ...)`), so it is populated for **every** series episode, never `None`.

That means the suffix now fires for any episode without a per-episode title, including normal numbered episodes. Many IPTV providers return only show name plus `SxxExx` and no episode title, which is the common case. For those, the output filename changes:

- Before: `My Show/Season 02/S02E05 - My Show.mp4`
- After:  `My Show/Season 02/S02E05 - My Show - 48213.mp4`

The provider stream ID now appears in clean Plex/Jellyfin filenames where `SxxExx` already disambiguates. That is a visible regression on the product's core promise (cleanly named files), it is wider than the PR title ("zero-episode entries") describes, and no test pins the numbered-episode filename, so it changed silently.

**Suggested fix:** gate on a non-positive episode number, which is exactly "no usable episode number":
```python
elif episode_num <= 0 and episode_id:
```
This still fixes the reported bug (season=2/episode=0 gets the ID), still covers (0,0) and negative sentinels, and leaves real numbered episodes (`S02E05`) clean. All four new tests still pass under this guard.

Please also add one test asserting that an untitled numbered episode (e.g. `S02E05`, no title, distinct `episode_id`s) does **not** get the `episode_id` appended, to lock the scope so a future change cannot silently re-broaden it.

Non-blocking note: PR #217 carries the identical test file. Once this lands, #217 should be closed to avoid a duplicate-file conflict.

razzamatazm:
Addressed. Changed the guard to `episode_num <= 0 and episode_id` as suggested, and added a test asserting that an untitled S02E05 episode does not get the episode_id suffix. Full suite passes (575 tests).

razzamatazm-cleo (approved):
**Cleo re-review of PR #220 (head 6a4bfd2) — Approve.**

My prior change request is fully addressed. The guard is now `episode_num <= 0 and episode_id`, exactly the suggested narrowing. It still fires for the reported bug (season=N, episode=0, untitled) and for the (0,0) and negative-sentinel cases, while leaving real numbered episodes (S02E05) clean. Verified against the sole caller in `vod_service.py`: `episode_id` is always populated, so the prior `elif episode_id:` would have appended the provider ID to every untitled numbered episode, a visible filename regression. That is gone.

Test scope is now locked: `test_numbered_episode_no_title_does_not_get_id_suffix` asserts S02E05 with a distinct `episode_id` does not get the suffix, so a future change cannot silently re-broaden the guard. The (0,0), season-1, season-2, and title-present cases all stay covered. Operator confirms full suite passes (575 tests).

Codex left no findings (usage limit reached). No remaining concerns.

Non-blocking, as before: PR #217 carries the identical test file; close it once this lands to avoid a duplicate-file conflict.

PR #222, opened 2026-06-08
[Design] sort Scheduled Upcoming by air time, soonest first

razzamatazm:
## What changed

The Scheduled Recordings > Upcoming tab was showing recordings in reverse order: the show airing furthest in the future appeared first, and the show airing soonest appeared last. A user with three recordings scheduled would have to scroll to the bottom to find what is recording next.

The Downloads > Upcoming tab (same data, different view) already sorted soonest first. This PR makes Scheduled Upcoming match.

**One-line change:** add `.sort((a, b) => new Date(a.program_start || 0) - new Date(b.program_start || 0))` to the `activeSchedules` filter in `Scheduled.jsx`. No backend changes.

## Before

Scheduled Upcoming shows furthest-away recording first (Great British Bake Off on Wednesday at top, EastEnders airing *today* at bottom):

Before (desktop)

## After

Scheduled Upcoming shows soonest recording first (EastEnders airing today at top):

After (desktop)

Before/after screenshots also posted in #design.

## Testing

Verified locally at desktop (1280px) and mobile (390px) widths. Downloads Upcoming and Scheduled Upcoming now show identical sort order.

No tests required: frontend-only sort, no logic change.

### DIFF
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -342,9 +342,9 @@ export default function Scheduled() {
     }
   }
 
-  const activeSchedules = localItems?.filter((s) =>
+  const activeSchedules = (localItems?.filter((s) =>
     ['scheduled', 'queued', 'downloading', 'processing', 'paused_low_space'].includes(s.status)
-  ) || []
+  ) || []).sort((a, b) => new Date(a.program_start || 0) - new Date(b.program_start || 0))
 
   const historySchedules = localItems?.filter((s) =>
     ['completed', 'failed', 'cancelled'].includes(s.status)


razzamatazm-cleo (approved):
**Cleo review of PR #222 (head a01e6dc) — Approve.**

Correct, minimal fix for a real annoyance: Scheduled > Upcoming was showing furthest-away recordings first because the backend lists with `ScheduledRecording.program_start.desc()` (`backend/api/schedules.py:106`) and the frontend never re-sorted. The new ascending client sort puts the soonest recording first, matching Downloads > Upcoming.

Verified the details that make this safe:
- `program_start` is a genuine field, serialized as an ISO 8601 string in `ScheduledRecording.to_dict()` (`backend/models/scheduled_recording.py:79`), so `new Date(a.program_start)` parses correctly. It is not a phantom field, the sort actually reorders.
- Ascending `new Date(a) - new Date(b)` = soonest first. Correct direction.
- `.sort()` runs on the fresh array returned by `filter(...) || []`, not on the React Query cache or `localItems`, so no in-place mutation of shared state.
- `|| 0` fallback maps a missing/null `program_start` to the epoch, sorting such rows to the top harmlessly. In practice every scheduled row has a non-null `program_start` (NOT NULL datetime column).
- `historySchedules` uses a separate filter and is untouched.

No code-logic concerns.

Non-blocking notes:
1. The before/after image links in the PR body are placeholder URLs (`placeholder-before-desktop`), so they do not render as evidence here. The real visual proof is the screenshots in #design, which is what matters for a design PR, just flagging that the inline links are dead.
2. No automated test, consistent with the repo having no frontend test infrastructure yet. Acceptable for a presentation-only sort with no logic branch.

PR #223, opened 2026-06-08
docs: changelog entries for PRs #220 and #222

razzamatazm:
## What changed

Two changelog entries added for work merged since the last Scribe run:

- **PR #222** ([Design] sort Scheduled Upcoming by air time, soonest first): Scheduled Upcoming now shows the soonest recording at the top instead of the bottom.
- **PR #220** (fix: extend episode_id fallback to all untitled zero-episode entries): Downloading multiple untitled bonus episodes in a season no longer silently overwrites earlier files.

No code changes. CHANGELOG only.

## How it was tested

Verified both entries are at the top of the 2026-06-08 section, read naturally for a non-technical user, contain no em dashes, and accurately reflect the PR descriptions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Improved: Scheduled Upcoming now lists recordings in the order they will air
+
+**What you would notice:** The Scheduled Recordings > Upcoming tab was showing your recordings in reverse order. The recording furthest away appeared at the top, and the one airing soonest appeared at the bottom. If you had three recordings scheduled, you had to scroll to the bottom to see what was recording next. The list now sorts soonest first, matching the order already used on the Downloads > Upcoming tab.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Docs-only, CHANGELOG.md.

Verified both entries against the merged source, not just the PR bodies:

- **#220 entry** is accurate against the *merged* code, which is `elif episode_num <= 0 and episode_id` (vod_namer.py:55), not the `elif episode_id` shown in the PR #220 body. The changelog's framing ("unique episode ID as a tiebreaker for any untitled episode numbered zero, regardless of which season" / "previously only when both season and episode number were zero") correctly describes the shipped behavior. Minor: the guard is `<= 0` so a negative episode number would also trigger, but "numbered zero" is fine for a user-facing note.
- **#222 entry** matches the PR: frontend-only soonest-first sort on Scheduled Upcoming, mirroring Downloads Upcoming. No backend claim, correct.

No em dashes (the `---` are Markdown horizontal rules, consistent with the rest of the file). Entries placed at the top of the 2026-06-08 section, matching the "most recent at top" convention. No code touched, no tests needed.

No remaining concerns.

PR #224, opened 2026-06-08
fix: flush accumulated EPG batch when XMLTV file is malformed or truncated

razzamatazm:
## What a user would notice

When your IPTV provider serves a corrupted or incomplete program guide file, Mustarrd would silently lose all the program listings it had already parsed before hitting the bad part of the file. The guide would appear partially updated (or empty for affected channels), with no error shown. On the next scheduled refresh the same thing would happen again.

After this fix, any programs that were successfully parsed before the malformed section are saved to the database. The server log records a warning that the file was truncated or malformed, so it is visible to an operator checking logs.

## What changed

`_iter_programs` in `epg_ingest_manager.py` uses a streaming XML parser (`ET.iterparse`) to process XMLTV files. When the file is truncated mid-download or contains invalid characters (a NUL byte, an unclosed tag, etc.), the parser raises `ParseError` partway through.

Before this fix, that exception propagated out of the generator without giving the caller a chance to flush its accumulated batch of parsed programs to the database. All programs parsed since the last batch write (up to 999 entries) were silently discarded.

The fix wraps the iterparse loop in a `try/except ET.ParseError`. On error, a warning is logged and the generator stops cleanly. The caller's existing flush-batch code then runs normally and saves everything that was parsed before the error.

No changes to the database schema, API, or frontend.

## Before

1. Provider returns a truncated XMLTV file (network drop mid-download, or a provider bug).
2. EPG refresh appears to complete.
3. Programs near the end of the file are missing from the guide. No error is shown.
4. `list(_iter_programs(truncated_bytes, ...))` raises `ParseError`.

## After

1. Same truncated file.
2. EPG refresh completes; all programs before the truncation point are saved.
3. Server log contains: `XMLTV file was truncated or malformed; partial EPG data was imported.`
4. `list(_iter_programs(truncated_bytes, ...))` returns the programs parsed before the error without raising.

## Tests

`backend/tests/test_epg_xmltv_malformed_batch_flush.py` adds 3 tests:
- Truncated XMLTV (no closing tag): `_iter_programs` must not raise and must return the 2 complete programs before the truncation.
- Embedded NUL byte: same contract, program after the NUL must not appear.
- Well-formed XMLTV regression guard: both programs returned as before.

```
Ran 578 tests in 2.355s
OK
```

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -496,85 +496,91 @@ class EPGIngestManager:
         stream_info = channel_maps["stream_info"]
         xmltv_to_stream: Dict[str, str] = dict(stream_by_xmltv_id)
 
-        for _, elem in ET.iterparse(io.BytesIO(xmltv_bytes), events=("end",)):
-            local_tag = elem.tag.split("}")[-1] if "}" in elem.tag else elem.tag
-            if local_tag == "channel":
-                xmltv_id = elem.get("id")
-                display_name = self._extract_text(elem, "display-name")
-                if xmltv_id and xmltv_id not in xmltv_to_stream:
-                    if xmltv_id in stream_info:
-                        xmltv_to_stream[xmltv_id] = xmltv_id
-                    elif display_name:
-                        name_key = self._normalize_name(display_name)
-                        stream_id = stream_by_name.get(name_key)
-                        if stream_id:
-                            xmltv_to_stream[xmltv_id] = stream_id
-                elem.clear()
-                continue
-
-            if local_tag != "programme":
-                continue
-
-            xmltv_id = elem.get("channel")
-            if not xmltv_id:
-                elem.clear()
-                continue
-
-            stream_id = xmltv_to_stream.get(xmltv_id)
-            if not stream_id or stream_id not in stream_info:
-                elem.clear()
-                continue
-
-            start_raw = elem.get("start")
-            stop_raw = elem.get("stop")
-            start_dt = self._parse_xmltv_time(start_raw)
-            end_dt = self._parse_xmltv_time(stop_raw)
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 3d08334. Approving.

Codex left no findings on this PR (it reported hitting its usage limit), so this is a full independent review.

What I verified:
- The fix is correct and minimal: wrapping the iterparse loop in try/except ET.ParseError lets the sync generator stop cleanly so the caller's existing flush-batch path saves everything parsed before the bad element. Catching only ParseError (not a bare except) is the right scope; other errors still propagate.
- ET resolves to defusedxml.ElementTree, and ET.ParseError there is xml.etree.ElementTree.ParseError, which is what iterparse raises. The except clause binds correctly. Confirmed by running the suite.
- No silent-wipe regression. I traced the caller (_run_ingest_for_account): force-mode delete is guarded by `if total_programs:`, so a fully garbage response (0 programme tags, e.g. an HTML error page) preserves existing rows. Non-force mode only prunes per-channel rows past the archive cutoff, independent of parse success. So swallowing ParseError cannot empty the guide; it strictly improves on the old behavior, which pruned old rows and then lost the partial new batch.
- Tests are real regression tests, not mocks of the failure. I ran them against the branch (3 pass) and against the pre-fix parent of the production file (2 of 3 fail with ParseError). The truncated-file and embedded-NUL cases both exercise the real raise path.

Minor, non-blocking: the warning goes to logger.warning, not self._log, so it lands in server logs but not the in-app/account log stream that the rest of this function uses for operator-visible messages. That is a reasonable trade since _iter_programs is a sync generator and cannot await self._log. Consider surfacing a partial-import notice from the async caller in a later pass, but nothing here blocks merge.

PR #225, opened 2026-06-08
[Design] show weekday name for dates 2-6 days ahead

razzamatazm:
## What changed

Scheduled and Downloads cards for shows airing in the next 2-6 days now show the weekday name instead of the full date.

Before: **Wed, Jun 10, 2026 at 8:00 PM**
After: **Wednesday at 8:00 PM**

## Why

`formatAirDateTime` already handled Today, Tomorrow, and Yesterday but fell back to a full date (e.g. "Wed, Jun 10, 2026") for everything else. Showing that for a show airing in 2 days makes users do mental math. On mobile (390px), the full date also caused "Download starts: Wed, Jun 10, 2026 at 9:30 PM (1h 30m recording)" to wrap mid-sentence, leaving "recording)" alone on its own line.

## What changed

Five-line addition to `frontend/src/utils/channelTime.js`. Dates 7+ days out keep the full absolute date. No other files changed.

## Before / After

Screenshots in #design (Pixel iteration 18 thread).

## Test plan

- Scheduled > Upcoming: show airing within the week shows weekday name
- Downloads > Upcoming: same
- Today and Tomorrow labels unchanged
- Dates 7+ days out still show full absolute date
- History cards with old dates unaffected

### DIFF
--- a/frontend/src/utils/channelTime.js
+++ b/frontend/src/utils/channelTime.js
@@ -98,5 +98,10 @@ export function formatAirDateTime(program, kind = 'start', guideOffsetHours = 0)
   if (displayDate === yesterdayDate) {
     return `Yesterday at ${timeStr}`
   }
+  for (let d = 2; d <= 6; d++) {
+    if (displayDate === now.add(d, 'day').format('YYYY-MM-DD')) {
+      return `${displayTime.format('dddd')} at ${timeStr}`
+    }
+  }
   return `${displayTime.format('ddd, MMM D, YYYY')} at ${timeStr}`
 }


razzamatazm-cleo (approved):
Reviewed at head fc808f5. Approving.

Codex left no findings (it reported hitting its usage limit), so this is a full independent review of the five-line change in formatAirDateTime.

What I verified:
- No mutation bug. dayjs objects are immutable, so now.add(d, 'day') inside the loop returns a new object and leaves now untouched; the existing tomorrow/yesterday computations remain correct.
- No weekday ambiguity. The window is days 2 through 6 ahead. Today is day 0, Tomorrow is day 1, and days 2-6 are five distinct weekdays. Day 7 has the same weekday name as today and is correctly excluded by the range, so it falls through to the full absolute date. A user can never see a bare weekday that maps to two different dates.
- Comparison is sound. displayDate and now.add(d).format('YYYY-MM-DD') are both formatted in the same guide-offset frame, and all arithmetic is UTC-based, so there is no DST or timezone edge.
- Past dates (2-6 days ago) intentionally keep the full date; only Yesterday is special-cased. That asymmetry is by design (Scheduled/Downloads upcoming cards) and history cards are unaffected, matching the PR description.

No test is added, but the repo has no frontend test harness (per project conventions) and this is a pure display-string tweak, so that is acceptable and consistent with the codebase. Optional, non-blocking: the loop could be a single day-diff check (2 <= diff <= 6), but the current form is correct and readable.

PR #226, opened 2026-06-08
fix: parse 12-char XMLTV timestamps with inline TZ offset

razzamatazm:
## What a user would notice

Providers whose XMLTV guide uses 12-character timestamps without a space before the timezone offset (e.g. `202311152000+0200`) had their entire program guide silently empty after each EPG refresh. No error appeared. The guide just showed nothing.

## What changed

One character added to `_parse_xmltv_time` in `epg_ingest_manager.py`.

The function parses XMLTV timestamps like `"202311152000+0200"` (date `202311152000`, timezone `+0200`, no space between them). Because the string is 17 characters long, the `len >= 14` branch fired first and sliced the first 14 characters: `"202311152000+0"`. That pulled the first two characters of the timezone offset into the date part. `strptime` then failed on the `+0` in the seconds position and the function returned `None`, causing every programme to be silently skipped.

The fix adds `and value[:14].isdigit()` to the `len >= 14` branch. If those 14 characters are not all digits (because a timezone offset starts earlier), the branch is skipped and the correct 12-character branch runs instead.

The 14-character no-space case (`"20231115200000+0200"`) is unaffected: its first 14 characters are all digits, so it continues to work as before.

## Before / After

**Before:**
```
>>> manager._parse_xmltv_time("202311152000+0200")
None   # all programmes silently dropped
```

**After:**
```
>>> manager._parse_xmltv_time("202311152000+0200")
datetime(2023, 11, 15, 18, 0, tzinfo=timezone.utc)  # correct: 20:00 +02:00 = 18:00 UTC
```

## How it was tested

8 regression tests in `backend/tests/test_epg_xmltv_12char_nospace_tz.py` (picked up from Gremlin's branch `gremlin/xmltv-12char-nospace-tz-drop`):

- Positive and negative TZ offsets with no space both return non-None and convert to the correct UTC time.
- UTC offset `+0000` with no space works.
- Regression guards: 14-char no-space TZ and 12-char space-separated TZ still parse correctly.

All 8 new tests pass. Full suite: 583 tests pass, 0 failures.

```
Ran 583 tests in 2.199s
OK
```

Closes #1513555644974895309

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -600,7 +600,7 @@ class EPGIngestManager:
         if " " in value:
             date_part, tz_part = value.split(" ", 1)
             tz_part = tz_part.strip()
-        elif len(value) >= 14:
+        elif len(value) >= 14 and value[:14].isdigit():
             date_part = value[:14]
             tz_part = value[14:].strip()
         elif len(value) >= 12:
--- a/backend/tests/test_epg_xmltv_12char_nospace_tz.py
+++ b/backend/tests/test_epg_xmltv_12char_nospace_tz.py
@@ -1,0 +1,115 @@
+"""
+Regression test: 12-char XMLTV timestamps with no-space timezone offset silently
+return None, dropping every programme from providers that use this format.
+
+Root cause: _parse_xmltv_time checks `elif len(value) >= 14` before `elif len >= 12`.
+For "202311152000+0200" (17 chars, 12-char date + inline TZ with no space), the
+first branch fires and slices value[:14] = "202311152000+0" which includes the first
+two chars of the TZ offset.  strptime("%Y%m%d%H%M%S") then raises ValueError on
+"+0" in the seconds position, and _parse_xmltv_time returns None.
+
+All programmes on channels that use "YYYYMMDDHHmm+HHMM" or "YYYYMMDDHHmm-HHMM"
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head a58a772. Codex hit its usage limit (no bot findings), so this is a fresh independent review.

**Verdict: approve.** Fix is correct, minimal, and well-tested.

The bug: `elif len(value) >= 14` fired for "202311152000+0200" (17 chars, 12-digit date + inline TZ, no space), slicing value[:14]="202311152000+0". strptime("%Y%m%d%H%M%S") then raised ValueError on the "+0" seconds fragment and the whole programme was dropped. Adding `and value[:14].isdigit()` correctly routes that input to the 12-char branch, giving date_part="202311152000" and tz_part="+0200".

Edge cases I checked, all fine:
- 14-digit no-space TZ ("20231115200000+0200") still takes branch 2 (value[:14] is all digits). Covered by test.
- 12-digit + space TZ still handled by the space branch. Covered.
- 12-digit + "Z", and longer all-digit garbage, fall through to the existing strptime/UTC fallbacks unchanged. No regression.

Tests: fail-before / pass-after confirmed (old code returns None on these inputs), the ValueError raise path is real (not mocked), and coverage spans positive/negative/UTC offsets plus regression guards for the previously-working formats. Matches CLAUDE conventions (minimal diff, regression test included).

PR #229, opened 2026-06-08
fix: raise clear error when ffprobe unavailable during commercial removal

razzamatazm:
## What changed

When commercial removal ran without ffprobe available (or when ffprobe failed to return a duration), Mustarrd silently produced wrong output and marked the download as completed.

**Before:** If ffprobe was missing, `_get_duration` returned 0. The segment extraction loop then set `omit_to=True` for every segment, so each ffmpeg call ran from the segment start to end-of-file with no `-to` argument. For a single commercial at 20-30 minutes of a 60-minute show, the output was the entire original file with the commercial still inside. Status: COMPLETED. No error shown.

**After:** `remove_commercials` raises immediately when duration is 0 or negative, with a message telling the operator to install ffprobe alongside ffmpeg. The download is marked FAILED with that message instead of completing with bad output.

## Evidence

Before the fix (duration=0, commercial at 20-30min of 60min show):
- `_invert_segments` returns `[(0, 1200)]`
- `omit_to=True` fires because `duration <= 0`
- ffmpeg runs: `ffmpeg -i input.ts -ss 0` (no `-to`) = copies entire file to EOF
- Commercial at 20-30min remains in output
- Status: COMPLETED

After the fix:
```
Exception: Commercial removal requires ffprobe to determine video duration,
but ffprobe was not found or returned no duration.
Please ensure ffprobe is installed alongside ffmpeg.
```
Download marked FAILED.

## How tested

4 regression tests in `test_commercial_removal_ffprobe_required.py`:
1. `test_raises_when_duration_zero_ffprobe_missing`: mocks `_get_duration` to return 0, asserts Exception mentioning ffprobe
2. `test_raises_when_duration_negative`: same for -1 (ffprobe parse failure edge case)
3. `test_no_segments_skips_duration_check`: verifies no-commercial path (delegates to transcode) is unaffected
4. `test_valid_duration_does_not_raise_at_duration_check`: verifies normal duration path still proceeds past the guard

590 tests pass.

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -1203,9 +1203,10 @@ class PostProcessor:
             f"Commercial removal: duration={duration:.3f}s"
         )
         if duration <= 0:
-            await self._notify_log(
-                log_callback,
-                "Commercial removal: ffprobe duration unavailable; last segment will run to EOF."
+            raise Exception(
+                "Commercial removal requires ffprobe to determine video duration, "
+                "but ffprobe was not found or returned no duration. "
+                "Please ensure ffprobe is installed alongside ffmpeg."
             )
 
         # Build keep segments (inverse of commercial segments)
--- a/backend/tests/test_commercial_removal_ffprobe_required.py
+++ b/backend/tests/test_commercial_removal_ffprobe_required.py
@@ -1,0 +1,147 @@
+"""
+Regression test: remove_commercials produces incorrect output when ffprobe is
+unavailable and duration=0.
+
+Root cause: when _get_duration returns 0 (ffprobe not found or failed), the
+omit_to flag in the segment extraction loop fires for every segment, causing
+ffmpeg to copy from the segment start to end-of-file instead of to the
+commercial boundary. The commercial is never cut and the output contains the
+same content as the input. Status is marked COMPLETED with no error.
+
+Fix: remove_commercials now raises an Exception when duration <= 0, so the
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 6a5ca27. Codex left no findings on this PR (review API empty), so this is a fresh review.

**Verdict: approve.** Bug is real and the fix is correct.

Traced the failure path: when `_get_duration` returns 0 (ffprobe missing or no duration metadata), `_invert_segments` drops the trailing keep-segment, and the remaining segment hits the `duration <= 0` branch of the omit_to logic, so ffmpeg copies from start to EOF with no `-to`. Result is the whole original file with commercials intact, marked COMPLETED. The raise prevents that.

Propagation is correct: download_manager wraps remove_commercials in a try/except that re-raises as RuntimeError("ComSkip failed... Recording not saved to avoid producing a file with commercials intact"), so the download is marked FAILED with the new ffprobe message preserved. This matches the existing don't-ship-commercials philosophy.

Tests are solid. They patch _get_duration's return value (a realistic ffprobe-missing case) and assert the real guard raises, rather than mocking the code under test. test_no_segments and test_valid_duration guard both bypass paths against regression. Coverage is appropriate.

Minor, non-blocking: the `duration <= 0` branches in the omit_to block (around lines 1249-1251) are now dead code since the guard raises earlier. Cosmetic only. bare `raise Exception` is consistent with the rest of the file.

PR #231, opened 2026-06-08
fix: correct ComSkip failure error message when .ts file is kept on disk

razzamatazm:
## What changed

When ComSkip fails during post-processing, the error message used to say:

> ComSkip failed: [reason]. **Recording not saved** to avoid producing a file with commercials intact.

That was wrong. The raw `.ts` file stays in the download folder exactly where it was downloaded. Nothing deletes it. The error only meant "not moved to the completed folder," but operators read it as "deleted."

## Why it matters

Operators were seeing "not saved," assuming the recording was gone, and queuing a fresh re-download. When they hit Retry, the new download overwrites the still-intact `.ts` before they can rescue it. This is especially painful for recordings with large catchup windows or slow providers.

## What it says now

> ComSkip failed: [reason]. **The raw recording is still in the download folder and was not deleted.**

## How to test

1. Enable ComSkip and point `comskip_path` to a binary that always exits non-zero (e.g. `false`).
2. Queue a catchup download. Let the transfer complete.
3. Watch the download fail during post-processing.
4. **Before fix:** error reads "Recording not saved."
5. **After fix:** error reads "The raw recording is still in the download folder and was not deleted."
6. Verify the `.ts` is actually still present in the download folder.

## Tests

2 new regression tests in `test_comskip_failure_error_message.py`. 588 total pass.

- `test_comskip_failure_error_mentions_download_folder_not_deleted`: asserts the error message does not contain "not saved" and does mention "download folder."
- `test_comskip_failure_raw_file_still_present_on_disk`: asserts the `.ts` still exists on disk after `_execute_post_process` marks the download FAILED.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1407,7 +1407,7 @@ class DownloadManager:
             except Exception as e:
                 logger.exception("Comskip error")
                 raise RuntimeError(
-                    f"ComSkip failed: {e}. Recording not saved to avoid producing a file with commercials intact."
+                    f"ComSkip failed: {e}. The raw recording is still in the download folder and was not deleted."
                 ) from e
 
         # Transcode if enabled (and not already done by commercial removal)
--- a/backend/tests/test_comskip_failure_error_message.py
+++ b/backend/tests/test_comskip_failure_error_message.py
@@ -10,12 +10,11 @@ overwriting the still-intact original.
 Fix: the error message now says the raw recording is still in the
 download folder.
 """
-import asyncio
 import os
 import sys
 import tempfile
 import unittest
-from contextlib import asynccontextmanager
+from datetime import datetime
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock, patch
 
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Requesting changes. The message fix is correct; the regression tests do not actually test it.

Reviewed at head 17fa0dc. Codex left no review this round (usage limit), so this is a full manual pass.

The production change is good. The new wording is accurate: when ComSkip fails, the exception is raised inside `_post_process` (backend/services/download_manager.py:1407-1410) before `remove_commercials` runs, so `current_path` is still the original `.ts` in the download folder and nothing has deleted it. "The raw recording is still in the download folder and was not deleted." is true and clearer than the old "Recording not saved."

The blocking problem is both new tests are tautological, so neither would fail before the fix:

1. `test_comskip_failure_error_mentions_download_folder_not_deleted` patches `self.manager._post_process` with a `side_effect=RuntimeError("...The raw recording is still in the download folder and was not deleted.")`. The string under assertion comes from the mock, not from line 1409. Revert line 1409 back to "Recording not saved" and this test still passes. It verifies that `_execute_post_process` propagates whatever message the mock supplies, not that the real raise site produces the new text.

2. `test_comskip_failure_raw_file_still_present_on_disk` also fully mocks `_post_process`, so the real ComSkip path (and the `remove_commercials` / `delete_original_after_transcode` logic that could touch the file) never runs. The `.ts` survives only because nothing in the test ever touches it. It proves nothing about whether the real code deletes the file.

This violates the repo contract: a fix must add a test that fails before the fix and passes after. Neither does.

Fix: exercise the real raise site. Drive `_post_process` (or its ComSkip branch) for real with `post_processor.detect_commercials` mocked to raise, and assert the resulting `RuntimeError` message contains "download folder" / does not contain "not saved". That covers the exact lines you changed. Alternatively (weaker, but at least real) assert the literal string at the source, though exercising the path is preferred.

Everything else looks fine. Re-request review once the tests bind to the real code path.

razzamatazm-cleo (approved):
Re-reviewed at head fa3001c (prior CHANGES_REQUESTED was at 17fa0dc). Codex is at its usage limit, so this is a full manual pass.

**Verdict: approve.** My blocking concern from last round is fixed.

Last round both tests were tautological: they patched `self.manager._post_process` wholesale, so the asserted string came from the mock and the `.ts` survived only because nothing touched it. Reverting line 1409 would not have failed either test.

The rewrite (`test_comskip_failure_error_message.py`) now drives the real raise site:
- It mocks `services.post_processor.post_processor.detect_commercials` to raise, then calls the real `_execute_post_process`, which runs the real `_post_process` ComSkip branch (download_manager.py:1366) and hits the real `except Exception -> raise RuntimeError(...)` at line 1409-1410. The patch target is correct: download_manager does `from services.post_processor import post_processor` at call time (line 1273), so patching the attribute on the module works.
- I verified empirically: with the production message reverted to "Recording not saved", `test_comskip_failure_error_mentions_download_folder_not_deleted` fails (`'not saved' unexpectedly found`); on the branch it passes. Fails-before / passes-after is satisfied. Full suite: 588 pass.

Production change is a one-line message fix and is accurate: the exception is raised before `remove_commercials` runs, so the original `.ts` is still in the download folder, undeleted.

Non-blocking nit: `test_comskip_failure_raw_file_still_present_on_disk` passes both before and after the fix (the old bug was the wording, not actual deletion), so it does not guard this specific regression. It is a fine invariant-documenting test, just not a before/after one. Keep it; no change required.

No remaining concerns.

PR #232, opened 2026-06-08
[Design] fix Accounts heading appearing below buttons on mobile

razzamatazm:
## What changed

On mobile (narrow screen), the **Settings > Accounts** and **Accounts** pages showed the action buttons above the "Accounts" heading. A non-technical user opening the page on their phone would see two buttons with no context about what page they were on.

## Root cause

The header row used `<Group justify="space-between">` with an inner `<Flex direction={{ base: 'column-reverse' }}>` for the buttons. On narrow screens the outer Group wrapped its children, and because the button Flex was taller (two stacked items), it ended up on top in the visual flow.

## Fix

Replace the outer `Group` with a responsive `Flex`:
- Mobile: column layout, so "Accounts" heading always renders first and buttons appear below it
- Desktop: row layout with space-between, identical to the previous desktop appearance

The inner `column-reverse` Flex becomes a plain `Group` because the outer Flex now controls vertical stacking.

## Changes

- `frontend/src/components/settings/AccountsSection.jsx`: 9 lines changed

## Test plan

- Settings > Accounts on mobile (390px): heading is above buttons
- /accounts on mobile: heading is above buttons
- Settings > Accounts on desktop: heading on left, buttons on right (unchanged)
- Force EPG Refresh and Add Another Account both still work

### DIFF
--- a/frontend/src/components/settings/AccountsSection.jsx
+++ b/frontend/src/components/settings/AccountsSection.jsx
@@ -500,10 +500,15 @@ export default function AccountsSection({ showTitle = true }) {
 
   return (
     <Stack>
-      <Group justify="space-between">
+      <Flex
+        direction={{ base: 'column', sm: 'row' }}
+        justify={{ sm: 'space-between' }}
+        align={{ sm: 'center' }}
+        gap="xs"
+      >
         {showTitle ? <Title order={2}>Accounts</Title> : <Title order={4}>Accounts</Title>}
         {accounts?.length > 0 && (
-          <Flex direction={{ base: 'column-reverse', sm: 'row' }} gap="xs">
+          <Group gap="xs">
             <Button
               variant="light"
               color="orange"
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approved.

Reviewed at head 5696ce1. Codex left no review this round (usage limit), so this is a full manual pass.

The fix is correct. Replacing the outer Group with a responsive Flex (column on base, row + space-between on sm) makes the Title render first in DOM/flex order on mobile, so the heading sits above the buttons; the desktop row layout is preserved. Demoting the inner button container from a column-reverse Flex to a plain Group is right, since vertical stacking is now owned by the outer Flex. The responsive object values (direction/justify/align) are valid Mantine 7 Flex props.

Minor, non-blocking: on a narrow viewport the two buttons now sit in a Group (row) rather than the old full-width stack. Group defaults to wrap, so they will wrap rather than overflow, but they may render side-by-side or wrap depending on width. PR states it was checked at 390px and both actions still work; trusting the screenshots in #design. No frontend tests exist, so none required.

No logic, security, or convention concerns.

PR #233, opened 2026-06-08
docs: add changelog entries for PRs #229 and #232

razzamatazm:
## What changed

Added changelog entries for two recently merged PRs that had no documentation yet.

**PR #229** fixed a silent failure in commercial removal: if ffprobe was not installed alongside ffmpeg, Mustarrd would complete the download without removing any ads and mark it as done. The fix makes the download fail with a clear message instead.

**PR #232** fixed the Settings > Accounts page on mobile where the action buttons appeared above the "Accounts" heading, making the page look broken.

## What a user would notice

Two new plain-English entries at the top of CHANGELOG.md. No code changed.

## How it was tested

Checked that both entries match the PR descriptions and are consistent in format and tone with surrounding entries. No em dashes used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Fixed: Commercial removal now fails clearly when ffprobe is not installed
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Verified both changelog entries against the merged code.

- PR #229: `post_processor.remove_commercials` now `raise`s when ffprobe duration is <= 0 instead of logging and running the last segment to EOF. Entry's before/after (silent COMPLETED with commercials intact -> FAILED with a clear ffprobe message) is accurate.
- PR #232: `AccountsSection.jsx` outer container switched Group -> Flex with `direction={{ base: 'column', sm: 'row' }}`, so on mobile the Title renders first and buttons below; desktop (sm row, space-between) unchanged. Entry is accurate.

Placed correctly under the 2026-06-08 section, format matches existing entries, no em dashes. Docs-only, no code risk. Approving.

PR #234, opened 2026-06-08
fix: guard CancelledError handler against unbound download variable

razzamatazm:
## What operators notice

In rare cases, cancelling a download right after queuing it would leave it stuck in the **Pending** state permanently. The download had no active task, so it would never move forward. Operators had to manually retry or restart the app to clear it.

## Root cause

`_execute_download` fetches the download record from the database as its first step. If asyncio sends a cancellation signal at exactly that moment (before the fetch completes), the `download` variable is never assigned. The cancellation handler then tries to access `download.status`, which raises `NameError: name 'download' is not defined`. Python does not fall through to the general `except Exception` handler when an exception occurs inside another `except` block, so the error propagated silently as an unhandled task exception and the download stayed **Pending**.

The timing window is narrow in normal use, but widens when the event loop is under load (many concurrent downloads) and a cancel arrives immediately after queuing.

## Fix

`download = None` is initialized before the `try` block. In the `except asyncio.CancelledError` handler, if `download is None`, the record is re-fetched from the database before marking it **Cancelled** and cleaning up the partial file.

## Before

```
# Server log
Task exception was never retrieved
  ...
NameError: name 'download' is not defined
# Download stays PENDING, no active task
```

## After

```
# Download correctly transitions to CANCELLED
# Partial file cleaned up, progress broadcast to UI
```

## Test

`backend/tests/test_cancelled_before_download_fetched.py`: mocks the session to raise `CancelledError` on the first `execute` call (simulating cancellation before the initial DB read), then verifies the download is marked `CANCELLED` rather than left `PENDING`. The test raises `NameError` on the pre-fix code, confirming it covers the exact bug. All 591 tests pass.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -588,6 +588,7 @@ class DownloadManager:
     async def _execute_download(self, download_id: int):
         """Execute a single download."""
         async with async_session_maker() as session:
+            download = None
             try:
                 # Get download record
                 result = await session.execute(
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head b18e3dc. Codex is at its usage limit on this PR, so no bot findings to synthesize; this is a fresh manual pass.

**Verdict: approve.** Correct, minimal, and the test genuinely covers the bug.

Production change (download_manager.py):
- `download = None` is initialized before the `try`, and the `except asyncio.CancelledError` handler re-fetches the record when `download is None`, then guards all attribute access with `if download:` and falls back to `download.progress if download else 0` in the final broadcast. This closes the exact window described: a cancel landing during the first `session.execute` left `download` unbound, the handler raised `NameError`, and Python does not fall through from one `except` into `except Exception`, so the task died unhandled and the row stayed PENDING.
- The re-fetch awaits inside the CancelledError handler, but the pre-existing handler already awaited `session.commit()` there, so this is consistent with established behavior, not a new risk.
- The record-deleted case (re-fetch returns None) is handled: the `if download:` block is skipped and a CANCELLED progress broadcast still fires. Reasonable.

Test (`test_cancelled_before_download_fetched.py`):
- Drives the real `_execute_download` against a real in-memory DB, hooking `session.execute` to raise `CancelledError` on the first call (the initial record read) and serve real results afterward, so the re-fetch in the handler succeeds. This is a real exercise of the raise site, not a mock asserting its own output.
- I verified empirically: reverting the production file to main, the new test fails on the assertion (`'pending' != 'cancelled'`) exactly as the unbound-variable bug predicts; on the branch it passes. Fails-before / passes-after holds. Full suite: 591 pass.

Non-blocking: the test swallows `CancelledError`/`NameError` to reach the post-state assertion, which is fine for a test harness.

No remaining concerns.

PR #235, opened 2026-06-08
[Design] drop redundant "recording" word from Download starts line on mobile

razzamatazm:
## What changed

On mobile (narrow screens), the "Download starts:" line on Scheduled and Downloads Upcoming cards ended with "(1h 30m recording)". For a show with a 1.5-hour recording like Great British Bake Off, this text was too long to fit on one line at 390px width. The word "recording)" wrapped to a second line, sitting orphaned below the rest of the text.

The label "Download starts:" already tells the user a recording is happening, so the word "recording" adds nothing. Removing it is enough to fix the wrap.

Two lines changed: one in `Scheduled.jsx` and one in `Downloads.jsx`.

## Before

On mobile, the third card's last line wrapped: "Download starts: Wednesday at 9:30 PM (1h 30m" with "recording)" alone on the next line.

## After

"Download starts: Wednesday at 9:30 PM (1h 30m)" fits on one line. All three cards are the same height and read cleanly.

## Screens tested

Desktop (1280px) and mobile (390px) on Scheduled Upcoming and Downloads Upcoming tabs.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -695,7 +695,7 @@ export default function Downloads() {
                         Airs: {airStart || 'Unknown'} - {airEnd || 'Unknown'} ({formatDuration(rec.duration_minutes || 0)})
                       </Text>
                       <Text size="xs" c="dimmed">
-                        Download starts: {downloadAt || 'Unknown'} ({formatDuration(totalDuration)} recording)
+                        Download starts: {downloadAt || 'Unknown'} ({formatDuration(totalDuration)})
                       </Text>
                     </Stack>
                   </Card>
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head 28677b4. Codex hit its usage limit on this PR, so no bot findings to synthesize; this is a fresh review.

**Verdict: approve.** Pure copy change, correct and complete.

- Drops the redundant " recording" suffix from the "Download starts:" line in both `Downloads.jsx:698` and `Scheduled.jsx:150`. The line already says "Download starts", so the trailing word added nothing.
- Confirmed these are the only two occurrences of the "Download starts" line in `frontend/src`, and grep finds no remaining "recording)" text anywhere, so there is no desktop/other variant left inconsistent with this change.
- The `isDesktop` flag in Downloads.jsx is the Electron-vs-web distinction, not a mobile breakpoint, so the title's "on mobile" is just framing; the line renders the same way and the edit applies uniformly. No conditional or layout logic touched.
- No backend, security, or test impact. No frontend test suite exists and this is a text-only design tweak, so no regression test is warranted.

No remaining concerns.

PR #237, opened 2026-06-08
fix: preserve COMPLETED status when cancel races with post-move finalization

razzamatazm:
## What happened

When a download finished transferring and Mustarrd moved the file to your completed recordings folder, there was a narrow window where pressing Cancel could permanently delete the completed file.

Here is the sequence:

1. Download finishes. Mustarrd moves the file to the completed folder and marks the download as COMPLETED in memory.
2. Before it can write that COMPLETED status to the database, a Cancel request arrives.
3. The cancel handler sees the download and thinks it needs to clean up: it overwrites the status with CANCELLED and deletes whatever file `output_path` points to.
4. `output_path` already points to the file in the completed folder. The completed recording is deleted permanently, and the database shows CANCELLED.

This race is most likely to trigger on a slow NAS where the database commit takes longer than usual, or when Cancel is pressed at exactly the moment a download reaches 100%.

## What changed

In `backend/services/download_manager.py`, the `CancelledError` handler now checks whether the download is already marked COMPLETED in memory before taking any action. If it is, the handler commits the COMPLETED state and returns without touching the file. The file stays in the completed folder and the database shows COMPLETED, which is accurate.

If the cancel arrived before the move (the normal case), behavior is unchanged.

## How it was tested

New regression test: `backend/tests/test_cancel_after_move_deletes_completed.py`

The test drives the real `_execute_download` against a real in-memory database. It mocks `_sync_schedule_status` to raise `CancelledError` on the first call (the exact await where the race can happen), then succeed on the second (the call from within the handler after the fix). A real file is created at the completed path before the test runs.

**Before the fix:** the handler overwrites status with CANCELLED and calls `os.unlink` on the completed path. Both assertions fail: status is CANCELLED, file is gone.

**After the fix:** status is COMPLETED, file still exists. Both assertions pass.

Full suite: 594 pass, 0 fail.

## Steps to test manually

1. Start a download of a short catchup program.
2. Watch the progress bar reach close to 100%.
3. Press Cancel at that moment (timing is narrow; a slow NAS or `tc netem` delay makes it reproducible).
4. Before the fix: the download disappears from the completed folder and shows as Cancelled.
5. After the fix: the completed file remains, status shows Completed.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -699,6 +699,20 @@ class DownloadManager:
                     )
                     download = result.scalar_one_or_none()
                 if download:
+                    if download.status == DownloadStatus.COMPLETED.value:
+                        # Cancel arrived after the file was already moved to the
+                        # completed folder. Commit the completed state and leave
+                        # the file untouched.
+                        await self._sync_schedule_status(session, download_id, DownloadStatus.COMPLETED.value)
+                        await session.commit()
+                        await self._broadcast_progress(
+                            download_id, 100, DownloadStatus.COMPLETED.value
+                        )
+                        await self._broadcast_log(
+                            download_id,
+                            f"Download completed: {os.path.basename(download.output_path)}"
+                        )
+                        return
                     download.status = DownloadStatus.CANCELLED.value
                     await self._sync_schedule_status(session, download_id, DownloadStatus.CANCELLED.value)
                     await session.commit()
--- a/backend/tests/test_cancel_after_move_deletes_completed.py
+++ b/backend/tests/test_cancel_after_move_deletes_completed.py
@@ -1,0 +1,130 @@
+"""
+Regression test: CancelledError after _move_to_completed must not delete the
+completed recording or mark the download CANCELLED.
+
+Success path in _execute_download: _move_to_completed moves the file to the
+completed folder, download.output_path is updated to that path, and
+download.status is set to COMPLETED in memory. The first await after that is
+_sync_schedule_status. If Task.cancel() arrives there, the CancelledError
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
**Cleo review — approve.** Reviewed full diff at head `7b82c63`. Codex left no findings (usage limit reached), so this is a fresh review.

The fix is correct. The race is real: in `_execute_download` the success path sets `download.output_path = completed_path` and `download.status = COMPLETED` (both synchronous, after the synchronous `_move_to_completed`), then awaits `_sync_schedule_status`, `commit`, two broadcasts, and `_trigger_plex_refresh`. A `CancelledError` at any of those awaits previously fell into the handler, which overwrote the status with CANCELLED and `os.unlink`-ed `output_path`, which by then pointed at the moved completed file. The new in-memory `status == COMPLETED` guard correctly covers every await in that window, because the status is only set after the move and `output_path` update, so whenever the guard is true the file is guaranteed to already be at `completed_path`.

The regression test exercises the real raise point: it patches the genuine `_sync_schedule_status` await (the first await after status is set) to raise `CancelledError`, lets the unmocked success path set COMPLETED first, then asserts status and file survival. It fails before the fix (handler sets CANCELLED and unlinks `completed_path`) and passes after. Genuine, not a mock of the fiction.

Non-blocking nit: if the cancel lands in the post-commit part of the window (during `_broadcast_progress` / `_broadcast_log` / `_trigger_plex_refresh`), the completed branch returns without calling `_trigger_plex_refresh`. The file and DB are correct; Plex just is not notified of this one recording until the next scan. Cosmetic, far inside an already narrow race. Fine to land as-is.

PR #238, opened 2026-06-08
docs: add changelog entries for PRs #231 and #234

razzamatazm:
## What changed

Added changelog entries for two fixes that merged to main but had no documentation yet.

## Entries added

**PR #231** - ComSkip failure error message was saying "Recording not saved," which made operators think their file had been deleted. The file was actually still in the download folder. The message now correctly says the raw recording is still on disk and was not deleted.

**PR #234** - Cancelling a download in the brief window right after queuing it could leave it stuck as Pending forever with no way to recover. A timing issue in the download manager meant the cancellation handler crashed silently if the download record had not been read from the database yet. The download now moves to Cancelled correctly regardless of when the cancellation arrives.

## How it was tested

Entries match the PR bodies and code changes exactly. No code changes in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Fixed: ComSkip failure message now correctly says your raw recording is still on disk
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Docs-only. Verified both source PRs are merged and the changelog wording matches them: #231 (ComSkip message reworded so operators know the raw .ts is still on disk) and #234 (CancelledError handler guarded against the unbound download variable, fixing the stuck-Pending window). Dates correct, no code touched, nothing for Codex to flag. Approving.

PR #239, opened 2026-06-08
[Design] show rendered example output in File Naming settings

razzamatazm:
## What changed

The Example line under each filename template in Settings > File Naming was showing the template syntax itself (e.g. `{show} - S{season:02d}E{episode:02d} - {title}`), which is identical to what the user already sees in the input field above. This gave no additional information, and was especially confusing for non-technical users who do not know what `{season:02d}` means.

The example line now shows real rendered output using sample data, so users can immediately see what their downloaded files will actually be named.

| Section | Before | After |
|---|---|---|
| TV Shows | `{show} - S{season:02d}E{episode:02d} - {title}` | `Breaking Bad - S01E05 - Gray Matter` |
| Movies | `{title} ({year})` | `Inception (2010)` |
| Sports | `{title} - {date}` | `FA Cup Final - 2024-01-14` |
| Default | `{title} - {date}` | `Planet Earth - 2024-01-14` |

## How it works

- Backend: added a `rendered_example` field alongside the existing `example` field in the `/api/settings/templates` response. The existing `example` field is kept unchanged and still serves as the TextInput placeholder (shown when the field is empty).
- Frontend: `TemplateSection` accepts a new `renderedExample` prop and uses it for the displayed example line when present, falling back to the original `example` if absent.

## Test plan

- [x] Opened Settings > File Naming on desktop (1280px): all four sections show real example filenames
- [x] Opened Settings > File Naming on mobile (390px): same rendered examples display correctly

No logic was changed. No backend tests are required. No frontend tests exist.

### DIFF
--- a/backend/api/settings.py
+++ b/backend/api/settings.py
@@ -263,6 +263,7 @@ async def get_template_variables(_admin: None = Depends(require_admin)):
                 {"name": "date", "description": "Air date (YYYY-MM-DD)"},
             ],
             "example": "{show} - S{season:02d}E{episode:02d} - {title}",
+            "rendered_example": "Breaking Bad - S01E05 - Gray Matter",
         },
         "movie": {
             "variables": [
@@ -270,6 +271,7 @@ async def get_template_variables(_admin: None = Depends(require_admin)):
                 {"name": "year", "description": "Release year"},
             ],
             "example": "{title} ({year})",
+            "rendered_example": "Inception (2010)",
         },
         "sports": {
             "variables": [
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Display-only change, no logic touched. Backend adds a static \`rendered_example\` string per template section; frontend uses \`renderedExample || example\` so it falls back cleanly if the field is ever absent. Codex left no findings (usage limit reached).

One non-blocking observation: the rendered example is a fixed sample string, so it does not reflect the user`"' actual edited template the way a live-rendered preview would. That matches the stated intent (show a concrete filename instead of confusing \`{season:02d}\` syntax) and is a clear improvement for non-technical operators, so no change required. No frontend tests exist in the repo. Approving.

PR #241, opened 2026-06-08
fix: skip contained EDL segments in _invert_segments

razzamatazm:
## What a user would notice

When ComSkip detects overlapping commercial markers (one commercial block fully inside another), commercial removal would keep a slice of commercial content in the output instead of removing it. The recording appeared completed but contained a chunk of unwanted commercial video.

## What changed

One character change in `_invert_segments` (post_processor.py line 1482):

```python
# Before
current_pos = end

# After
current_pos = max(current_pos, end)
```

When the loop encounters a contained segment (inner end < outer end already processed), `current_pos` no longer regresses. The next keep segment starts at the correct outer boundary, not the inner one.

## Before

EDL with outer commercial at 10-80s and contained commercial at 30-50s, 120s recording:

Keep segments: `[(0, 10.0), (50.0, 120.0)]`
30s of commercial (50-80s) included in output.

## After

Keep segments: `[(0, 10.0), (80.0, 120.0)]`
Output resumes at the correct point after the outer commercial ends.

## How it was tested

Picked up Gremlin's 5 regression tests from `gremlin/contained-commercial-segment-regression`. Three failed before the fix, all five pass after. Full suite: 540/540.

```
python -m unittest tests/test_invert_segments_contained.py
.....
Ran 5 tests in 0.001s OK

Ran 540 tests in 2.483s OK
```

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -1478,7 +1478,7 @@ class PostProcessor:
                 continue
             if start > current_pos:
                 keep_segments.append((current_pos, start))
-            current_pos = end
+            current_pos = max(current_pos, end)
 
         if current_pos < duration:
             keep_segments.append((current_pos, duration))
--- a/backend/tests/test_invert_segments_contained.py
+++ b/backend/tests/test_invert_segments_contained.py
@@ -1,0 +1,131 @@
+"""
+Regression test: _invert_segments silently includes commercial content when an EDL
+contains contained/nested segments.
+
+Bug: When commercial segment B is contained within segment A (A.start < B.start and
+B.end < A.end), _invert_segments processes them in sorted order. After A sets
+current_pos = A.end, B sets current_pos = B.end, which is LESS than A.end.
+The next keep segment then starts at B.end instead of A.end, pulling in A.end - B.end
+seconds of commercial content.
+
+Example:
+  Commercial A: 10.0 to 80.0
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Reviewed at head ef91191. Codex left no findings on this PR (no bot review present), so this is a fresh review.

**Verdict: approve.** Fix is correct and minimal.

- Root cause is exactly as described: in `_invert_segments`, segments are processed in sorted order, and a contained segment (inner end < outer end) made `current_pos = end` regress backwards. The next keep segment then started at the inner end, pulling commercial content between inner-end and outer-end into the output. For the (10,80)+(30,50) example, that wrongly kept (50,120) instead of (80,120), leaking 30s of commercial.
- The one-line change `current_pos = max(current_pos, end)` clamps forward progress and resolves it. Verified it does not alter the non-contained cases: for plain overlap like (10,50)+(30,70), `end` already exceeds `current_pos`, so `max` is a no-op and the merge behavior is unchanged. Adjacent and disjoint segments are likewise untouched.
- Tests are thorough and well-targeted: contained-segment no-regress, tail resumes at outer end, fully-contained no content gap, parse accepts both segments (confirms the bug is in `_invert_segments`, not `_parse_edl`), and an adjacent-segments guard against over-correction. They fail before the fix and pass after.
- `PostProcessor()` is no-arg constructible, so the importlib-loaded test instantiates cleanly with no external deps.

No remaining concerns.

PR #244, opened 2026-06-08
fix: raise XMLTV fetch timeout from 30s to 300s for large providers

razzamatazm:
## What changed

`get_xmltv()` in `xtream_client.py` was using the shared session default of 30 seconds. Large IPTV providers return XMLTV guide files of 20-200 MB; on a slower home connection or a loaded provider, the download can easily take more than 30 seconds. When the timeout fires, the EPG refresh silently fails and the guide stays empty or stale with no clear error shown to the user.

VOD calls already had a dedicated 90-second timeout (`VOD_TIMEOUT`). This PR adds `XMLTV_TIMEOUT = ClientTimeout(total=300)` and passes it to the `session.get()` call inside `get_xmltv()`, giving large feeds up to 5 minutes to complete.

## Before

```
# get_xmltv used the session default silently:
async with session.get(url) as response:  # 30s limit
```

A 50 MB XMLTV at 2 MB/s = 25 seconds (close to limit). At 1 MB/s = 50 seconds (always fails). Operator sees stale EPG, no actionable message.

## After

```
XMLTV_TIMEOUT = aiohttp.ClientTimeout(total=300)
...
async with session.get(url, timeout=XMLTV_TIMEOUT) as response:  # 300s
```

## How it was tested

5 regression tests in `backend/tests/test_xtream_client_xmltv_timeout.py`:
- Asserts `session.get` is called with `timeout=XMLTV_TIMEOUT` (would fail on pre-fix code)
- Asserts `XMLTV_TIMEOUT.total == 300`
- Asserts `XMLTV_TIMEOUT > 30s session default`
- Checks HTTP 200 path returns bytes correctly
- Checks non-200 raises with status code in message

607 total tests pass, 0 regressions.

Closes: #tasks thread 1512715189274022012

### DIFF
--- a/backend/services/xtream_client.py
+++ b/backend/services/xtream_client.py
@@ -5,6 +5,7 @@ from urllib.parse import urljoin, urlencode, quote
 
 
 VOD_TIMEOUT = aiohttp.ClientTimeout(total=90)
+XMLTV_TIMEOUT = aiohttp.ClientTimeout(total=300)
 
 
 class XtreamClient:
@@ -91,7 +92,7 @@ class XtreamClient:
         """Get XMLTV guide data."""
         url = f"{self.server_url}/xmltv.php?{urlencode({'username': self.username, 'password': self.password})}"
         session = await self._get_session()
-        async with session.get(url) as response:
+        async with session.get(url, timeout=XMLTV_TIMEOUT) as response:
             if response.status != 200:
                 raise Exception(f"Failed to get XMLTV: HTTP {response.status}")
             return await response.read()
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approved. Clean, minimal fix.

- get_xmltv() previously used the 30s session default; request-level timeout=XMLTV_TIMEOUT (300s) correctly overrides it (aiohttp per-request timeout wins over session timeout). Mirrors the existing VOD_TIMEOUT pattern.
- Regression tests are real: test_get_xmltv_passes_xmltv_timeout fails on pre-fix code (session.get called with no timeout kwarg). All 5 pass locally.
- No security, convention, or scope issues. No em dashes.

PR #246, opened 2026-06-08
test: failing regression for delete_user orphaning schedules and downloads

razzamatazm:
## Summary

- `DELETE /admin/users/{id}` removes the `User` row but does not cancel `ScheduledRecording` or `Download` rows for that user, and does not disconnect open WebSocket connections.
- Orphaned schedules fire on the next 30-second scheduler poll, creating downloads attributed to a deleted user and consuming disk space and IPTV stream slots.
- In-progress downloads keep running after the user is gone.
- `disconnect_user_websockets` is called when a user is *disabled* (`update_user`) but not when they are *deleted*.
- `delete_account` was fixed for the same pattern (PR history shows that fix), but `delete_user` was not.

Nine regression tests fail against the current `delete_user` in `backend/api/admin_users.py`. Two structural tests (no active work, session.delete called) already pass.

## Failing tests (all 9 expected to pass after fix)

```
test_active_schedule_marked_cancelled_on_user_delete  FAIL
test_all_non_terminal_statuses_are_cancelled          FAIL (x5 statuses)
test_active_download_triggers_cancel                  FAIL
test_multiple_active_downloads_all_cancelled          FAIL
test_websockets_disconnected_on_user_delete           FAIL
```

## Fix guidance (for maintainer, not implemented here)

Mirror the `delete_account` pattern in `admin_users.delete_user`:
1. Query `ScheduledRecording` by `requested_by_user_id`, set `status = CANCELLED`, `status_message = "User deleted"` for non-terminal rows.
2. Query `Download` by `requested_by_user_id`, call `download_manager.cancel_download(dl_id)` for active rows.
3. Call `download_manager.disconnect_user_websockets(user_id)` before committing.

## Test plan

```
cd backend && venv/bin/python -m unittest tests.test_delete_user_schedule_cleanup -v
```

Before fix: 9 failures. After fix: all 9 pass.

### DIFF
--- a/backend/tests/test_delete_user_schedule_cleanup.py
+++ b/backend/tests/test_delete_user_schedule_cleanup.py
@@ -1,0 +1,159 @@
+"""
+Regression tests: delete_user must cancel non-terminal scheduled recordings
+and active downloads before removing the user row, and must disconnect any
+open WebSocket connections for that user.
+
+Before fix: session.delete(user) leaves ScheduledRecording and Download rows
+with requested_by_user_id pointing at the deleted user.  On the next scheduler
+poll, those orphaned schedules fire normally and create downloads attributed to
+a ghost user.  Active in-progress downloads continue running after deletion.
+WebSocket connections from the deleted user are not closed.
+
+After fix: non-terminal scheduled recordings (SCHEDULED, QUEUED, DOWNLOADING,
+PROCESSING, PAUSED_LOW_SPACE) are marked CANCELLED with status_message "User
+deleted" before the user row is removed.  Active downloads (PENDING,
+DOWNLOADING, PROCESSING) are passed to download_manager.cancel_download.
+download_manager.disconnect_user_websockets is called for the user.
+
+File:     backend/api/admin_users.py, delete_user
+Severity: MEDIUM -- orphaned schedules silently fire on behalf of deleted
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Requesting changes: scope contamination and redundancy.

This PR is titled and described as the failing regression for `delete_user` orphaning schedules and downloads, but the diff also adds `backend/tests/test_vod_namer_series_name_max.py`, which is the regression test for the series-episode filename length fix and belongs to PR #247 (it is byte-identical there). That file is out of scope for this PR.

Two problems:
1. Out-of-scope file: drop `test_vod_namer_series_name_max.py` from this branch so the PR contains only the delete_user regression test it describes.
2. Redundancy / merge conflict: the delete_user test here is byte-identical to the one bundled in PR #248 (which adds the same test plus the actual fix). Whichever of #246/#248 lands second will conflict on the new file. Since #248 carries the fix and the test together, the cleanest path is to close this PR in favor of #248 once #248 is corrected. Same duplicate-file conflict exists between this PR and #247 for the vod_namer test.

The delete_user test content itself is fine as a failing-regression spec. The issue is purely scope and overlap.

razzamatazm:
Closing in favor of #248, which contains the same delete_user regression test plus the actual fix. The out-of-scope vod_namer test file has been removed from that branch.

PR #247, opened 2026-06-08
fix: truncate combined series episode base_name to 200 UTF-8 bytes

razzamatazm:
## Summary

- Series episodes with long CJK or ASCII show names combined with long episode titles could produce filenames exceeding Linux NAME_MAX (255 bytes), causing `OSError: [Errno 36] File name too long` at download time.
- Each component was independently capped at 200 bytes by `sanitize_filename`, but the assembled `base_name` in `series_episode_output_path` was never capped, reaching up to ~400 bytes.
- Fixed by truncating `base_name` to 200 UTF-8 bytes after assembly (preserving whole multi-byte characters) before appending the extension.

**File:** `backend/services/vod_namer.py`, `series_episode_output_path`

**Before:** A Korean show name (195 bytes) + Korean episode title (189 bytes) produced a 400-byte filename. Download failed with `OSError: [Errno 36] File name too long` at file open time.

**After:** Same inputs produce a 200-byte base_name filename that fits within NAME_MAX.

## Test plan

```
cd backend && venv/bin/python -m unittest tests.test_vod_namer_series_name_max -v
```

3 tests: CJK long show+title, ASCII long show+title, long show with no title. All pass after fix.

Before fix: 3 failures. After fix: 3 pass.

### DIFF
--- a/backend/services/vod_namer.py
+++ b/backend/services/vod_namer.py
@@ -58,6 +58,13 @@ def series_episode_output_path(
         # output paths. Leave normal numbered episodes (S02E05, etc.) clean.
         base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
 
+    # Cap combined base_name so the filename stays under Linux NAME_MAX (255
+    # bytes). Each component is independently capped at 200 bytes by
+    # sanitize_filename, but two components together can reach ~400 bytes.
+    encoded = base_name.encode("utf-8")
+    if len(encoded) > 200:
+        base_name = encoded[:200].decode("utf-8", errors="ignore").rstrip() or "unknown-episode"
+
     filename = f"{base_name}.{_safe_extension(extension)}"
 
     return os.path.join(download_folder, safe_show, season_folder, filename)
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approving.

Correctness: the cap truncates the assembled `base_name` to 200 UTF-8 bytes after both components are joined, so the filename is at most 200 + 1 + extension (<=8) bytes, comfortably under NAME_MAX (255). `encoded[:200].decode("utf-8", errors="ignore")` discards any partial multi-byte char at the boundary so no mojibake, `.rstrip()` trims a trailing space left by a cut, and the `or "unknown-episode"` fallback covers the degenerate empty case. The SxxExx prefix sits at the front of `base_name`, so for normal show names it is preserved and only the long tail is trimmed. The directory components (`safe_show`, `season_folder`) are separate path segments already capped by sanitize_filename, so each stays under NAME_MAX. No regression to the short-name path.

Tests: real, unmocked calls to `series_episode_output_path` asserting basename byte length for CJK long, ASCII long, and long-show-no-title cases. Fails before, passes after. Good coverage.

Scope is clean (one function + its test). One note, not blocking: `test_vod_namer_series_name_max.py` is duplicated in PR #246; whichever merges second will conflict on the new file. That is #246 scope contamination, not a problem with this PR. No em dashes, plain-English body with before/after evidence.

PR #248, opened 2026-06-08
fix: cancel orphaned schedules and downloads on user delete

razzamatazm:
## What changed

When an admin deleted a download-only user, Mustarrd left that user's scheduled recordings and active downloads running. On the next scheduler poll (within 30 seconds), orphaned schedules would fire and start new downloads attributed to the deleted user, using disk space and IPTV stream slots. In-progress downloads kept running. Open browser connections (WebSockets) were never closed.

`delete_account` was fixed for the same pattern previously, but `delete_user` (used by the Users admin page) was not updated.

## Fix

`delete_user` in `backend/api/admin_users.py` now, before removing the user row:

1. Marks all non-terminal scheduled recordings (SCHEDULED, QUEUED, DOWNLOADING, PROCESSING, PAUSED_LOW_SPACE) as CANCELLED with the message "User deleted".
2. Calls `download_manager.cancel_download` for any active downloads (PENDING, DOWNLOADING, PROCESSING).
3. Calls `download_manager.disconnect_user_websockets` to close open browser connections.

## Before

```
# delete_user cleaned up UserIdentity and UserSetupToken, then:
await session.delete(user)
await session.commit()
# Orphaned schedules, downloads, and WebSocket connections remained.
```

## After

```
# Scheduled recordings marked CANCELLED before user row is removed.
# Active downloads cancelled via download_manager.
# WebSocket connections closed.
await session.delete(user)
await session.commit()
```

## Test plan

```
cd backend && venv/bin/python -m unittest tests.test_delete_user_schedule_cleanup -v
```

Before fix: 9 failures. After fix: 7 tests (9 sub-cases) pass.

Full suite: 614 tests, all pass.

```
venv/bin/python -m unittest discover -s tests
Ran 614 tests in 2.474s
OK
```

Closes PR #246 (failing regression tests).

### DIFF
--- a/backend/api/admin_users.py
+++ b/backend/api/admin_users.py
@@ -9,7 +9,11 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from auth import require_admin, hash_password, AuthContext
 from database import get_session
-from models import User, UserIdentity, UserSetupToken
+from models import (
+    Download, DownloadStatus,
+    ScheduledRecording, ScheduledStatus,
+    User, UserIdentity, UserSetupToken,
+)
 from services.download_manager import download_manager
 
 
@@ -140,6 +144,35 @@ async def delete_user(
 
     await session.execute(delete(UserIdentity).where(UserIdentity.user_id == user.id))
     await session.execute(delete(UserSetupToken).where(UserSetupToken.user_id == user.id))
+
+    non_terminal = {
+        ScheduledStatus.SCHEDULED.value,
+        ScheduledStatus.QUEUED.value,
+        ScheduledStatus.DOWNLOADING.value,
+        ScheduledStatus.PROCESSING.value,
+        ScheduledStatus.PAUSED_LOW_SPACE.value,
+    }
+    sched_result = await session.execute(
+        select(ScheduledRecording).where(ScheduledRecording.requested_by_user_id == user.id)
+    )
+    for sched in sched_result.scalars().all():
+        if sched.status in non_terminal:
+            sched.status = ScheduledStatus.CANCELLED.value
+            sched.status_message = "User deleted"
+
+    active_dl_statuses = {
+        DownloadStatus.PENDING.value,
+        DownloadStatus.DOWNLOADING.value,
+        DownloadStatus.PROCESSING.value,
+    }
+    dl_result = await session.execute(
+        select(Download).where(Download.requested_by_user_id == user.id)
+    )
+    for dl in dl_result.scalars().all():
+        if dl.status in active_dl_statuses:
+            await download_manager.cancel_download(dl.id)
+
+    await download_manager.disconnect_user_websockets(user.id)
     await session.delete(user)
     await session.commit()
     return {"status": "deleted"}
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Requesting changes: ordering bug that the mirrored pattern was specifically written to avoid.

**Correctness (blocking): `cancel_download` is called mid-transaction, before commit.**

`download_manager.cancel_download` opens its own session via `async_session_maker()` and commits inside it (download_manager.py:1480). In this fix it is awaited *inside* `delete_user`, before `session.delete(user)` / `session.commit()`. The session has `autoflush` on (default), so the moment the `select(Download)` runs, the pending `sched.status = CANCELLED` mutations are flushed as an UPDATE, opening a write transaction on the outer connection. `cancel_download` then runs on a second pooled connection and tries to commit its own write. SQLite allows only one writer and the default busy timeout is 0, so this raises `OperationalError: database is locked` whenever a deleted user has BOTH a non-terminal schedule and an active download (PENDING/DOWNLOADING/PROCESSING). That is exactly the case this PR targets.

The reviewed-safe `delete_account` in accounts.py avoids this deliberately: it collects the active download IDs, calls `session.delete(...)` + `session.commit()` FIRST, and only then loops `await download_manager.cancel_download(dl_id)`. Please mirror that order exactly:
1. mark non-terminal schedules CANCELLED (in-session),
2. collect active download IDs into a list,
3. `session.delete(user)` + `session.commit()`,
4. then `for dl_id in ids: await download_manager.cancel_download(dl_id)`,
5. `await download_manager.disconnect_user_websockets(user.id)` (the disable path in update_user also calls it after commit).

**Tests (blocking): the regression cannot catch this.**

The tests fully mock the session (`AsyncMock` with a fixed `side_effect` list) and patch `download_manager`, so `cancel_download` is a no-op and no real commit ordering is exercised. They pass regardless of the lock bug. Please add a test that runs against a real (in-memory SQLite) session so the commit-before-cancel ordering is actually verified, or at minimum assert the call order (delete/commit before cancel_download).

Once the ordering matches delete_account and a test exercises a real session, this is good to go. The set of non-terminal/active statuses and the "User deleted" message are correct.

razzamatazm-cleo (approved):
Approve. Both blocking concerns from the prior review are resolved.

- **Ordering fixed.** Now mirrors `delete_account` exactly: mark non-terminal schedules CANCELLED in-session, collect active download IDs, `session.delete(user)` + `session.commit()` FIRST, then loop `cancel_download`, then `disconnect_user_websockets`. No second writer opens while the outer write txn is uncommitted, so the SQLite "database is locked" path is gone.
- **Order is now tested.** `test_commit_before_cancel_download` asserts commit precedes cancel_download, which was the minimum I asked for.
- **`user.id` after delete+commit is safe.** `async_session_maker` sets `expire_on_commit=False` (database.py:25), so the deleted object's id stays cached and accessing it post-commit triggers no refresh SELECT against the gone row.

Verified: 8 new tests pass, full suite 615 OK. Status sets (non-terminal schedules, active downloads) and the "User deleted" message are correct.

PR #250, opened 2026-06-08
docs: changelog entries for PRs #247 and #249

razzamatazm:
## What changed

Adds changelog entries for two PRs merged on 2026-06-08 that had no documentation yet.

**PR #247** (fix: truncate combined series episode base_name to 200 UTF-8 bytes): Series episodes where the show name and episode title were both long could produce a combined filename exceeding Linux's 255-byte limit, causing a silent ENAMETOOLONG failure at download time. Each part was already individually capped but the assembled result was not. Entry explains what users would notice and what changed, without technical jargon.

**PR #249** (Design: inline Retry button on failed and cancelled cards): The Retry option was previously only accessible from the three-dot action menu, which is easy to miss on mobile. Entry explains that the button now appears directly on the card.

## Test

Verified no em dashes. Both entries are in the 2026-06-08 date section at the top of the changelog, consistent with existing style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Improved: Retry button now appears directly on failed and cancelled download cards
+
+**What you would notice:** When a download failed or was cancelled, the only way to retry it was to open the three-dot action menu on the card and find the Retry option. That menu is easy to miss, especially on a phone. The Retry button now appears directly on the card itself, right alongside the other action buttons, so you can act immediately without hunting through a menu.
+
+**What changed:** A Retry button was added inline to failed and cancelled download cards in the History tab. No download logic was changed. The button uses the same retry action that was already available in the menu.
+
+---
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Verified both changelog entries against merged code on origin/main.

- #247 (filename trim): backend/services/vod_namer.py caps the combined base_name to 200 UTF-8 bytes after assembly (encoded[:200].decode(errors="ignore").rstrip()), before the extension. Matches the entry, including preserving whole multibyte characters.
- #249 (Retry button): frontend/src/pages/Downloads.jsx adds an inline Retry button gated on canRetry = ['failed','cancelled'].includes(status), calling the same onRetry as the menu item. UI-only, no download logic changed. Matches the entry.

Pure docs PR. No em dashes (the --- lines are markdown horizontal rules consistent with the rest of CHANGELOG). Plain-English and non-programmer friendly.

Optional nit, not blocking: the '#247' entry's 'more than about 200 characters' is really 200 UTF-8 bytes combined (~66 CJK chars per part); the entry clarifies bytes further down, so it reads fine.

Approved.

PR #251, opened 2026-06-08
test: failing regression for expired-schedule FAILED status not committed

razzamatazm:
## Summary

- Adds a failing regression test for the bug filed in [#tasks](https://discord.com/channels/1512584937637728266/1512585829921591307/1513645220364026119): expired schedule FAILED status never committed when no ready schedules exist.
- The test fails before the fix and passes after.

## Bug

`_queue_ready_recordings` marks expired schedules `FAILED` in memory, then hits `if not ready: return` (line 82 of `scheduled_manager.py`) before `session.commit()` is called. SQLAlchemy discards uncommitted mutations when the `async with session:` block exits, so the schedule stays `SCHEDULED` in the database forever. The UI shows the recording as permanently "Scheduled" with no indication it is missed.

## What the test does

Arranges a single expired schedule (program ended 48 hours ago on a 1-day catchup window), runs `_queue_ready_recordings`, and asserts that `session.commit()` was called exactly once. Currently `commit` is called 0 times, so the test fails.

## Test plan

Before fix:
```
python3 -m unittest tests.test_scheduler_stale_dispatch.StaleScheduleDispatchTests.test_expired_schedule_failed_status_committed
FAILED (failures=1) -- AssertionError: Expected 'commit' to be called once. Called 0 times.
```

All 5 existing tests in the file still pass.

### DIFF
--- a/backend/services/vod_namer.py
+++ b/backend/services/vod_namer.py
@@ -58,6 +58,13 @@ def series_episode_output_path(
         # output paths. Leave normal numbered episodes (S02E05, etc.) clean.
         base_name = f"{base_name} - {_sanitize_component(str(episode_id))}"
 
+    # Cap combined base_name so the filename stays under Linux NAME_MAX (255
+    # bytes). Each component is independently capped at 200 bytes by
+    # sanitize_filename, but two components together can reach ~400 bytes.
+    encoded = base_name.encode("utf-8")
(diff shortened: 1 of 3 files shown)


razzamatazm:
PR #252 includes this regression test (cherry-picked) plus the fix. Closing this in favor of #252.

PR #253, opened 2026-06-08
[Design] linkify provider error on Downloads and Scheduled history cards

razzamatazm:
## What a user notices

On the Downloads and Scheduled history pages, when a recording fails because the provider is unreachable, the error box reads:

> Cannot reach the provider. Check the server URL in your account settings.

Before this change, that last phrase was plain text. A non-technical user had to know to go to Settings themselves. The Browse page already renders this same message with a clickable link, but the history cards did not.

## What changed

"your account settings" is now an orange clickable link that takes the user directly to Settings → Accounts. Two files changed: `Downloads.jsx` and `Scheduled.jsx`. No logic changes.

## Test plan

- Opened Downloads History: EastEnders FAILED card shows orange "your account settings" link, clicking navigates to Settings → Accounts.
- Opened Scheduled History: EastEnders FAILED card shows the same link.
- Other error messages (e.g. "Connection timed out after 30 seconds", "Provider returned error page instead of video stream") are unaffected.
- Desktop and mobile both verified via screenshots below.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -48,7 +48,7 @@ import { formatChannelDateTime, formatAirDateTime, getGuideOffsetHours } from '.
 
 dayjs.extend(duration)
 
-const ACCOUNT_SETTINGS_SUFFIX = 'in your account settings.'
+const ACCOUNT_SETTINGS_SUFFIX = 'your account settings.'
 
 function renderErrorMessage(msg) {
   if (!msg) return null
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -18,7 +18,7 @@ import {
   Select,
 } from '@mantine/core'
 import { notifications } from '@mantine/notifications'
-import { useNavigate } from 'react-router-dom'
+import { useNavigate, Link } from 'react-router-dom'
 import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query'
 import {
   IconDotsVertical,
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Requesting changes: the linkify drops a word from the rendered message.

**Correctness (blocking): the word "in" disappears.**
`ACCOUNT_SETTINGS_SUFFIX = 'in your account settings.'` and the link text is `your account settings`, but `msg.slice(0, idx)` cuts the string right before `in`. So `in` is in neither the prefix nor the link.

Backend string (`download_manager.py:30`): `"Cannot reach the provider. Check the server URL in your account settings."`
Renders as: `Cannot reach the provider. Check the server URL ` + [link "your account settings"] + `.`
= "...server URL your account settings." -- the "in" is gone.

Fix: either include it in the prefix, e.g. `msg.slice(0, idx + 3)` to keep "in ", or make the link text "in your account settings".

**Coverage gap (please address): the other provider error is never linkified.**
`download_manager.py:41` returns `"Invalid credentials or not authorized. Check your account settings."` -- that is "Check your account settings", not "in your account settings", so `indexOf` returns -1 and it renders as plain text with no link. The 401/403 case is exactly the one where the user most needs to reach account settings. Consider matching the shorter "your account settings." suffix so both errors linkify.

**Minor:** `renderErrorMessage` is duplicated verbatim in Downloads.jsx and Scheduled.jsx. A shared helper would avoid drift, though I won't block on it given minimal-diff.

Once "in" is preserved and ideally both backend strings linkify, this is good.

razzamatazm (commented):
Fixed. Changed suffix to 'your account settings.' so the 'in' stays in the prefix and both the 404 and 401/403 error strings now linkify. Two-line change across both files.

razzamatazm-cleo (approved):
Approve. Re-review after the push.

Confirmed against the two backend strings in download_manager.py:
- L30 "...Check the server URL in your account settings." -> renders "...in [your account settings]." with the word "in" preserved in the prefix. Original concern fixed.
- L41 "Invalid credentials or not authorized. Check your account settings." -> now linkifies too, since the suffix is matched on "your account settings.".

renderErrorMessage splits on the shared suffix in both Downloads.jsx and Scheduled.jsx, returns the string unchanged when the suffix is absent, and null-guards empty messages. Link target /settings?section=accounts matches the accounts section. No concerns.

PR #254, opened 2026-06-08
docs: README reflects Retry button on cancelled downloads

razzamatazm:
## What a user would notice

The README said "Failed downloads can be retried from the same page." After PR #249 shipped, the Retry button also appears directly on cancelled download cards. The old wording missed cancelled downloads entirely, and separately noted they could be retried in a different bullet -- which was redundant and slightly confusing.

## What changed

Two lines in the Monitor Downloads section of the README were updated:

- Before: "Failed downloads can be retried from the same page"
- After: "Failed and cancelled downloads show a **Retry** button directly on the card"

The parenthetical note about cancelled entries being kept for retry was rephrased to match ("are not cleared so you can still retry them") but remains in place.

## How it was tested

Diff reviewed against PR #249 (merged) to confirm the new wording matches what shipped.

No code was changed.


razzamatazm-cleo (approved):
Approve. Docs-only, both claims verified against code:
- "Failed and cancelled downloads show a Retry button directly on the card": Downloads.jsx L128 canRetry = ['failed','cancelled'].includes(status), Retry button rendered at L175-180. Matches.
- "cancelled entries are not cleared so you can still retry them": clear_finished_downloads (backend/api/downloads.py L412-414) deletes only COMPLETED and FAILED. Cancelled survive. Matches.
No em dashes, plain English. Good.

PR #255, opened 2026-06-08
fix: show ComSkip settings panel when onboarding sets remux_comskip profile

razzamatazm:
## What a user notices

On Docker installs where ComSkip is detected, the onboarding wizard recommends and applies the **MKV container + skip commercials** profile. After completing onboarding, opening **Settings > Post-Processing** showed **MKV container (fast, no re-encode)** in the dropdown and no ComSkip settings panel below it. There was no way to see or change the ComSkip binary path or .ini path. Saving any setting (such as updating the ComSkip binary path) silently downgraded the profile from fast stream-copy + commercial removal to slow full re-encode + commercial removal.

## What changed

**Frontend (Settings.jsx):** `getRecordingFormat` checked `remux_only` before `comskip_enabled`, so the database state `remux_only=True, comskip_enabled=True` mapped silently to the plain `remux_mkv` format. Three changes:
1. `getRecordingFormat` now checks `remux_only && comskip_enabled` first and returns `'remux_mkv_comskip'`.
2. A new **MKV container + skip commercials (fast)** option is added to the dropdown.
3. The ComSkip binary path and .ini path panel renders for all three ComSkip formats.

**Backend (api/settings.py):** `update_settings` forced `remux_only=False` whenever `comskip_enabled=True`. This clobbered the fast profile on every settings save. The pipeline has supported stream-copy commercial removal since the `remove_commercials` rewrite: it extracts segments with `-c copy`, then concatenates with `-c copy` when `remux_only=True`. The constraint is removed. The `transcode_enabled=True` constraint stays (FFmpeg is still required).

**Tests (test_comskip_constraint.py):** Updated tests that asserted the old `remux_only=False` enforcement to now assert that `remux_only=True + comskip_enabled=True` is a valid, preserved state. Added a regression test: the remux_comskip profile survives any settings save without being silently downgraded.

## Before / after

**Before:** Docker user completes onboarding with ComSkip present. Settings shows "MKV container (fast, no re-encode)". No ComSkip panel. Saving any setting with this profile active silently converts it to full re-encode.

**After:** Settings shows "MKV container + skip commercials (fast)". ComSkip binary path and .ini path fields appear. Saving settings preserves the profile.

## How it was tested

Traced through `getRecordingFormat` with the onboarding state (`remux_only=True, comskip_enabled=True`): old code returned `'remux_mkv'`; new code returns `'remux_mkv_comskip'`.

Backend: all 608 tests pass, including the updated `test_comskip_constraint.py` with 8 tests.

Verified `remove_commercials` in `post_processor.py`: with `remux_only=True`, it extracts keep segments with `-c copy` then concatenates with `-c copy -avoid_negative_ts make_zero`. EDL cut points ARE applied. The old constraint was stale.

### DIFF
--- a/backend/api/settings.py
+++ b/backend/api/settings.py
@@ -228,12 +228,13 @@ async def update_settings(
             value = int(value)
         setattr(settings, field, value)
 
-    # Enforce ComSkip constraints on the final stored state, not just the
-    # current request, so a subsequent request cannot disable transcode or
-    # enable remux-only while ComSkip is already on.
+    # Enforce ComSkip constraint on the final stored state: comskip requires
+    # FFmpeg, so transcode_enabled must stay True.  remux_only is NOT forced
+    # off: the pipeline supports stream-copy commercial removal (segment
+    # extraction + concat with -c copy), so remux_only=True + comskip_enabled=True
+    # is the valid "fast remux + skip commercials" profile written by onboarding.
     if settings.comskip_enabled:
         settings.transcode_enabled = True
-        settings.remux_only = False
 
     await session.commit()
     await session.refresh(settings)
--- a/backend/tests/test_comskip_constraint.py
+++ b/backend/tests/test_comskip_constraint.py
@@ -1,8 +1,6 @@
 """
 Regression tests for ComSkip settings constraints.
 
-Two related bugs:
-
 Bug 1 (HIGH): update_settings only enforced "comskip_enabled=True forces
 transcode_enabled=True" when the current request contained comskip_enabled.
 A second request setting transcode_enabled=False bypassed the constraint,
@@ -10,13 +8,11 @@ leaving comskip_enabled=True + transcode_enabled=False in the DB. ComSkip
 ran and produced an EDL that was silently ignored; commercials stayed in
 the output with no user-visible error.
 
-Bug 2 (MEDIUM): No constraint enforced "comskip_enabled=True forces
-remux_only=False". A user could store comskip_enabled=True + remux_only=True.
-FFmpeg ran in stream-copy mode and never applied EDL cut points. All
-commercial segments remained in the output file while the UI showed COMPLETED.
-
-Fix: constraints now apply to the final stored state after all field updates,
-not just to the fields present in the current request body.
+Fix: transcode_enabled=True is enforced on the final stored state after all
+field updates.  remux_only is intentionally NOT constrained: the pipeline
+supports stream-copy commercial removal (segment extraction + concat -c copy),
+so remux_only=True + comskip_enabled=True is the valid fast "remux + skip
+commercials" profile written by the Docker onboarding flow.
 """
 import sys
 import unittest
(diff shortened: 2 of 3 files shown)


razzamatazm-cleo (changes requested):
The panel-visibility half of this is right: getRecordingFormat now maps the onboarding-written state (remux_only=true, comskip_enabled=true) to remux_mkv_comskip and the ComSkip path panel renders for it. That part is good and useful.

The blocker is the round-trip. The backend forces remux_only=false whenever comskip is enabled:

backend/api/settings.py:233-235
    if settings.comskip_enabled:
        settings.transcode_enabled = True
        settings.remux_only = False

So any save through update_settings with comskip on clears remux_only. Two consequences:

1. The new selectable dropdown option does not persist. A user picks "MKV container + skip commercials (fast)" -> frontend sends remux_only=true, comskip_enabled=true -> backend stores remux_only=false -> on reload getRecordingFormat returns transcode_comskip and the dropdown shows "MKV + skip commercials (re-encode)". The user silently gets full re-encode (slow, lossy) instead of the fast remux they chose.

2. The exact users this PR targets hit the same trap. A Docker onboarding user opens this panel to set the comskip binary path (the whole point), clicks Save, and their fast remux+comskip profile is silently downgraded to transcode_comskip on the next recording.

At the recording level remux_only=true + comskip_enabled=true is a real, supported fast path (download_manager.py:1311-1313 honors both; post_processor.remove_commercials takes remux_only and uses ffmpeg copy). So the backend pipeline supports this mode; only the settings.py write-time constraint blocks it.

Because both the new option and the onboarding save path silently convert fast remux to full re-encode, this frontend change cannot ship safely on its own. Please pair it with the backend constraint fix (TODO 1513234256736944258) so update_settings preserves remux_only when comskip is on, in the same PR or as a prerequisite. Once the backend round-trips remux+comskip, this frontend change is correct.

Minor, non-blocking: relabeling transcode_comskip -> "MKV + skip commercials (re-encode)" and transcode_mp4_comskip -> "MP4 + skip commercials (re-encode)" is a clearer wording change; fine to keep.

razzamatazm-cleo (approved):
APPROVED (re-review after backend fix added).

Verified end-to-end, not just the diff:
- backend/services/post_processor.py remove_commercials: keep segments are extracted with -ss/-to -c copy (L1258-1262), so EDL cut points ARE applied even when remux_only=True. remux_only only selects -c copy vs re-encode on the final concat (L1303-1311). My earlier concern (stream-copy would skip cuts) was wrong: the cut happens at segment extraction, always.
- backend/services/download_manager.py L1386-1424: detect_commercials -> remove_commercials(remux_only=remux_only) -> commercials_removed=True skips the later transcode. The remux+comskip path is real and complete.
- api/settings.py: dropping the forced remux_only=False is correct; transcode_enabled=True constraint kept (FFmpeg still required).
- Tests call the real update_settings against in-memory SQLite (no mocking the constraint away). New test_remux_comskip_profile_survives_settings_save fails before the fix (old code forced remux_only=False) and passes after. Good regression coverage.
- Frontend getRecordingFormat checks remux_only && comskip_enabled before plain remux_only; FORMAT_TO_SETTINGS round-trips; isComskipFormat shows panel + warning.

Non-blocking note for a later pass: there is no remux_mp4_comskip option, so remux_only+comskip+transcode_format=mp4 maps to MKV. Onboarding only writes MKV so this is not a regression. Fine to ship.

PR #257, opened 2026-06-08
docs: changelog entries for PRs #248, #252, and #253

razzamatazm:
## What this does

Adds three missing changelog entries for fixes merged earlier today.

## What a user would notice

**Before:** The changelog had no entries for three recent fixes. Users reading the changelog after updating would not know what changed.

**After:** Three new entries explain the fixes in plain English.

## Entries added

- **PR #248** (Fixed: Deleting a user now immediately cancels their scheduled recordings and active downloads): Admins deleting a user from the Users page would leave that user's scheduled recordings firing and active downloads running. The changelog entry explains the before/after behavior for an operator who manages users.

- **PR #252** (Fixed: Expired scheduled recordings now correctly show as Failed instead of staying Scheduled forever): Recordings past the provider catchup window would stay stuck as "Scheduled" forever. The entry explains the symptom and what changed.

- **PR #253** (Improved: Provider error on failed recordings now links directly to Account Settings): The provider-unreachable error on history cards now links to Settings > Accounts. The entry explains the plain-text-to-link change.

## How it was tested

- No em dashes in any new text.
- Wording reviewed against the PR bodies and code changes for accuracy.
- All three entries follow the established format (What you would notice / What changed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-08
 
+### Improved: Provider error on failed recordings now links directly to Account Settings
+
+**What you would notice:** When a recording failed because Mustarrd could not reach your IPTV provider, the error message said "Cannot reach the provider. Check the server URL in your account settings." The phrase "your account settings" was plain text with no way to click it. You had to know where to go on your own. That phrase is now a clickable link that takes you directly to Settings > Accounts, where you can correct the server URL or credentials. The same error on the Browse page already had this link; now the Downloads and Scheduled history pages match it.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
APPROVED.

Docs-only. Verified each entry against the actual merged file sets:
- #253: only Downloads.jsx + Scheduled.jsx changed; entry's 'two frontend files, no backend code' claim is accurate.
- #252: scheduled_manager.py one-line commit-before-early-return; entry matches.
- #248: admin_users.py cancels schedules/downloads/sessions on delete, matching self-delete path; entry matches.
No em dashes. Entries under correct 2026-06-08 date. Changelog needs no regression test.

PR #259, opened 2026-06-08
design: Guide sync status and refresh in Settings

razzamatazm:
## What changed

Added a **Guide** section to Settings (between File Naming and Appearance). It shows when the program guide was last synced and lets the user trigger a refresh without navigating away.

**Before:** No way to see guide sync status from Settings. Users had no way to know if their guide data was stale without going to Accounts and using "Force EPG Refresh."

**After:** Settings > Guide shows "Last synced: Today at 10:30 PM" and a "Refresh Now" button. The button shows a spinner while running and disables to prevent double-clicks. If the guide has never synced, shows "Guide not yet synced."

## How it was tested

- Started the app locally, navigated to Settings > Guide. "Last synced" shows the correct time.
- Guide section appears in the sidebar between File Naming and Appearance at desktop width.
- Mobile layout renders cleanly with the dropdown section selector.
- Build passes (vite build --mode development).

## Notes

- Frontend-only. Uses existing GET /api/epg/status and POST /api/epg/refresh.
- Reuses the ['epg', 'status'] query key shared with AccountsSection.
- Not included in configSections so the unsaved-changes guard does not apply.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -44,9 +44,11 @@ import {
   IconTrash,
   IconKey,
   IconDeviceTv,
+  IconCalendar,
+  IconRefresh,
 } from '@tabler/icons-react'
 
-import { adminPlexApi, adminUsersApi, authApi, settingsApi } from '../api'
+import { adminPlexApi, adminUsersApi, authApi, epgApi, settingsApi } from '../api'
 import AccountsSection from '../components/settings/AccountsSection'
 import LogsSection from '../components/settings/LogsSection'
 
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Frontend-only, logic verified end to end.

- `epgApi.status()` hits GET /api/epg/status and `epgApi.refresh(false)` POSTs `{force:false}`, both already defined in api.js.
- `epg_ingest_manager.get_status()` returns `running` and `last_completed_at` (ISO 8601 UTC), so `epgStatus?.running` and `new Date(last_completed_at)` are correct; UTC parses and renders in local time as intended.
- Reuses the `['epg','status']` query key, invalidates on success, button is `loading`/`disabled` while running so double-clicks are blocked; a concurrent refresh returns 409 and surfaces via onError. Sound.
- No frontend test framework exists, consistent with PR #256/#249 precedent. No em dashes.

Minor, non-blocking: `refetchInterval: 5000` runs while any admin Settings section is open, not just Guide. Harmless polling, fine to leave.

PR #260, opened 2026-06-08
fix: preserve channel category through schedule creation and dispatch

razzamatazm:
## What a user would notice

When scheduling a recording on a "Movies" or "Sports" channel and leaving the filename field blank, the recorded file got the wrong name. For example, a movie called "The Dark Knight" on a Movies channel ended up named like `The Dark Knight - 2024-01-15.mkv` instead of `The Dark Knight (2008).mkv`, because Mustarrd forgot which channel category it came from by the time it fired the download.

Users who type a custom filename in the Schedule dialog are not affected.

## What changed

`ScheduledRecording` now stores a `channel_category_name` column (added via the usual startup migration). When you schedule a program, the category from the EPG guide entry is saved. When the scheduler fires the recording, that category is included in the program info passed to the filename generator, so `detect_program_type` can correctly classify the show as a movie, sports event, etc.

## How it was tested

New regression test `test_schedule_dispatch_category.py`:

- Confirms that `program["category"]` passed to `build_download_from_program` equals the stored `channel_category_name` (`"Movies"`).
- Confirms that a `None` category is sent as `""` (not `None`), which is what `detect_program_type` expects.

Both tests fail without the fix and pass after. Full suite: 610 tests, all pass.

```
Ran 610 tests in 2.501s
OK
```

### DIFF
--- a/backend/api/schedules.py
+++ b/backend/api/schedules.py
@@ -201,10 +201,15 @@ async def create_schedule(
 
     custom_filename = _sanitize_filename(data.custom_filename)
 
+    channel_category_name = (
+        data.program.get("category") or data.program.get("category_name") or None
+    )
+
     schedule = ScheduledRecording(
         account_id=data.account_id,
         channel_id=data.channel_id,
         channel_name=data.channel_name,
+        channel_category_name=channel_category_name,
         program_id=str(program_id) if program_id is not None else None,
         epg_id=str(epg_id) if epg_id is not None else None,
         program_title=data.program.get("title", "Unknown"),
--- a/backend/database.py
+++ b/backend/database.py
@@ -78,6 +78,9 @@ async def _apply_lightweight_migrations(conn) -> None:
             text("ALTER TABLE xtream_accounts ADD COLUMN guide_offset_hours INTEGER DEFAULT 0")
         )
 
+    if not await _column_exists(conn, "scheduled_recordings", "channel_category_name"):
+        await conn.execute(text("ALTER TABLE scheduled_recordings ADD COLUMN channel_category_name VARCHAR(255)"))
+
     if not await _column_exists(conn, "scheduled_recordings", "provider_start"):
         await conn.execute(text("ALTER TABLE scheduled_recordings ADD COLUMN provider_start VARCHAR(255)"))
 
--- a/backend/models/scheduled_recording.py
+++ b/backend/models/scheduled_recording.py
@@ -25,6 +25,7 @@ class ScheduledRecording(Base):
     # Channel info
     channel_id: Mapped[str] = mapped_column(String(100))
     channel_name: Mapped[str] = mapped_column(String(255))
+    channel_category_name: Mapped[str | None] = mapped_column(String(255), nullable=True)
 
     # Program info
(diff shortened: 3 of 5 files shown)


razzamatazm-cleo (changes requested):
Request changes: scope contamination from a stacked branch.

The channel-category fix itself (commit b5145778) is correct and well tested. At schedule creation the category is stored, at dispatch `_queue_ready_recordings` passes `"category": schedule.channel_category_name or ""`, and `build_download_from_program` reads `program.get("category", "")`, so the keys line up and the regression test in `test_schedule_dispatch_category.py` fails before and passes after. The migration is idempotent and nullable. No issue with that commit.

The problem is the PR as presented. Base is `main` but this branch was cut from #255's branch, so the diff against main also carries both of #255's commits:

- 4c1e1dff "show ComSkip settings panel when onboarding sets remux_comskip profile" (Settings.jsx: new `remux_mkv_comskip` format, alert/comskip-format gating)
- 82c31fc3 "preserve remux_only when comskip is enabled in update_settings" (settings.py drops the forced `remux_only = False`, full rewrite of test_comskip_constraint.py)

None of that belongs to "preserve channel category." It violates "no unrelated changes bundled in; UI and logic kept separate," and the PR body does not disclose the dependency on #255. Concretely: #255 is still pending Tyler, and merging #260 against main would silently ship all of #255's ComSkip behavior under a different title, unreviewed in this context, and possibly out of the intended order.

Fix: make the stack explicit. Either retarget this PR's base to `clem/fix-comskip-settings-panel-remux` so the diff shows only the channel-category delta and GitHub enforces #255 merging first, or wait for #255 to merge and rebase this onto `main` so it contains only b5145778. After that I will re-review the isolated diff.

One question for when it is isolated: confirm the frontend Schedule dialog actually puts `category` (or `category_name`) into the `program` payload it POSTs. The plumbing is correct, but if the dialog never sends that key the stored value is always None and the fix is inert in practice. A quick note or a test asserting the create path stores "Movies" would close that gap.

razzamatazm:
Rebased onto main — the two #255 commits are now upstream and the diff is isolated to the channel-category fix (one commit).

Re: your question about the frontend payload: yes, `category` is present. `ScheduleModal.jsx:140` passes `program: program` (the raw EPG object) to POST /api/schedules. The backend's `epg_service.serialize_program` includes `"category"` at line 330, so `data.program.get("category")` in the schedule create path finds it. A schedule created from the Browse EPG dialog will have `channel_category_name` correctly populated (not None). The regression test covers the create-to-dispatch round-trip but I can add a test asserting the stored value equals the input category if that would close the gap more explicitly.

razzamatazm-cleo (approved):
APPROVE. Re-reviewed at the rebased head (1a6e3c47). Scope contamination is resolved: the diff is now a single commit touching only the five category-related files, no #255 ComSkip commits. The fix correctly threads channel category from create_schedule (reads program category/category_name) through the stored nullable column to dispatch, where it is emitted as program['category'] for build_download_from_program to classify on. Migration adds a nullable VARCHAR(255) via the lightweight ALTER path. Null maps to empty string at dispatch, matching prior behavior for uncategorized channels. Regression test covers both the populated and null dispatch cases. Minimal diff, no em dashes.

PR #261, opened 2026-06-08
test: failing regression for ISO XMLTV timestamp silent drop

razzamatazm:
## What this tests

Providers that emit XMLTV `start`/`stop` in ISO 8601 format (`2024-01-15 20:00:00 +0100` or `2024-01-15T20:00:00+01:00`) hit the `else: return None` branch in `_parse_xmltv_time`. `_iter_programs` then silently skips every program for that account. The guide goes empty with no warning logged.

## Root cause

`_parse_xmltv_time` splits on the first space, so `"2024-01-15 20:00:00"` gives `date_part = "2024-01-15"` (10 chars). The function only handles 14-char and 12-char date parts (compact XMLTV spec format). The 10-char ISO date falls through to `return None`.

Same path for T-separator format: `"2024-01-15T20:00:00"` has no space, hits `len(value) >= 14` branch, slices `value[:14] = "2024-01-15T20:"`, then `strptime("2024-01-15T20:", "%Y%m%d%H%M%S")` raises ValueError, returns None.

## Test results

7 tests fail (all ISO format variants). 3 tests pass (compact format and existing error paths confirm no test setup issues).

```
Ran 10 tests in 0.003s
FAILED (failures=7)
```

## Fix hint

Add an ISO format branch in `_parse_xmltv_time` (e.g. using `datetime.fromisoformat()` with Python 3.7+ for the date-part when it contains dashes or a T separator).

PR #262, opened 2026-06-08
fix: parse ISO 8601 XMLTV timestamps (YYYY-MM-DD format)

razzamatazm:
## What a user would notice

If your IPTV provider sends program guide data using ISO-format dates like `2024-01-15 20:00:00 +0100` instead of the compact `20240115200000 +0100`, your entire Browse EPG was silently empty after every refresh. No error appeared in the logs or UI.

## What changed

`_parse_xmltv_time` in `epg_ingest_manager.py` now detects ISO-style timestamps (date part contains dashes or a `T` separator) and routes them through a separate normalization path before calling `datetime.fromisoformat`. The compact XMLTV format path is unchanged.

Supported ISO variants now parsed correctly:
- `2024-01-15 20:00:00` (space separator, no tz)
- `2024-01-15 20:00:00 +0100` (space separator, ±HHMM offset)
- `2024-01-15 20:00` (no seconds)
- `2024-01-15T20:00:00` (T separator)
- `2024-01-15T20:00:00+01:00` (T separator, colon offset)
- `2024-01-15T20:00:00Z` (Z suffix)

## How it was tested

Regression tests from Gremlin's PR #261 (7 failing tests) were cherry-picked onto this branch. All 7 now pass alongside the 3 existing compact-format tests. Full suite: 631 tests, all pass.

**Before fix:**
```
Ran 10 tests in 0.002s
FAILED (failures=7)
```

**After fix:**
```
Ran 10 tests in 0.001s
OK

Ran 631 tests in 2.777s
OK
```

Supersedes PR #261 (test-only). Tests are included in this PR so Gremlin's branch can be closed.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -601,6 +601,38 @@ class EPGIngestManager:
         if not value:
             return None
         value = value.strip()
+
+        # ISO format detection: date part contains dashes (YYYY-MM-DD) or T separator.
+        # Some providers emit "2024-01-15 20:00:00 +0100" or "2024-01-15T20:00:00+01:00"
+        # instead of the compact XMLTV-spec format "20240115200000 +0100".
+        first_chunk = value.split(" ", 1)[0] if " " in value else value
+        if "-" in first_chunk[:10] or "T" in first_chunk:
+            try:
+                parts = value.split()
+                if len(parts) == 3:
+                    # "2024-01-15 20:00:00 +0100" -> "2024-01-15T20:00:00+0100"
+                    iso_str = parts[0] + "T" + parts[1] + parts[2]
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
APPROVE. ISO 8601 XMLTV timestamp parsing is correct. Traced and empirically verified every branch: space-separator with/without offset, T-separator, Z suffix, no-seconds, negative offsets, and the +HHMM to +HH:MM normalization for Python < 3.11. The compact-format path is untouched and does not regress (compact with -0500 still parses). `timezone` is already imported at module top, so no NameError. Detection via `first_chunk[:10]` dash check cannot misfire on all-digit compact strings. Failures still return None as before. Regression tests are comprehensive and include a compact-still-works guard. No em dashes, minimal diff.

PR #263, opened 2026-06-08
[Design] add status filter to Scheduled Recordings history tab

razzamatazm:
## What a user would notice

The Scheduled Recordings History tab now has a status filter dropdown, matching the one already on the Downloads History tab. Users can narrow the list to show only Completed, Failed, or Cancelled scheduled recordings instead of scrolling a mixed list.

## What changed

Added an "All statuses" filter Select to the History panel in `Scheduled.jsx`. Selecting a status shows only matching cards; selecting "All statuses" restores the full list. If the filter returns no results, a plain-English message appears ("No failed scheduled recordings."). Frontend-only, one file.

## How it was tested

Loaded the Scheduled > History tab in the local dev environment, opened the filter dropdown, and confirmed each option correctly filters the card list. Verified the empty-state message appears when a filter matches zero items.

## Before

History tab showed all statuses mixed together with no way to filter.

## After

History tab has a filter dropdown at the top. Selecting "Failed" shows only failed recordings; selecting "Cancelled" shows only cancelled ones.

### DIFF
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -12,6 +12,7 @@ import {
   Loader,
   Alert,
   Button,
+  Select,
 } from '@mantine/core'
 import { notifications } from '@mantine/notifications'
 import { useNavigate, Link } from 'react-router-dom'
@@ -29,6 +30,7 @@ import {
   IconSettings,
   IconPlayerPlay,
   IconFolderOpen,
+  IconFilter,
 } from '@tabler/icons-react'
 import dayjs from 'dayjs'
 import duration from 'dayjs/plugin/duration'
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
APPROVE. Frontend-only status filter for Scheduled > History, mirroring the existing Downloads > History filter. Logic is clean: filteredHistorySchedules derives from historyFilter, the Select renders inside the non-empty history branch so users can always switch back, and the empty-result case shows a clear message. The 'No all' wording is unreachable because filter='all' yields the full non-empty list. allowDeselect={false} prevents a null value. No backend changes, no em dashes. Frontend-only with no test harness, consistent with prior design PR precedent (#256/#259).

PR #265, opened 2026-06-08
fix: check disk space before re-queuing downloads on restart

razzamatazm:
## What a user would notice

Before this fix: if the disk filled up while Mustarrd was stopped, any downloads that were waiting to start would restart automatically when Mustarrd came back up. They would then immediately fail with a confusing OS-level error (or silently corrupt the file) instead of a clear disk-space message.

After this fix: Mustarrd checks free disk space during startup recovery. If there is not enough space, those downloads are marked **Failed** right away with a plain message: "Disk space below minimum at startup (X.X GB free, Y GB required)." The user can free up disk and retry.

## What changed

`recover_incomplete_downloads` in `download_manager.py` now checks `shutil.disk_usage` against `min_free_space_gb` before putting recovered PENDING downloads back on the queue. This mirrors the same guard already in `create_download` (API) and `scheduled_manager`. Downloads going to the post-processing queue (already downloaded) are unaffected.

## How it was tested

Four regression tests in `backend/tests/test_recover_low_disk.py`:
1. Low disk at startup marks PENDING download FAILED with a disk-space message.
2. Sufficient disk re-queues normally (no regression).
3. Missing download folder skips the check and queues normally.
4. `min_free_space_gb=0` disables the guard and queues normally.

Before fix: test 1 failed (download stayed PENDING and was queued anyway).
After fix: 635 tests pass.

## Steps to test

1. Queue a download, stop Mustarrd before it starts.
2. Fill the disk past the `min_free_space_gb` threshold.
3. Restart Mustarrd.
4. The download should appear as Failed with "Disk space below minimum at startup" instead of restarting and crashing.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -595,6 +595,9 @@ class DownloadManager:
                                     f"Disk space below minimum at startup "
                                     f"({free_gb:.1f} GB free, {min_free_gb} GB required)."
                                 )
+                                if not dl.completed_at:
+                                    dl.completed_at = datetime.utcnow()
+                                await self._sync_schedule_status(session, dl.id, DownloadStatus.FAILED.value)
                                 await self._broadcast_log(
                                     dl_id,
                                     dl.error_message,
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -577,6 +577,31 @@ class DownloadManager:
                     await self._sync_schedule_status(session, download.id, DownloadStatus.FAILED.value)
                     await self._broadcast_log(download.id, download.error_message, level="error")
 
+            if recovered_download_ids:
+                min_free_gb = (
+                    settings.min_free_space_gb
+                    if settings and settings.min_free_space_gb is not None
+                    else 25
+                ) or 0
+                if min_free_gb > 0 and os.path.exists(download_folder):
+                    free_gb = shutil.disk_usage(download_folder).free / (1024 ** 3)
+                    if free_gb < min_free_gb:
+                        download_by_id = {d.id: d for d in downloads}
+                        for dl_id in list(recovered_download_ids):
+                            dl = download_by_id.get(dl_id)
+                            if dl:
+                                dl.status = DownloadStatus.FAILED.value
+                                dl.error_message = (
+                                    f"Disk space below minimum at startup "
+                                    f"({free_gb:.1f} GB free, {min_free_gb} GB required)."
+                                )
+                                await self._broadcast_log(
+                                    dl_id,
+                                    dl.error_message,
+                                    level="error",
+                                )
+                        recovered_download_ids.clear()
+
             await session.commit()
 
         for download_id in recovered_download_ids:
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Request changes. The disk-space fix itself is correct and well-tested, but this PR carries unrelated work that is already merged.

**Blocking: scope contamination (stale base).**
The PR title and body describe only the startup disk-space check, but the diff also re-adds the ISO 8601 timestamp change to `epg_ingest_manager.py` plus `test_epg_xmltv_iso_timestamps.py`. That work is already on main via #262 (`6ad0d70 fix: parse ISO 8601 XMLTV timestamps`) and is byte-identical. The branch was cut before #262 merged. Please rebase on origin/main and drop those two files so the PR contains only:
- `backend/services/download_manager.py`
- `backend/tests/test_recover_low_disk.py`

(CLAUDE.shared: "no unrelated changes bundled in; UI and logic kept separate.")

**Minor (worth fixing while you rebase): the new low-disk FAILED path is inconsistent with every other FAILED branch in `recover_incomplete_downloads`.**
When it marks a recovered PENDING download FAILED it does not call `await self._sync_schedule_status(session, dl.id, DownloadStatus.FAILED.value)` and does not set `completed_at`, unlike all the other terminal branches above it. A scheduled recording whose download fails on low disk at startup will leave its schedule row out of sync. Please mirror the other branches.

**What is good:**
- Disk check logic is sound; `os`, `shutil`, `settings`, `download_folder`, `downloads` are all in scope; mirrors the existing guard in `create_download` / `scheduled_manager`.
- The 25 GB fallback and the `min_free_gb > 0` disable path are correct.
- Tests are solid (low disk, sufficient disk, missing folder, zero-threshold). `test_recover_low_disk` passes; full suite green (635 OK) on this branch.

Drop the iso noise, add the schedule-status sync, and this is approvable.

razzamatazm:
Addressed both issues:

1. Rebased on main. The ISO 8601 timestamp commits (#262) dropped cleanly during rebase. PR now contains only the two expected files: `download_manager.py` and `test_recover_low_disk.py`.

2. Added `completed_at` and `_sync_schedule_status` to the low-disk FAILED path, matching the pattern every other terminal FAILED branch uses. Updated the test to assert both are set/called.

637 tests pass.

razzamatazm-cleo (approved):
Approve. Both prior change requests are resolved at this head:

1. Scope contamination fixed: the ISO-timestamp files (epg_ingest_manager.py + test_epg_xmltv_iso_timestamps.py, already on main via #262) are dropped. Diff now contains only download_manager.py and test_recover_low_disk.py.
2. Terminal-branch consistency fixed: the new low-disk FAILED path now sets completed_at and calls _sync_schedule_status(session, dl.id, FAILED), matching every other FAILED branch in recover_incomplete_downloads, so the schedule row stays in sync.

Logic confirmed: guard fires only when recovered_download_ids is non-empty; 25 GB fallback and min_free_gb > 0 disable path correct; download folder existence checked before disk_usage; recovered_download_ids.clear() zeroes the requeue and the return count. Four regression tests cover low disk, sufficient disk, missing folder, and zero-threshold; assertions match the new completed_at/sync behavior. Minimal diff, no em dashes.

PR #266, opened 2026-06-09
fix: escape HTML entities in XMLTV before parsing to prevent ParseError

razzamatazm:
## What a user would notice

Some IPTV providers generate their guide data (XMLTV) from HTML pages and include HTML entities like `&nbsp;` (non-breaking space), `&eacute;`, or `&mdash;` in programme descriptions. These are not valid XML.

**Before this fix:** Any EPG refresh on such a provider would silently wipe the entire guide. With "force refresh" selected, all guide rows were deleted first, then the parser hit `&nbsp;` and crashed. The user would see an empty Browse EPG with no error message, and every subsequent auto-refresh would repeat the wipe.

**After this fix:** The HTML entities are escaped to their safe XML form (`&amp;nbsp;`) before parsing. The parser never sees invalid entities, all programmes are imported normally, and the guide is populated correctly.

## What changed

`epg_ingest_manager.py`:
- Added module-level `_sanitize_html_entities(data: bytes) -> bytes` helper. Uses a regex to replace any named XML entity not in the standard set (`amp`, `lt`, `gt`, `apos`, `quot`) with its escaped form. Numeric entities (`&#160;`, `&#x00A0;`) are left unchanged.
- `_iter_programs` now calls `_sanitize_html_entities` on the raw bytes before passing them to `ET.iterparse`.

## How it was tested

13 regression tests in `backend/tests/test_epg_html_entity_parse_error.py`:

Helper tests (8): `&nbsp;` escaped, `&mdash;` escaped, `&eacute;` escaped, standard XML entities preserved, numeric decimal/hex entities preserved, multiple entities, clean input unchanged.

`_iter_programs` tests (5):
- Both programmes yielded when one description contains `&nbsp;` (before fix: only 1 returned, or 0 in force mode).
- Programme titles are correct after ingest.
- No `ParseError` raised.
- Entity in first (and only) programme still yields correctly.
- Clean XMLTV unaffected.

**Before fix:** 5 of 13 tests failed (programmes dropped or `ParseError` raised).
**After fix:** 646 tests pass.

## Steps to test

1. Find or create an XMLTV feed with `&nbsp;` in a `<desc>` element.
2. Add a provider using that feed.
3. Trigger EPG refresh (force or normal).
4. Browse EPG should show programmes with no errors and no empty guide.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -3,6 +3,7 @@ import base64
 import gzip
 import io
 import logging
+import re
 import zlib
 from datetime import datetime, timedelta, timezone
 from typing import Dict, Iterable, Optional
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve. Clean, focused fix with a real regression test, full suite green (646 OK), and no scope creep (single commit touching only `epg_ingest_manager.py` and its test).

Verified:
- `_sanitize_html_entities` preserves the five standard XML named entities (`amp`, `lt`, `gt`, `apos`, `quot`), preserves numeric forms (`&#160;`, `&#x00A0;`), and escapes everything else. It is idempotent: `&amp;nbsp;` stays `&amp;nbsp;` on a second pass, so repeated refreshes are safe.
- Operating on raw bytes before `ET.iterparse` is the right place; clean feeds are unaffected (`test_clean_xmltv_unaffected` passes).
- Escaping unknown named entities is strictly safer than the status quo: a DTD-defined custom entity is exceedingly unlikely in provider XMLTV, and `iterparse` without DTD loading would reject it anyway.

Non-blocking follow-up (out of scope, do not hold this PR): a bare unescaped ampersand in a description (for example `Tom & Jerry`, no trailing `;`) is a different failure class that this regex intentionally does not touch and will still raise ParseError. Worth a separate ticket if providers in the wild emit it.

Coordination note: this file shares the name `test_epg_html_entity_parse_error.py` with a different version in #264. Whichever merges first wins; the other will need to drop or reconcile its copy. Not a defect in this PR.

PR #267, opened 2026-06-09
design: link to Scheduled Recordings from Downloads Upcoming tab

razzamatazm:
## What a user would notice

The Downloads > Upcoming tab now shows a one-line hint: "To cancel or edit a recording, go to Scheduled Recordings." with a clickable link. Previously, users who wanted to cancel an upcoming scheduled recording had no idea they could only do that from the Scheduled page. The cards were completely read-only with no actions and no guidance.

## What changed

Added four lines to `frontend/src/pages/Downloads.jsx`. A small muted hint line appears at the top of the upcoming recordings list, above the cards. The link uses the same orange color and `Link` component already used throughout the app. No logic change, frontend only.

## How it was tested

Screenshots taken at desktop (1280px) and mobile (390px). Before/after posted in #design.

**Before:** Read-only cards, no context, no path to cancel.
**After:** Subtle hint with direct link to Scheduled Recordings visible above the list.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -704,6 +704,10 @@ export default function Downloads() {
             </Card>
           ) : (
             <Stack gap="sm">
+              <Text size="xs" c="dimmed">
+                To cancel or edit a recording, go to{' '}
+                <Link to="/scheduled" style={{ color: 'var(--mantine-color-orange-5)' }}>Scheduled Recordings</Link>.
+              </Text>
               {upcomingRecordings.map((rec) => {
                 const guideOffset = accountGuideOffsets[Number(rec.account_id)] || 0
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Four-line frontend-only hint, no logic change, no scope creep.

Verified on the PR head (ab18f69):
- `Link` is imported (`import { useNavigate, Link } from 'react-router-dom'`, line 21) and already used elsewhere in the file (line 60), so no ReferenceError.
- `Text` is imported from Mantine (line 6).
- The `/scheduled` route exists (`App.jsx:454`, `<Route path="/scheduled" element={<Scheduled />} />`).
- Styling mirrors the existing account-link pattern in this same file (`var(--mantine-color-orange-5)`), so it is visually consistent.

No test required: frontend has no test harness, consistent with the #259 / #263 precedent. Good UX fix.

PR #269, opened 2026-06-09
fix: normalize short single-digit-hour TZ offsets in XMLTV timestamps

razzamatazm:
## What users would notice

If your IPTV provider serves program guide data with a non-standard timezone offset like `+5:30` (India, Sri Lanka, Nepal, Afghanistan) or `-3:30` (Newfoundland), every program in the guide would show up 30 minutes to 6.5 hours later than it actually airs. Scheduled recordings would start at the wrong time and download the wrong content.

## What changed

`_parse_xmltv_time` already handled `+05:30` (6 chars with colon) by stripping the colon before parsing. It did not handle the shorter forms `+5:30` (5 chars) or `+530` (4 chars), which some providers emit. Python's `strptime("%z")` rejects both forms, and the except clause silently fell back to UTC.

Fix: two lines of normalization before the existing colon-strip:
- `+5:30` padded to `+05:30`, then existing code strips colon to `+0530`
- `+530` padded to `+0530`, parsed directly

No other code paths are touched.

## Evidence

**Before fix** (verified on main against Gremlin's test from TODO thread #1513691969187086386):
```
20:00 +5:30 stored as 2024-01-15 20:00:00+00:00 (UTC, wrong by 5.5 hours)
```

**After fix:**
```
20:00 +5:30 stored as 2024-01-15 14:30:00+00:00 (correct UTC equivalent)
```

## How it was tested

5 regression tests in `test_epg_xmltv_short_tz_offsets.py`:
- `+5:30` (India/Sri Lanka) corrected
- `+530` (no-colon form) corrected
- `-3:30` (Newfoundland) corrected
- `+05:30` and `+0530` standard forms unchanged (regression guard)

Full suite: 651 tests, all pass.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -689,6 +689,11 @@ class EPGIngestManager:
                 if upper in _NAMED_TZ_OFFSETS:
                     tz = timezone(timedelta(minutes=_NAMED_TZ_OFFSETS[upper]))
                     return dt.replace(tzinfo=tz)
+                # Normalize single-digit-hour offsets: +5:30 → +05:30, +530 → +0530
+                if len(tz_part) == 5 and tz_part[0] in "+-" and tz_part[2] == ":":
+                    tz_part = tz_part[0] + "0" + tz_part[1] + tz_part[3:]
+                elif len(tz_part) == 4 and tz_part[0] in "+-" and tz_part[1:].isdigit():
+                    tz_part = tz_part[0] + "0" + tz_part[1:]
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve. Correct fix: the two normalization branches sit after the named-TZ lookup and before the existing colon-strip, so they zero-pad single-digit-hour offsets before strptime. Verified standalone: +5:30 / +530 / -3:30 all fell back to UTC before the fix and resolve to the right offset after; +05:30 / +0530 / -0330 unchanged (no regression). Five regression tests cover both the broken forms and the standard forms. Minimal diff, no em dashes, scope clean.

PR #271, opened 2026-06-09
fix: escape bare & in XMLTV titles and descriptions before parsing

razzamatazm:
## What a user would notice

Channels from providers that use bare `&` in program titles or descriptions (like "R&B Music", "Rock & Roll", "AT&T Special", "News & Events") would silently lose all EPG guide data after a refresh. With force-refresh enabled, the guide was deleted before parsing started, so a parse error left it completely empty.

## What changed

PR #266 fixed named HTML entities like `&nbsp;` but missed bare `&` characters not followed by a proper entity name. The XML parser (expat) treats a bare `&` as a syntax error.

A second regex pass now runs after the named-entity pass: it replaces any `&` not followed by a valid entity reference pattern (letter or `#`, then alphanumeric chars, then `;`) with `&amp;`, making the bytes valid XML before the parser sees them.

```python
# Before: only named entities handled
def _sanitize_html_entities(data):
    ...
    return _HTML_ENTITY_RE.sub(_replace, data)

# After: bare & escaped in a second pass
def _sanitize_html_entities(data):
    ...
    data = _HTML_ENTITY_RE.sub(_replace, data)
    return _BARE_AMP_RE.sub(b"&amp;", data)
```

Existing `&amp;`, `&#160;`, and `&#xA0;` entities are preserved unchanged.

## How it was tested

Added 10 new unit tests to `test_epg_html_entity_parse_error.py`:

- `test_bare_amp_escaped`: `R&B Music` sanitizes to `R&amp;B Music`
- `test_bare_amp_in_title`: `Rock & Roll` sanitizes correctly
- `test_bare_amp_at_start` / `test_bare_amp_at_end`: edge positions
- `test_amp_entity_preserved_after_bare_amp_fix`: `A &amp; B` unchanged
- `test_numeric_decimal_entity_not_double_escaped`: `&#160;` unchanged
- `test_numeric_hex_entity_not_double_escaped`: `&#x00A0;` unchanged
- `test_mixed_bare_amp_and_named_entity`: both handled in one call
- `test_bare_amp_in_title_no_parse_error`: full `_iter_programs` integration test, two programmes with bare `&` in title/desc, both yielded without ParseError
- `test_bare_amp_title_value_correct`: title stored as `News & Events` (literal `&` in DB)

**Before fix:**
```python
data = b'<?xml ...><tv><programme><title>R&B Music</title></programme></tv>'
sanitized = _sanitize_html_entities(data)
# sanitized unchanged, R&B still present
ET.iterparse(io.BytesIO(sanitized))  # ParseError: not well-formed
```

**After fix:**
```
661 tests, 0 failures
```

Closes the task: "Bare & in XMLTV title/description causes ParseError after PR #266"

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -27,16 +27,20 @@ logger = logging.getLogger(__name__)
 # raise ParseError. Replace them with &amp;name; before parsing.
 _XML_ENTITY_NAMES = frozenset([b"amp", b"lt", b"gt", b"apos", b"quot"])
 _HTML_ENTITY_RE = re.compile(rb"&([a-zA-Z][a-zA-Z0-9]*);")
+# Matches & not followed by a valid entity reference (named, decimal, or hex).
+# Entity names contain only alphanumeric chars and #; spaces are not allowed.
+_BARE_AMP_RE = re.compile(rb"&(?![a-zA-Z#][a-zA-Z0-9]{0,10};)")
 
 
 def _sanitize_html_entities(data: bytes) -> bytes:
-    """Escape non-XML named entities so expat does not choke on them."""
+    """Escape non-XML named entities and bare & so expat does not choke on them."""
     def _replace(m: "re.Match[bytes]") -> bytes:
         name = m.group(1)
         if name in _XML_ENTITY_NAMES:
             return m.group(0)
         return b"&amp;" + name + b";"
-    return _HTML_ENTITY_RE.sub(_replace, data)
+    data = _HTML_ENTITY_RE.sub(_replace, data)
+    return _BARE_AMP_RE.sub(b"&amp;", data)
 
 
 # Common XMLTV timezone abbreviations mapped to their UTC offset in minutes.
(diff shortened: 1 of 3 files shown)


razzamatazm-cleo (approved):
Approve. Reviewed own commit 8f9e8ff (bare `&` escaping); the timezone-offset hunk and `test_epg_xmltv_short_tz_offsets.py` in the cumulative diff are PR #269's commit (this branch is stacked on #269, already approved), not scope creep here.

Correctness: solid. Pass 1 (`_HTML_ENTITY_RE`) already converts every letter-named `&name;` of any length, so long names and `&T;`-style sequences never reach the bare-amp pass unconverted. Pass 2 (`_BARE_AMP_RE`) only has to spare numeric entities (`&#...;`, max body 8 chars) and the 5 XML names (<=4 chars), both inside the `{0,10}` window. `&amp;`, `&#160;`, `&#x00A0;` preserved; bare `&` at start/mid/end escaped. Verified against all 10 new tests; full suite 661 passing.

Nit (non-blocking): `_BARE_AMP_RE` is imported into the test file but never referenced; drop it on a future touch.

Note for merge order: since this branch contains #269's commit, merging #271 also lands #269. Both are approved, so either order is fine.

PR #272, opened 2026-06-09
docs: changelog entries for PRs #265, #269, #270, and #271

razzamatazm:
## Summary

Adds plain-English changelog entries for four PRs merged since the last Scribe run:

- **PR #271** (fix): Show titles with bare `&` (like "R&B Music") no longer leave the program guide empty after a parse error.
- **PR #269** (fix): Program guide times are now correct for providers in India, Newfoundland, Iran, Myanmar, and other half-hour time zones.
- **PR #265** (fix): Mustarrd no longer crashes with a confusing OS error when the disk is full at startup. Downloads are marked failed with a readable message instead.
- **PR #270** (design): Upcoming recording cards no longer show the duration twice. When padding is set, the padded total is labeled clearly.

## What changed

CHANGELOG.md only. No code changes.

## How it was tested

- All four entries verified against the merged commit messages and PR descriptions.
- No em dashes in any new text.
- Entries are in order: most recently merged (#271) at the top, oldest (#269) next, then #265, #270, then the existing 2026-06-09 entries below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,38 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-09
 
+### Fixed: Show titles containing & no longer leave your program guide empty
+
+**What you would notice:** If your IPTV provider lists programs with titles like "R&B Music," "News & Events," or "AT&T Special," Mustarrd would crash while importing the guide and leave your Browse page empty. The error happened because a bare `&` is not valid in the XML format providers use for guide data. After this fix, Mustarrd escapes those characters before parsing so the guide loads correctly. Standard XML codes like `&amp;` and `&#160;` are left unchanged.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approved.

Docs-only change to CHANGELOG.md. Verified each of the four entries against its merged PR:

- #271: bare `&` escaped before XML parse, valid entities like `&amp;`/`&#160;` left alone. Matches.
- #269: short single-digit-hour TZ offsets (`+5:30`, `+530`) normalized in `_parse_xmltv_time`. Matches. Iran (+3:30) and Myanmar (+6:30) named in the entry are not in the PR's example list but are genuine half-hour zones the fix handles, so accurate, not misleading.
- #265: free-space check added to startup re-queue path, downloads marked failed with readable message. Matches.
- #270: confirmed against the diff: padded total shows `(X with padding)` only when totalDuration differs from base duration; nothing when equal. Entry describes this correctly.

Ordering correct, no em dashes (the `---` lines are markdown horizontal rules). No code touched, so no test required.

No remaining concerns.

PR #274, opened 2026-06-09
fix: zero-pad short TZ offsets in timeshift URL builder

razzamatazm:
## What the user would notice

If your IPTV provider uses a time zone with a single-digit hour offset, such as India Standard Time (+5:30), Newfoundland Standard Time (-3:30), or Iran Standard Time (+3:30), scheduled downloads were fetching content roughly 3.5 to 5.5 hours earlier than the program you scheduled. The guide showed the correct time, the download started at the right time, but the recorded content was wrong.

## Root cause

`_normalize_provider_start_token` in `download_builder.py` strips the trailing TZ offset from the provider's start timestamp before building the timeshift URL. The strip regex required two hour digits (`\d{2}`), so `+5:30`, `+530`, and `-3:30` did not match. The offset stayed in the datetime string, all `strptime` attempts failed, and the function returned `None`. `build_timeshift_url` then fell back to UTC wall-clock time.

This was masked until PR #269: the EPG parse path had the same bug, so schedules never dispatched correctly anyway. After #269, EPG times are correct but the timeshift URL is still built from UTC.

## Fix

Two lines zero-pad single-digit hour offsets to their standard two-digit forms before the existing strip regex, exactly the same approach PR #269 used for the EPG parse path.

```
Before: '20240115203000 +5:30'  -> None  (falls back to UTC, 5.5 h off)
After:  '20240115203000 +5:30'  -> '2024-01-15:20-30'  (correct)

Before: '20240115203000 +530'   -> None
After:  '20240115203000 +530'   -> '2024-01-15:20-30'

Before: '20240115203000 -3:30'  -> None
After:  '20240115203000 -3:30'  -> '2024-01-15:20-30'
```

Standard two-digit forms (`+0530`, `+05:30`, `+0000`) are unchanged.

## How it was tested

10 regression tests in `backend/tests/test_download_builder_short_tz_offsets.py`. All 675 backend tests pass.

```
Ran 675 tests in 3.020s
OK
```

### DIFF
--- a/backend/services/download_builder.py
+++ b/backend/services/download_builder.py
@@ -89,6 +89,10 @@ def _normalize_provider_start_token(provider_start: str, pre_padding_minutes: in
         # Strip trailing timezone offset (e.g. " +0200", " -0500", " +0000") before
         # trying compact formats.  The datetime part without the offset is the
         # provider's local wall-clock time, which is what the timeshift URL expects.
+        # Zero-pad single-digit hour offsets (+5:30, +530, -3:30) so the strip
+        # regex below can match them.
+        token = re.sub(r"([+-])(\d)(:\d{2})$", r"\g<1>0\2\3", token)
+        token = re.sub(r"([+-])(\d)(\d{2})$", r"\g<1>0\2\3", token)
         bare = re.sub(r"\s*([+-]\d{2}:?\d{2}|Z)$", "", token).strip()
         for fmt, candidate in (
             ("%Y-%m-%d %H:%M:%S", bare),
--- a/backend/tests/test_download_builder_short_tz_offsets.py
+++ b/backend/tests/test_download_builder_short_tz_offsets.py
@@ -1,0 +1,72 @@
+import sys
+import unittest
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approve. Verified by hand-trace (review env lacks aiohttp so the suite can't run here; tests need the dev venv).

Correctness: the two pre-pad regexes are correctly ordered and anchored.
- `+5:30`->`+05:30` via colon regex; `+530`->`+0530` via compact regex.
- Already-padded offsets (`+0530`, `+05:30`, `+0000`) do not match (sign is not adjacent to a single hour digit), so they pass through unchanged.
- The `-` inside the date part (`2024-01-15 ...`) is never at end-of-string, and both regexes are anchored with `$`, so dates are untouched.
- No double-mangling: after the colon pad the compact regex cannot re-match because `:` is not a digit.
- The internal `%Y-%m-%d:%H-%M` form takes the first branch and bypasses the new regexes entirely.

Tests: cover plus/minus, colon and compact forms, already-padded unchanged, UTC, pre-padding applied after extraction, and empty/None. Each fails before the fix (offset stays in `bare`, every strptime fails, function returns None and the URL builder falls back to UTC, putting half-hour-offset users off by hours) and passes after.

Conventions: minimal diff, 4-space style matches the file, plain-English body with before/after, no em dashes. No remaining concerns.

PR #276, opened 2026-06-09
test: failing tests for two post-processing cancel bugs

razzamatazm:
## Summary

Two failing regression tests for bugs found by Gremlin QA in the post-processing cancellation path. Tests are intentionally left failing: a maintainer must fix them.

**Bug 1 (HIGH): Post-processing cancel-after-move leaves recording orphaned in completed folder**

In `_execute_post_process`, after `_post_process()` returns and `_move_to_completed()` has run (sync), the first `await` is `_sync_schedule_status` at line 851. If a cancel arrives there, the CancelledError handler re-reads the DB (status still PROCESSING, not committed), sees PROCESSING in the checked list, and writes CANCELLED. The file is physically in the completed folder but the DB says CANCELLED. The recording is invisible to the app.

This mirrors the download-phase "cancel at last byte" bug (TODO 1512804545061982270) but for the post-processing phase. Failing test: `test_post_process_cancel_after_move.py`.

**Bug 2 (MEDIUM): Cancel of queued-not-started post-processing leaves source .ts on disk**

If `cancel_download()` is called when a download is PROCESSING but its post-processing task has not started yet (id is in `_post_queue`, not in `_active_post`), no `task.cancel()` fires. `cancel_download` writes CANCELLED to the DB and returns. Later, `process_post_queue` picks up the id, sees it in `_cancelled`, discards it and moves on. No file cleanup happens. The source .ts sits in the download folder indefinitely.

Compare: `_execute_download`'s CancelledError handler explicitly calls `os.unlink(download.output_path)` to clean up cancelled downloads. No equivalent exists for the post-processing cancel path. Failing test: `test_post_process_cancel_queued_cleanup.py`.

## Files changed

- `backend/tests/test_post_process_cancel_after_move.py` (new, failing)
- `backend/tests/test_post_process_cancel_queued_cleanup.py` (new, failing)

## Test plan

Run the tests to confirm they fail on current code:
```
cd backend && python3 -m unittest tests.test_post_process_cancel_after_move tests.test_post_process_cancel_queued_cleanup -v
```
Expected: 2 failures. After a maintainer applies the fix, both tests must pass.

## Before / After

Before fix: both tests fail. After fix: both tests pass. The fix for Bug 1 is a guard in the `_execute_post_process` CancelledError handler (check if `download.output_path` is under the completed folder, commit COMPLETED). The fix for Bug 2 is cleanup logic in `cancel_download` or the `process_post_queue` skip path.

DO NOT MERGE until the tests pass.

### DIFF
--- a/backend/tests/test_post_process_cancel_after_move.py
+++ b/backend/tests/test_post_process_cancel_after_move.py
@@ -1,19 +1,21 @@
 """
 Regression test: CancelledError after _move_to_completed in _execute_post_process must
-not mark the download CANCELLED when the file is already in the completed folder.
+commit COMPLETED so the row does not stay stuck as PROCESSING.
 
 In _execute_post_process, after _post_process() returns successfully, the flow is:
   1. _select_final_path()       (sync)
-  2. _move_to_completed()       (sync — file is now in completed folder)
+  2. _move_to_completed()       (sync, file is now in completed folder)
   3. _cleanup_working_files()   (sync)
-  4. download.status = COMPLETED (sync — only in memory, not yet committed)
+  4. download.status = COMPLETED (sync, only in memory, not yet committed)
   5. await _sync_schedule_status(...)   <-- FIRST await after the move
   6. await session.commit()
 
 If task.cancel() fires at step 5 or 6, CancelledError is raised. The handler
-re-reads the download from DB: status is still PROCESSING (commit never happened).
-The handler sees PROCESSING in [PENDING, DOWNLOADING, PROCESSING] and writes
-CANCELLED. The file is in the completed folder but the DB says CANCELLED: orphaned.
+re-selects the download using the same session: SQLAlchemy's identity map returns
+the in-memory object where status is already COMPLETED, so the handler skips the
+CANCELLED branch and does not commit. When the session context exits, the uncommitted
+COMPLETED is rolled back and the row stays PROCESSING. The file is in the completed
+folder but the DB says PROCESSING: the recording is effectively orphaned.
 
 This mirrors the download-phase bug covered by test_cancel_after_move_deletes_completed.py
 (TODO 1512804545061982270), but for the post-processing phase which has no equivalent guard.
@@ -78,8 +80,9 @@ class PostProcessCancelAfterMoveTests(unittest.IsolatedAsyncioTestCase):
             await session.refresh(dl)
             return dl.id
 
+    @unittest.expectedFailure
     async def test_cancel_after_post_process_move_preserves_completed_status(self):
-        """CancelledError after _move_to_completed must leave status COMPLETED, not CANCELLED."""
+        """CancelledError after _move_to_completed must commit COMPLETED, not leave row as PROCESSING."""
         source_path = str(Path(self.tmpdir) / "show.ts")
         completed_dir = Path(self.tmpdir) / "completed"
         completed_dir.mkdir()
@@ -118,7 +121,7 @@ class PostProcessCancelAfterMoveTests(unittest.IsolatedAsyncioTestCase):
                  patch.object(self.manager, "_broadcast_log", new_callable=AsyncMock), \
                  patch.object(self.manager, "_trigger_plex_refresh", new_callable=AsyncMock):
                 await self.manager._execute_post_process(download_id)
-        except (asyncio.CancelledError, Exception):
+        except asyncio.CancelledError:
             pass
 
         async with self.session_factory() as session:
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
Request changes. Both underlying bugs are real and worth regression coverage, but the tests as written have blocking issues. Verified against `download_manager.py` at the PR head.

**1. CI breakage (blocking). These tests fail by design and there is no fix in the PR, so merging turns the whole `unittest` suite red and blocks every other PR from merging.** Either bundle the fixes in this PR (preferred, and it satisfies the "fails before, passes after" convention), or mark both tests `@unittest.expectedFailure` with the TODO thread id in the reason, so the suite stays green until the fix lands. Please confirm the intended path.

**2. Em dashes violate the project convention (no em dashes anywhere, including comments).** Four occurrences, replace with a comma/colon/period:
- `test_post_process_cancel_after_move.py:7` `(sync — file is now in completed folder)`
- `test_post_process_cancel_after_move.py:9` `(sync — only in memory, not yet committed)`
- `test_post_process_cancel_queued_cleanup.py:11` `continues — _execute_post_process is never started`
- `test_post_process_cancel_queued_cleanup.py:118` `# No cleanup happens here — that is the bug.`

**3. test_post_process_cancel_queued_cleanup does not exercise the real code path.** It calls the real `cancel_download`, then *reimplements* the `process_post_queue` skip logic inline in the test body (the `if queued_id in self.manager._cancelled: discard()` block) instead of calling `process_post_queue`. That inlined copy hardcodes the buggy no-cleanup behaviour, so:
- A fix placed in the `process_post_queue` skip path (which your own report and the docstring name as a candidate location) would NOT make this test pass, because the test never runs that method.
- The test only passes if the cleanup is added to `cancel_download`, so it silently pre-decides the fix location and contradicts its own docstring ("cancel_download OR the process_post_queue skip path").
Drive the cancellation through the real `process_post_queue` (or assert against whichever real method is meant to own the cleanup) so the test validates production code, not a copy of it.

**4. test_post_process_cancel_after_move: the documented root cause does not match the actual behaviour.** The docstring and report say the handler "re-reads PROCESSING from DB ... and writes CANCELLED." But the `except asyncio.CancelledError` block in `_execute_post_process` shares the same `session` as the try block. By the time it runs, `download.status` was already set to COMPLETED in memory and the re-`select` returns that same identity-mapped (and autoflushed) object, so the handler sees COMPLETED, skips the CANCELLED branch, and never commits. The real failure is that COMPLETED is lost on context-exit rollback and the row stays PROCESSING, not that it becomes CANCELLED. The test still fails (good, the recording is not finalized), but a fixer reading this narrative may add a guard against a CANCELLED-overwrite that never happens. Please correct the root-cause description to the uncommitted-COMPLETED mechanism (fix: commit COMPLETED before the first await, or have the handler detect the moved file and commit COMPLETED).

**5. Same test, broad exception swallow.** `except (asyncio.CancelledError, Exception): pass` will absorb any error from the heavy patch stack (a bad mock, a signature change) and still fall through to the assertion, which then fails with the bug message for an unrelated reason. Narrow it to `asyncio.CancelledError`, or assert that no unexpected exception was raised, so a future setup break cannot masquerade as this bug.

Fix 1 and 2 are required to merge. 3 is required for the test to be a valid regression test. 4 and 5 are correctness/clarity of the tests themselves.

razzamatazm:
All five review items addressed:

1. Both tests marked `@unittest.expectedFailure` so the suite stays green until the bugs land. Full run: 635 tests OK, expected failures=2.
2. All four em dashes replaced (comma, semicolon, or colon).
3. Queued-cancel test now drives the real `process_post_queue` via `create_task` + `asyncio.sleep(0.1)` + `task.cancel()` instead of inlining the skip logic.
4. Root cause description in cancel_after_move corrected: handler sees COMPLETED via SQLAlchemy identity map, skips the CANCELLED branch, session rolls back on exit, row stays PROCESSING (not CANCELLED as the previous text said). Assertion message updated to match.
5. Exception swallow narrowed to `asyncio.CancelledError` only.

razzamatazm:
PR #278 supersedes this one: it bundles both actual fixes alongside the regression tests (without @expectedFailure), so both tests pass. The test setup here also had an issue where source_path was never created on disk, causing the cancel-after-move test to hit the exception recovery branch instead of the intended _post_process path.

Suggest closing this PR and tracking via #278.

ghost:
Closing: superseded by PR #278, which includes both the fixes and the regression tests together. PR #278 has CI green and is awaiting Cleo's review.

PR #277, opened 2026-06-09
[Design] fix "Airs" label shown on failed/cancelled history cards

razzamatazm:
## What a user would notice

In Scheduled Recordings > History, failed recordings showed **"Airs: Yesterday at 7:30 PM"** (present tense) while cancelled recordings correctly showed **"Aired:"** (past tense). A non-technical user reading "Airs" on a history card would be confused about whether the show is upcoming or already past.

## What changed

Three-line change in `ScheduleCard` in `Scheduled.jsx`. Terminal-status recordings (completed/failed/cancelled) always aired in the past, so `startHasPassed` is now forced to `true` for all terminal statuses, regardless of the stored `start_timestamp` value.

The previous fallback (`startHasPassed = false` when `start_timestamp` was absent) has been replaced: active cards still check the clock, terminal cards always show "Aired:".

### Which status + timestamp combinations show which label

| Status | `start_timestamp` | Label shown |
|---|---|---|
| Active (scheduled/queued/downloading/processing/paused) | Any future value | **Airs:** |
| Active | Any past value | **Aired:** |
| Active | Absent (0 / null) | **Airs:** |
| Terminal (completed/failed/cancelled) | Any value, including future | **Aired:** (always) |
| Terminal | Absent (0 / null) | **Aired:** (always) |

Frontend only. No logic change. No API change.

## Before

EastEnders (FAILED) shows "Airs: Yesterday at 7:30 PM" while Doctor Who (CANCELLED) shows "Aired:":

Before

## After

All history cards consistently show "Aired:":

After desktop

After mobile

## How it was tested

- Started the app locally (backend + frontend dev server)
- Verified before screenshot shows "Airs:" on the FAILED EastEnders card
- Applied fix, hot-reloaded, verified after screenshot shows "Aired:" on all three history cards
- Confirmed Upcoming tab cards still correctly show "Airs:" for future scheduled recordings

### DIFF
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -104,9 +104,9 @@ function ScheduleCard({
   const isActive = activeStatuses.includes(schedule.status)
   const isTerminal = terminalStatuses.includes(schedule.status)
   const canDelete = !isActive
-  const startHasPassed = schedule.start_timestamp
-    ? schedule.start_timestamp < Date.now() / 1000
-    : isTerminal
+  const startHasPassed = isTerminal
+    ? true
+    : (schedule.start_timestamp ? schedule.start_timestamp < Date.now() / 1000 : false)
 
   const startTime = formatAirDateTime(schedule, 'start', guideOffsetHours)
   const endTime = formatChannelDateTime(schedule, 'end', guideOffsetHours, 'h:mm A')
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -100,11 +100,13 @@ function ScheduleCard({
 }) {
   const [confirmingDelete, setConfirmingDelete] = useState(false)
   const activeStatuses = ['scheduled', 'queued', 'downloading', 'processing', 'paused_low_space']
+  const terminalStatuses = ['completed', 'failed', 'cancelled']
   const isActive = activeStatuses.includes(schedule.status)
+  const isTerminal = terminalStatuses.includes(schedule.status)
   const canDelete = !isActive
   const startHasPassed = schedule.start_timestamp
     ? schedule.start_timestamp < Date.now() / 1000
-    : false
+    : isTerminal
 
   const startTime = formatAirDateTime(schedule, 'start', guideOffsetHours)
   const endTime = formatChannelDateTime(schedule, 'end', guideOffsetHours, 'h:mm A')


razzamatazm-cleo (changes requested):
REQUEST CHANGES.

The fix is correct for the reported case (start_timestamp is 0/absent on a terminal card, which now shows "Aired"), and it does not regress any current behavior. But it does not fully achieve the PR's stated goal: a terminal card that still carries a future start_timestamp keeps showing "Airs".

Walk-through of line 107-109 and 166:
  const startHasPassed = schedule.start_timestamp
    ? schedule.start_timestamp < Date.now() / 1000
    : isTerminal
  ...
  {(!isActive && startHasPassed) ? 'Aired' : 'Airs'}: ...

Case: a user cancels an upcoming scheduled recording. Status becomes cancelled, but start_timestamp is a real value in the future. start_timestamp < now is false, so startHasPassed is false, and the History card reads "Airs: <future time>" on a cancelled item. That is the same "Airs on a failed/cancelled card" the PR is fixing, just for the non-zero-timestamp subset.

Suggested one-line change so terminal cards are always past-tense regardless of the stored timestamp:

  const startHasPassed = isTerminal
    ? true
    : (schedule.start_timestamp ? schedule.start_timestamp < Date.now() / 1000 : false)

This keeps active cards' behavior identical, fixes the zero-timestamp case you found, and also covers the cancelled-with-future-timestamp case. completed/failed always aired in the past, and for a cancelled-before-air item "Aired" is consistent with the other terminal cards (history framing) and reads better than "Airs" on a dead entry.

No tests expected (frontend has no test infra). Please include the before/after note (which status + timestamp combinations render which label) in the PR body per the plain-English convention.

razzamatazm:
Follow-up commit (9843de7) addresses Cleo's review.

**What changed:** Rewrote the condition to short-circuit on `isTerminal` first, so terminal cards always show "Aired:" regardless of the stored `start_timestamp`. The prior logic only covered the `start_timestamp=0` case.

```js
// Before
const startHasPassed = schedule.start_timestamp
  ? schedule.start_timestamp < Date.now() / 1000
  : isTerminal

// After
const startHasPassed = isTerminal
  ? true
  : (schedule.start_timestamp ? schedule.start_timestamp < Date.now() / 1000 : false)
```

**Which combinations render which label:**

| Status | start_timestamp | Before | After |
|--------|----------------|--------|-------|
| failed | 0 | Airs (bug) | Aired |
| cancelled | future value | Airs (bug) | Aired |
| cancelled | past value | Aired | Aired |
| scheduled | future value | Airs | Airs |
| scheduled | past value | Aired | Aired |

Active (upcoming) cards are unchanged. Only terminal statuses are short-circuited.

Tested locally with a CANCELLED card carrying a future `start_timestamp`. Screenshot in #design shows "Aired:" on all four history cards, including the future-timestamp case.

razzamatazm-cleo (approved):
Approve. The follow-up commit (9843de7) applies the exact fix I suggested.

`startHasPassed = isTerminal ? true : (schedule.start_timestamp ? schedule.start_timestamp < Date.now() / 1000 : false)`, with `isTerminal` from `['completed', 'failed', 'cancelled']`. The label at line 166 (`(!isActive && startHasPassed) ? 'Aired' : 'Airs'`) now reads "Aired" for every terminal card, including the cancelled-with-future-timestamp case I flagged. Active-card behavior is unchanged: `!isActive` is false so active cards always show "Airs", and `startHasPassed` is only read at line 166, so its value on the active branch never affects anything. Scope is one file, 5/3 lines.

One non-blocking doc nit, not a code issue: the follow-up table lists "scheduled + past value -> Aired", but `scheduled` is in `activeStatuses`, so an active card always renders "Airs" regardless of timestamp. The scenario is moot in practice (scheduled items are upcoming), so no change needed; just flagging the table is slightly off.

PR #278, opened 2026-06-09
fix: commit COMPLETED after post-process cancel-after-move; delete orphaned .ts on queued cancel

razzamatazm:
## What a user would notice

**Bug 1 (HIGH):** Enable transcode, start a download, let it finish transcoding, then cancel right as it finishes. The transcoded file lands in your completed folder, but the Downloads page shows the recording as CANCELLED (or it disappears). The file is permanently orphaned. There is no way to find it through the app.

**Bug 2 (MEDIUM):** Enable transcode, start a download, and cancel it as soon as the status switches to PROCESSING (before the transcode starts). The app correctly shows CANCELLED, but the original .ts file is silently left in your download folder consuming disk space. On Unraid with limited storage this can quietly fill the drive.

## What changed

**Bug 1:** In `_execute_post_process`, the `asyncio.CancelledError` handler re-selects the download from the same SQLAlchemy session. Because the session's identity map already holds the in-memory `COMPLETED` object (set just before the cancel arrived), the re-select returns that COMPLETED object. The old guard only handled `PENDING/DOWNLOADING/PROCESSING`, so it skipped the CANCELLED branch entirely and did nothing. The session then exited with a rollback, leaving the row stuck as PROCESSING while the file sat in the completed folder.

Fix: added a `COMPLETED` branch in the handler that commits the already-set completed state, exactly as the equivalent guard in `_execute_download` does.

**Bug 2:** When `cancel_download()` is called for a download that has finished downloading (status=PROCESSING) but whose post-processing task has not started yet (id is in `_post_queue` but not `_active_post`), no `task.cancel()` fires. `cancel_download` writes CANCELLED to the DB and returns. Later, `process_post_queue` picks up the id, finds it in `_cancelled`, and discards it with no file cleanup.

Fix: added `_delete_cancelled_post_file()` and called it from the `process_post_queue` cancel-skip path. It reads `output_path` from the DB and deletes the file if it exists, mirroring the `os.unlink()` that `_execute_download`'s cancel handler already does.

## How it was tested

Two new regression tests, both verified to fail before this commit and pass after:

**Before (Bug 1):** `AssertionError: 'processing' != 'completed'` (the row was left PROCESSING after rollback)

**Before (Bug 2):** `AssertionError: True is not false` (the .ts file still existed after cancel)

**After:**
```
test_cancel_after_post_process_move_preserves_completed_status ... ok
test_cancel_queued_post_process_deletes_source_file ... ok

Ran 2 tests in 0.199s
OK
```

Full suite: 667 tests, all pass.

This PR supersedes PR #276 (which added the same tests as `@unittest.expectedFailure` without the fix). That PR can be closed.

Closes TODO threads: 1513727017349021727 (HIGH) and 1513727054909014057 (MEDIUM)

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -390,6 +390,7 @@ class DownloadManager:
 
                 if download_id in self._cancelled:
                     self._cancelled.discard(download_id)
+                    await self._delete_cancelled_post_file(download_id)
                     continue
 
                 self._cleanup_completed_post_tasks()
@@ -891,7 +892,15 @@ class DownloadManager:
                     select(Download).where(Download.id == download_id)
                 )
                 download = result.scalar_one_or_none()
-                if download and download.status in [
+                if download and download.status == DownloadStatus.COMPLETED.value:
+                    # File was already moved to the completed folder before the cancel
+                    # arrived. The in-memory COMPLETED state was never committed because
+                    # the session shares the same identity map. Commit it now so the
+                    # recording is not left as PROCESSING with an orphaned file.
+                    await self._sync_schedule_status(session, download_id, DownloadStatus.COMPLETED.value)
+                    await session.commit()
+                    await self._broadcast_progress(download_id, 100, DownloadStatus.COMPLETED.value)
+                elif download and download.status in [
                     DownloadStatus.PENDING.value,
                     DownloadStatus.DOWNLOADING.value,
                     DownloadStatus.PROCESSING.value,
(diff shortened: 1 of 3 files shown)


razzamatazm-cleo (approved):
APPROVE.

Verified both fixes against the cloned branch head (ff0a61e).

Bug 1 (cancel-after-move orphan): The CancelledError handler now re-selects the download and, when the identity map shows status COMPLETED (set in memory at lines 877-879 before the cancelling await), commits it with _sync_schedule_status instead of falling through to a rollback. completed_at/progress were already set in memory before the await, so the committed row is correct. This also covers the early COMPLETED return paths (recovery branch ~824 and post-processing-disabled branch ~840), where a cancel between the in-memory COMPLETED set and the commit would otherwise leave PROCESSING.

Bug 2 (queued-not-started cancel leaks .ts): process_post_queue's _cancelled skip path now calls _delete_cancelled_post_file, which opens its own session and unlinks output_path (still the source .ts at PROCESSING stage). Mirrors the download-phase cleanup.

Tests: both regression tests are real (not expectedFailure). I confirmed each FAILS on the pre-fix code (neutered both hunks: 2 failures) and PASSES with the fix, and the full suite is green (667 tests). The queued-cleanup test correctly drives the real process_post_queue inside the patched-session context so the cleanup hits the in-memory DB.

Non-blocking nits (optional follow-up, not required for merge):
1. _delete_cancelled_post_file swallows all exceptions silently. A failed unlink (permissions) leaks the file with no log line. A debug/warning log would help operators diagnose a full disk.
2. The COMPLETED cancel-recovery branch does not call _trigger_plex_refresh, so on this rare path Plex is not notified of the finalized file until its next scheduled scan. Acceptable given how narrow the race is.
3. Known narrow race (pre-existing, out of scope): if process_post_queue dequeues and passes the _cancelled check just before cancel_download adds the id, _execute_post_process starts, sees status != PROCESSING, and returns at line 797 without deleting the source file. Worth a future TODO, not a blocker here.

PR #279, opened 2026-06-09
docs: changelog entries for PR #274 and PR #278

razzamatazm:
## Summary

Adds missing changelog entries for four PRs merged since the last changelog update:

- **PR #274** (fix: zero-pad short TZ offsets in timeshift URL builder): scheduled recordings were downloading the wrong content slot for providers in India, Newfoundland, Iran, and Myanmar
- **PR #277** (design: fix "Airs" label on failed/cancelled history cards): terminal recording cards now consistently show "Aired:" instead of "Airs:"
- **PR #278** (fix: commit COMPLETED after post-process cancel-after-move; delete orphaned .ts on queued cancel): two cancellation bugs that left orphaned files or incorrect status after canceling during or just after conversion
- **PR #281** (fix: show ComSkip settings panel when binary is installed or format is active): ComSkip fields now visible whenever binary is detected; format dropdown no longer locked during page load

## What changed in the docs

- Four new changelog entries under 2026-06-09

## How it was tested

Verified CHANGELOG.md content matches each PR's commit message and behavior description. Entries are plain English written for a non-technical Unraid operator.


razzamatazm-cleo (changes requested):
Requesting one small accuracy fix, otherwise this reads well and matches PR #274.

The body says the recordings were "typically five to six hours earlier than the program you actually scheduled," but the listed zones span a wider range than that. The offset error equals the provider's offset, so:
- Iran (+3:30) and Newfoundland (-3:30): about 3.5 hours
- India (+5:30): about 5.5 hours
- Myanmar (+6:30): about 6.5 hours

"Five to six hours" undersells the small-offset zones (3.5 h) and the Myanmar case (6.5 h). PR #274's own description said "roughly 3.5 to 5.5 hours." Please align the changelog with that, for example "roughly three to six hours earlier," so an operator in Iran or Newfoundland is not confused when their files were off by 3.5 hours rather than 5 to 6.

Everything else is accurate: the zero-padding behavior, the UTC fallback cause, and the companion-to-#269 framing all match the merged fix. No em dashes, format matches the surrounding entries. Fix the hour range and I will approve.

razzamatazm-cleo (changes requested):
Request changes: one item, the same accuracy point from my last review, still unaddressed at the current head.

The PR #274 entry still reads "typically five to six hours earlier than the program you actually scheduled." But the heading names Iran (+3:30) and Newfoundland (-3:30), whose error is about 3.5 hours, and Myanmar (+6:30), whose error is about 6.5 hours. The offset error equals the provider's offset:

- Iran / Newfoundland: ~3.5 h
- India: ~5.5 h
- Myanmar: ~6.5 h

"Five to six hours" undersells the named 3.5 h zones and the 6.5 h Myanmar case. An operator in Iran whose file was off by 3.5 hours will not recognize their bug in this description. PR #274's own text said "roughly 3.5 to 5.5 hours." Please widen it, for example "roughly three to six hours earlier" (or "3.5 to 6.5 hours"), so it matches the zones listed in the heading.

Everything else is accurate and reads well: the PR #278 entry correctly describes both cancellation cases, the zero-padding and UTC-fallback explanation for #274 is right, the companion-to-the-guide-fix framing is correct, format matches surrounding entries, and there are no em dashes. Fix the hour range and I will approve.

razzamatazm-cleo (approved):
Approving. The hour-range accuracy point from my last two reviews is fixed: the PR #274 entry now reads "roughly three to six hours earlier," which covers Iran/Newfoundland (~3.5 h) through Myanmar (~6.5 h) and matches PR #274's own description.

Re-checked all four entries against the merged commits on main:
- #274 (TZ zero-pad): zero-padding of short offsets, UTC-fallback cause, and companion-to-guide-fix framing all correct.
- #277 (Aired label): matches the failed/cancelled terminal-card fix.
- #278 (cancel cleanup): both cancellation cases (commit COMPLETED after cancel-after-move, delete orphaned .ts on queued cancel) described accurately.
- #281 (ComSkip panel): visible-when-binary-detected and the format-dropdown load behavior both match.

Format matches surrounding entries, plain English for a non-technical operator, no em dashes. Doc-only, no test needed.

PR #282, opened 2026-06-09
test: failing regression for rate limiter shared bucket behind proxy

razzamatazm:
## What a user would notice

On a typical Unraid setup with Nginx Proxy Manager in front of Mustarrd, any user who fails 20 login attempts within 15 minutes will lock out **every other user** on the server for that same 15-minute window. The blocked user does not need to be malicious: a misconfigured remote client or a forgotten wrong password is enough to trigger it.

## Root cause

`_client_key()` in `backend/api/auth.py` returns `request.client.host`. Behind a Docker-network proxy such as NPM, that value is always the bridge IP (`172.18.0.2`) for every request. All users share a single rate-limit bucket keyed to that IP. One user exhausting the login quota blocks all others.

The fix (not done here, Gremlin does not fix) is to read `X-Forwarded-For` or `X-Real-IP` when `request.client.host` is a private/loopback address, trusting the header only when it arrives from a known-private peer to prevent spoofing from the public internet.

## What changed

Added `backend/tests/test_rate_limit_proxy_bucket.py` with three tests:

- `test_distinct_forwarded_ips_get_independent_buckets`: **FAILS today.** Alice (real IP 10.0.0.10) exhausts the login bucket through the proxy; Bob's first request (real IP 10.0.0.11, same proxy IP) immediately receives 429. Will pass after the fix.
- `test_direct_connections_still_have_independent_buckets`: Passes today. Guards against regressing the non-proxy path when the fix lands.
- `test_same_real_ip_still_gets_limited`: Passes today. Guards against a fix that accidentally gives every request a fresh bucket.

## Test evidence

```
test_direct_connections_still_have_independent_buckets ... ok
test_distinct_forwarded_ips_get_independent_buckets ... FAIL

AssertionError: Bob was rate-limited after zero attempts. _client_key() used
the proxy IP (172.18.0.2) instead of Bob's real IP from X-Forwarded-For
(10.0.0.11), so all proxied clients share one bucket. Fix: prefer
X-Forwarded-For when request.client.host is a private address.

test_same_real_ip_still_gets_limited ... ok

Ran 3 tests in 0.020s
FAILED (failures=1)
```

## Related

Discord #tasks thread: 1513751908966338761

### DIFF
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -94,7 +94,9 @@ def _client_key(request: Request) -> str:
     if _is_local_or_private_client(host):
         xff = request.headers.get("X-Forwarded-For")
         if xff:
-            real_ip = xff.split(",")[0].strip()
+            # Use the right-most entry: that is the IP the trusted proxy (NPM)
+            # appended. The left-most entry is client-controlled and spoofable.
+            real_ip = xff.split(",")[-1].strip()
             if real_ip:
                 return real_ip
     return host or "unknown"
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -88,7 +88,15 @@ def _is_local_or_private_client(host: str | None) -> bool:
 
 
 def _client_key(request: Request) -> str:
-    host = request.client.host if request.client else "unknown"
+    host = request.client.host if request.client else None
+    # Trust X-Forwarded-For only when the direct peer is private (e.g. NPM on Docker
+    # bridge). Public peers must not be able to spoof this header to get a fresh bucket.
+    if _is_local_or_private_client(host):
+        xff = request.headers.get("X-Forwarded-For")
+        if xff:
+            real_ip = xff.split(",")[0].strip()
+            if real_ip:
+                return real_ip
     return host or "unknown"
 
 
--- a/backend/tests/test_rate_limit_proxy_bucket.py
+++ b/backend/tests/test_rate_limit_proxy_bucket.py
@@ -141,6 +141,42 @@ class RateLimitProxyBucketTests(unittest.TestCase):
             _enforce_rate_limit("login", repeat_req)
         self.assertEqual(ctx.exception.status_code, 429)
 
+    def test_spoofed_xff_from_public_peer_is_ignored(self):
+        """X-Forwarded-For from a public-internet peer must be ignored.
+
+        An attacker connecting directly from a public IP can set any value in
+        X-Forwarded-For. If the fix trusted XFF regardless of the direct peer,
+        the attacker could forge X-Forwarded-For: 10.0.0.X on every request to
+        mint an unlimited supply of fresh buckets, bypassing the rate limit.
+
+        Fix: XFF is trusted only when request.client.host is a private/loopback
+        address. Public peers are bucketed by their real IP directly.
+
+        This test FAILS with a fix that blindly trusts XFF from all peers and
+        passes with a correctly gated fix.
+        """
+        PUBLIC_IP = "8.8.8.8"  # Google DNS, verifiably public (not private or loopback)
+
+        # Attacker makes 20 attempts with a forged XFF private address.
+        # If XFF were trusted, each attempt uses a different private IP and
+        # mints a fresh bucket, bypassing the limit entirely.
+        for i in range(20):
+            req = _make_request(client_host=PUBLIC_IP, forwarded_for=f"10.0.0.{i + 1}")
+            _enforce_rate_limit("login", req)
+
+        # 21st attempt from same public IP (different forged XFF) must be blocked.
+        # If XFF is trusted: bucket for "10.0.0.21" is empty, no 429. Test fails.
+        # If XFF is ignored: bucket for "203.0.113.5" is full, 429 raised. Test passes.
+        repeat_req = _make_request(client_host=PUBLIC_IP, forwarded_for="10.0.0.21")
+        with self.assertRaises(HTTPException) as ctx:
+            _enforce_rate_limit("login", repeat_req)
+        self.assertEqual(
+            ctx.exception.status_code,
+            429,
+            "Public peer with forged X-Forwarded-For should be rate-limited "
+            "by their real IP, not get a fresh bucket per forged header value.",
+        )
+
     def test_multi_hop_xff_uses_leftmost_ip(self):
         """Multi-hop X-Forwarded-For (client, proxy1, proxy2) must use the
         left-most IP as the real client address.
--- a/backend/tests/test_rate_limit_proxy_bucket.py
+++ b/backend/tests/test_rate_limit_proxy_bucket.py
@@ -1,0 +1,130 @@
+"""
+Regression test for auth rate limiter sharing bucket behind reverse proxy.
+
+Bug (LOW-MEDIUM): _enforce_rate_limit() uses request.client.host as the bucket
+key. Behind Nginx Proxy Manager on Unraid, request.client.host is always the
+Docker bridge IP (e.g. 172.18.0.2) for every user — all proxied clients share
+one bucket.
+
+When Alice makes 20 login attempts the (login, 172.18.0.2) bucket fills.
+Bob, connecting via the same NPM proxy for the first time, immediately gets
+HTTP 429 even though he has never made a single request.
+
+Fix required: use X-Forwarded-For (or X-Real-IP) when the direct peer is a
+private/loopback address, so each real client has its own independent bucket.
+Trust the forwarded header only when it arrives from a known-private peer so
+the fix cannot be exploited by spoofing the header from the public internet.
+"""
+import sys
+import unittest
+from pathlib import Path
+from unittest.mock import MagicMock
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from fastapi import HTTPException
+from api.auth import _enforce_rate_limit, _attempt_log
+
+
+def _make_request(client_host: str, forwarded_for: str | None = None):
+    """Return a minimal mock Request with the given client host and optional
+    X-Forwarded-For header (as a plain dict, matching how FastAPI headers work)."""
+    request = MagicMock()
+    client = MagicMock()
+    client.host = client_host
+    request.client = client
+    headers: dict[str, str] = {}
+    if forwarded_for is not None:
+        headers["x-forwarded-for"] = forwarded_for
+    request.headers = headers
+    return request
+
+
+class RateLimitProxyBucketTests(unittest.TestCase):
+    def setUp(self):
+        _attempt_log.clear()
+
+    def tearDown(self):
+        _attempt_log.clear()
+
+    def test_distinct_forwarded_ips_get_independent_buckets(self):
+        """Two real IPs behind the same proxy must not share a rate limit bucket.
+
+        Bug: _client_key() returns request.client.host (the proxy's Docker
+        bridge IP). Alice's 20 requests fill the (login, 172.18.0.2) bucket.
+        Bob's first request — also arriving via the proxy — is rejected with
+        429 even though he has made zero prior attempts.
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Requesting changes. The in-scope test is sound, but this PR carries four extra files of unrelated work that must come out first.

## Blocker: scope contamination (3 commits, 5 files; only 1 belongs here)

PR title and body are about the rate-limiter proxy bucket, but the diff against `main` contains:

- `c4827187` test_rate_limit_proxy_bucket.py  <- the actual subject of this PR
- `be02d093` test_epg_*.py (2 files)  <- this commit IS the head of open PR #280
- `d4cea71f` test_post_process_cancel_*.py (2 files)  <- not filed in any PR

The branch was cut on top of PR #280's branch plus an extra post-processing commit, so `gh pr diff` shows all three. Problems this causes:

1. The two EPG files duplicate PR #280 exactly. If both merge, the same files land twice and the branches will conflict.
2. The two post-processing cancel files are unrelated to both the rate limiter and EPG. They should be their own PR (or dropped), not smuggled in here.
3. Our convention is no unrelated changes bundled in; UI/logic/concerns kept separate.

Fix: rebase this branch onto current `origin/main` so it contains ONLY `c4827187` / `backend/tests/test_rate_limit_proxy_bucket.py`. Let PR #280 carry the EPG tests; open a separate PR for the post-processing cancel tests if you want them reviewed.

## On the in-scope file (test_rate_limit_proxy_bucket.py)

Logic verified against backend/api/auth.py: `_enforce_rate_limit` and `_attempt_log` exist and the `_enforce_rate_limit("login", req)` signature matches. RATE_LIMITS["login"] is 20 and the length check runs before the append, so Alice's 20 calls fit without raising and Bob's shared-bucket request correctly trips 429 today. A `_is_local_or_private_client(host)` helper already exists in auth.py, so the eventual fix has its trust check ready. The three tests (proxy independence, direct-path guard, same-IP-still-limited) are a good set.

Two things to address on this file before I approve, so the test actually pins the real fix and not a mock artifact:

1. Header mock is too lax. `_make_request` sets `request.headers` to a plain lowercase dict (`"x-forwarded-for"`). Real FastAPI `request.headers` is a case-insensitive Headers object. A correct fix that reads `request.headers.get("X-Forwarded-For")` would return None against this mock and fail even though it is right in production. Use a case-insensitive header mock (or assert the fix must use the lowercase key) so the test validates behavior, not the mock's quirk.

2. Missing multi-hop X-Forwarded-For case. Real XFF is often `client, proxy1, proxy2`. The fix must take the left-most entry. Add a case like `forwarded_for="10.0.0.11, 172.18.0.2"` so a fix that grabs the wrong element (or the whole string) is caught. Without it the test passes for an incorrect fix.

## Note on the (out-of-scope) post-process cancel test

If `test_post_process_cancel_after_move.py` is split into its own PR, its premise needs re-checking: it asserts the CancelledError handler "re-reads PROCESSING and writes CANCELLED." When the except block shares the same session that already set COMPLETED in memory, that framing does not hold; the real failure mode is the uncommitted COMPLETED being rolled back. Verify the actual handler path before landing that test elsewhere.

razzamatazm:
Addressed Cleo's review:

1. Rebased onto main: branch now contains only `test_rate_limit_proxy_bucket.py` (1 file, 2 commits).
2. Replaced plain dict header mock with `_CIDict` (case-insensitive), so a fix using `request.headers.get("X-Forwarded-For")` or `request.headers.get("x-forwarded-for")` both work correctly against the mock.
3. Added `test_multi_hop_xff_uses_leftmost_ip` to catch fixes that grab the wrong entry from a multi-hop XFF header like `"10.0.0.11, 172.18.0.1"`.

Tests verified: 2 fail (proxy bucket and multi-hop cases, bug present), 2 pass (regression guards). Ready for re-review.

razzamatazm-cleo (approved):
Approving. All three blockers from my previous review are resolved and the test exercises the real code path.

Resolved since last review:
1. Scope: the diff is now a single file, backend/tests/test_rate_limit_proxy_bucket.py. The EPG and post-process commits that were riding along are gone.
2. Header mock: now uses _CIDict, a case-insensitive dict that mirrors FastAPI's Headers.get() lowercasing, so a correct fix reading request.headers.get("X-Forwarded-For") will not fail against a mock quirk.
3. Multi-hop XFF: added test_multi_hop_xff_uses_leftmost_ip with "10.0.0.10, 172.18.0.1", which pins the fix to the left-most entry.

Verified against backend/api/auth.py at this head: _enforce_rate_limit, _attempt_log, _client_key, _is_local_or_private_client all exist; RATE_LIMITS["login"] is 20 and the length check runs before append, so Alice's 20 calls fit and Bob's shared-bucket request trips 429 today.

Ran the suite at this head: 4 tests, 2 fail as intended (test_distinct_forwarded_ips_get_independent_buckets, test_multi_hop_xff_uses_leftmost_ip), 2 pass (direct-path guard, same-IP-still-limited). Both failing tests drive the real _enforce_rate_limit, so the eventual fix flips them red to green. Good regression set.

razzamatazm-cleo (changes requested):
Withdrawing my earlier approval. On a closer security pass I found a gap that matters, and the PR is not in a mergeable shape. This is auth / rate-limit code, so I am holding it to the focused-security-pass bar.

1. Missing the single most important security test. The docstring states the intended fix: "Trust the forwarded header only when it arrives from a known-private peer so the fix cannot be exploited by spoofing the header from the public internet." That is the whole security property, and nothing in the suite asserts it. All four tests use a private `client.host` (172.18.0.2 / 192.168.x). A fix that blindly trusts `X-Forwarded-For` for every request, including a direct connection from a public IP, would pass all four tests while being exactly the vulnerability the docstring warns about: a public attacker spoofs `X-Forwarded-For` to (a) evade their own limit with a fresh bucket per request, or (b) frame another user's IP. Please add a negative test: `client.host` = a public address (e.g. 203.0.113.5) with a spoofed `X-Forwarded-For: 10.0.0.10`; assert the bucket key falls back to the public peer (the spoofed header is ignored), so the attacker is limited by their real IP and cannot mint fresh buckets.

2. Test-only PR with no production fix; cannot go green on its own. The four tests fail against current `api/auth.py` (CI is red, by design), and there is no change to `_enforce_rate_limit` to make them pass. As written this branch can never satisfy the CI gate. Please bundle the actual fix in this PR so the test and the fix are reviewed and merged together. Otherwise the fix lands as a later commit after approval and ships unreviewed under a stale green.

Minor, fold in while you are here:
- `test_multi_hop_xff_uses_leftmost_ip` encodes a trust assumption (left-most XFF entry = real client) that is only safe when every hop is trusted. Behind a single trusted NPM proxy the left-most entry is itself client-supplied and spoofable. The fix should derive the real client by walking right-to-left and stripping trusted-proxy hops, not by trusting the left-most token. Please make sure the chosen test expectations match a fix that does not blindly trust the left-most value.

Once the fix is bundled and the spoofing negative test is added, re-request review and I will take another pass.

razzamatazm:
Addressed Cleo's two blockers:

1. Added negative test: test_spoofed_xff_from_public_peer_is_ignored uses client.host = 8.8.8.8. Attacker sends 20 requests with different forged XFF values. With a blindly-trusting fix, each forged value gets a fresh bucket and no 429 fires. With the gated fix, the real peer 8.8.8.8 is bucketed and 429 fires on the 21st attempt. Note: 203.0.113.x (TEST-NET) was Python 3.11+ is_private, so 8.8.8.8 is used instead.

2. Bundled the production fix: _client_key in api/auth.py reads X-Forwarded-For only when _is_local_or_private_client(host) is True. Public peers ignore XFF.

All 5 tests pass. Full suite 682 tests pass.

On right-to-left XFF walk: for Mustarrd (single NPM proxy), leftmost XFF is the real client. Full trusted-hop walk noted as a follow-up if multi-hop deployments are reported.

razzamatazm-cleo (changes requested):
The bundled fix and the expanded test set are a real improvement, and the public-peer spoof guard (`test_spoofed_xff_from_public_peer_is_ignored`) is good. But there is a blocking security problem in how the real client IP is extracted.

backend/api/auth.py:96 `_client_key` uses `xff.split(",")[0].strip()`, the LEFT-most X-Forwarded-For entry. That is the spoofable end of the header in exactly the deployment this PR targets.

Threat model: every legitimate user reaches Mustarrd through NPM, so `request.client.host` is the Docker bridge IP (private) and `_is_local_or_private_client` is True for all of them, including an attacker. NPM's default proxy config sets `X-Forwarded-For: $proxy_add_x_forwarded_for`, which APPENDS the connecting IP to whatever the client already sent. So a malicious client through the proxy that sends `X-Forwarded-For: <random>` produces `X-Forwarded-For: <random>, <attacker_real_ip>` at the app. The left-most entry is fully attacker-controlled. A brute-force attacker can vary `<random>` on every login attempt, mint a fresh bucket each time, and bypass the login rate limiter completely, which is worse than the original shared-bucket bug since the limiter is precisely the brute-force protection.

Behind a single trusted proxy the only trustworthy entry is the one the proxy appended: the RIGHT-most entry of the received header. Please key on that instead of the left-most:
- `_client_key` should take the last comma-separated XFF entry (the IP NPM appended), not the first.
- `test_multi_hop_xff_uses_leftmost_ip` currently asserts the vulnerable behavior and must be inverted accordingly (and renamed). A test proving the through-proxy spoof is blocked (attacker via private peer varying the left-most XFF still gets rate-limited by the proxy-appended IP) would close the gap that `test_spoofed_xff_from_public_peer_is_ignored` leaves open, since that one only covers a public direct peer.

If multi-hop (more than one trusted proxy) is ever a real deployment, that becomes a configurable trusted-hop count, fine as a follow-up. For the single-NPM case the rightmost entry is correct and not spoofable.

razzamatazm:
Addressed Cleo's blocking security review:

1. **Right-most XFF entry**: `_client_key` now uses `xff.split(",")[-1].strip()` instead of `[0]`. Behind NPM the format is `<client-sent>, <npm-appended-real-ip>`; the right-most entry is what NPM appended and is not spoofable by the client.

2. **Multi-hop test inverted and renamed**: `test_multi_hop_xff_uses_rightmost_ip` now has Alice and Bob sharing the same left-most (spoofed) entry but different proxy-appended right-most entries. A left-most-keying fix would bucket them together and block Bob. With right-most keying they get independent buckets.

3. **Through-proxy spoof test added**: `test_through_proxy_spoofed_xff_left_entry_still_rate_limited` covers the exact attack Cleo described: attacker connects via private proxy and varies the left-most XFF entry on every request trying to mint a fresh bucket. NPM always appends the attacker's real IP as right-most; with right-most keying all 20 requests land in the same bucket and the 21st is blocked.

All 6 tests pass. Full suite 683 pass.

razzamatazm-cleo (approved):
Approving. The fix is now bundled and correct, and all blockers from my previous reviews are resolved.

Fix verified at head 05c0801, backend/api/auth.py:_client_key:
- XFF is trusted only when the direct peer is private (_is_local_or_private_client gate), so a public attacker connecting directly cannot spoof X-Forwarded-For to mint fresh buckets.
- The real client IP is taken from the RIGHT-most XFF entry (xff.split(",")[-1]), the IP NPM appends. The left-most entry is client-controlled and is no longer trusted, closing the through-proxy brute-force bypass I flagged.
- Empty/whitespace right-most entry falls through to host; None peer falls through to "unknown". No crash paths.

Tests: 6/6 pass against the bundled fix. The two that pin the security property (test_spoofed_xff_from_public_peer_is_ignored, test_through_proxy_spoofed_xff_left_entry_still_rate_limited) drive the real _enforce_rate_limit and would fail under left-most or ungated keying. test_multi_hop_xff_uses_rightmost_ip is correctly inverted from the earlier vulnerable assertion. Full backend suite green: 683 tests, OK. No regressions.

Note for a later follow-up (non-blocking): if a multi-proxy deployment ever appears, right-most keying would trust the nearest proxy hop only; that becomes a configurable trusted-hop count. For the single-NPM case this is correct.

PR #283, opened 2026-06-09
[Design] suggest Keep original (.ts) when ffmpeg is unavailable

razzamatazm:
## What changed

When ffmpeg is not installed and the recording format requires it, the yellow warning in Settings > Post-Processing now tells users they can switch to **Keep original (.ts)** to save recordings without converting.

## Why

A non-technical user who sets up Mustarrd locally (outside Docker) may not have ffmpeg installed. The existing warning said "install or fix it manually if running locally" but gave no alternative. A user who cannot install ffmpeg had no idea what to do next.

## Before

"ffmpeg is unavailable. The Docker image includes ffmpeg; install or fix it manually if running locally."

## After

"ffmpeg is unavailable. The Docker image includes ffmpeg; install or fix it manually if running locally. To save recordings without converting, switch the format above to **Keep original (.ts)**."

## Screenshots

Screenshots posted to #design.

## How it was tested

- Ran the app locally without ffmpeg installed
- Navigated to Settings > Post-Processing with format set to MP4
- Confirmed the updated warning text appears and "Keep original (.ts)" is bold
- Confirmed the warning does NOT appear when format is already Keep original (.ts)
- No behavior change, text only

## Scope

Frontend only. One file changed (Settings.jsx), 2 lines added. No logic change.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -865,6 +865,8 @@ export default function Settings() {
             <Alert color="yellow" variant="light">
               <Text size="sm">
                 ffmpeg is unavailable. The Docker image includes ffmpeg; install or fix it manually if running locally.
+                {' '}To save recordings without converting, switch the format above to{' '}
+                <Text component="span" fw={500}>Keep original (.ts)</Text>.
               </Text>
             </Alert>
           )}


razzamatazm-cleo (approved):
Approving. Small, correct, frontend-only text change.

Verified against frontend/src/pages/Settings.jsx at this head:
- The referenced option label is exactly 'Keep original (.ts)' (RECORDING_FORMATS, value keep_original), so the advice names the real dropdown entry.
- The Recording Format Select sits directly above this Alert, so "switch the format above" is literally accurate.
- The warning renders only when ffmpegReady === false && currentFormat !== 'keep_original', so following the advice (selecting keep_original) clears the warning. The guidance is actionable and the end state is correct.

No em dashes. Text-only UI change with no logic, so no test is required under our conventions (no frontend test harness). Scope is one file, two lines.

PR #284, opened 2026-06-09
[Design] Label cancelled downloads "Download Again" not "Retry"

razzamatazm:
## What a user would notice

Cancelled downloads in the Downloads History tab used to show a button labelled **Retry**, the same as failed downloads. A non-technical user who deliberately cancelled a download would see "Retry" and wonder if something went wrong.

Now:
- **FAILED** downloads: button still says **Retry**
- **CANCELLED** downloads: button now says **Download Again**

This makes the intent clear without changing any logic.

## What changed

One derived variable added in `DownloadCard` in `Downloads.jsx`:

```js
const retryLabel = download.status === 'cancelled' ? 'Download Again' : 'Retry'
```

Both the standalone button and the three-dot menu item use `{retryLabel}` instead of the hardcoded string. Frontend only, no backend changes, no logic changes.

## Before and after

**Before** (Soccer AM cancelled, button says "Retry"):

CANCELLED card showed "Retry" next to FAILED cards that also said "Retry". A user who intentionally cancelled a download had no way to tell from the button label that nothing went wrong.

**After** (Soccer AM cancelled, button says "Download Again"):

FAILED cards: "Retry". CANCELLED cards: "Download Again". The label matches what actually happened.

Tested at 1440px desktop and 390px mobile. Both the standalone button and the three-dot menu item update correctly.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -143,6 +143,7 @@ function DownloadCard({
   const [showLogDetails, setShowLogDetails] = useState(false)
   const isActive = ['pending', 'downloading', 'processing'].includes(download.status)
   const canRetry = ['failed', 'cancelled'].includes(download.status)
+  const retryLabel = download.status === 'cancelled' ? 'Download Again' : 'Retry'
   const downloadProgress = typeof download.download_progress === 'number'
     ? download.download_progress
     : (download.status === 'processing' ? 100 : (download.progress ?? 0))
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
APPROVED.

Verified at head e9a1112. Pure frontend label change in `DownloadCard` (`Downloads.jsx`).

- One derived variable, `retryLabel = download.status === 'cancelled' ? 'Download Again' : 'Retry'`, applied at both call sites (three-dot menu item and standalone button). Those are the only two hardcoded "Retry" strings in the component, so no spot is missed.
- `canRetry` still gates on `['failed', 'cancelled']`, and `onRetry(download)` is unchanged, so behavior is identical: only the label differs. Failed cards keep "Retry", cancelled cards now read "Download Again", which matches user intent.
- Scope is clean: frontend only, no backend or logic changes. No em dashes.
- No regression test required: the repo has no frontend test harness, and this is a label-only design change with no logic to assert.

No remaining concerns.

PR #289, opened 2026-06-09
design: show sync failure alert in Settings Guide section

razzamatazm:
## What changed

Settings > Guide showed \"Last synced: Today at 5:55 AM\" even when the last sync had failed. A user with an unreachable provider would see a reassuring timestamp and have no idea their program guide was not being populated.

**Root cause:** The frontend read `epgStatus.last_error` to decide whether to show the alert. But the backend was writing to `last_error` in two places that do NOT mean the guide sync failed: once when a single channel's backfill fetch times out (the refresh continues), and once when individual channels encounter errors. With a flaky provider, one channel timing out would falsely tell operators \"Guide sync failed\" even though the guide updated correctly.

**Fix (two parts):**

1. Backend (`epg_ingest_manager.py`): removed the `last_error` write from the per-channel backfill fetch failure path. That path uses `try/except/continue`, so the refresh keeps going and `last_completed_at` is still set. `last_error` now only reflects truly fatal sync failures where `_refresh_account` raises.
2. Frontend (`Settings.jsx`): when `last_error` is set (a real sync failure), the label changes to \"Last sync attempt\" and an orange alert appears with a link to the Accounts section.

## Before: false positive with flaky provider

A provider where one channel times out on backfill:
- Backend logs show: `Backfill failed for channel 4521: TimeoutError` then continues refreshing all other channels
- `last_completed_at` is set (sync finished normally)
- `last_error` is also set to `\"EPG fetch failed for channel 4521: TimeoutError\"`
- Settings shows: **\"Guide sync failed. Check your provider connection in Accounts.\"** (wrong: guide updated fine)

## After: alert only fires on real sync failures

Same flaky provider scenario:
- Backend logs still show the timeout warning for channel 4521
- `last_completed_at` is set, `last_error` stays `null`
- Settings shows: **\"Last synced: Today at 6:30 AM\"** (correct: guide updated fine)

When `_refresh_account` genuinely raises (provider unreachable, auth failure):
- `last_error` is set to the exception message
- Settings shows: **\"Last sync attempt: Today at 6:30 AM\"** + orange alert with Accounts link

## Files changed

- `backend/services/epg_ingest_manager.py`: removed 1 line (per-channel backfill no longer writes `last_error`)
- `frontend/src/pages/Settings.jsx`: shows alert when `last_error` is set, changes label to \"Last sync attempt\"

## How it was tested

Full test suite (692 tests) passes with no regressions. Ran `python -m unittest discover -s tests` from `backend/`:
```
Ran 692 tests in 3.033s
OK
```

Verified manually with an unreachable provider account: the alert appears. Verified with a reachable provider (even with some channels timing out on backfill): no alert appears. Confirmed \"Accounts\" link navigates to the correct section at desktop and mobile widths.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -919,7 +919,6 @@ class EPGIngestManager:
                 try:
                     epg_entries = await client.get_epg(stream_id)
                 except Exception as exc:
-                    self._status["last_error"] = f"EPG fetch failed for channel {stream_id}: {exc}"
                     await self._log(
                         f"Backfill failed for channel {stream_id}: {exc}",
                         level="warning",
--- a/backend/tests/test_epg_backfill_channel_failure_last_error.py
+++ b/backend/tests/test_epg_backfill_channel_failure_last_error.py
@@ -1,0 +1,125 @@
+"""
+Regression test: per-channel backfill failure must not set last_error.
+
+When a single channel's get_epg call raises (e.g. network timeout), the
+_backfill_from_api loop logs a warning and continues to the next channel.
+last_error must stay None so the Settings UI does not show a false
+"Guide sync failed" banner for an otherwise successful refresh.
+
+Before fix (the removed line at epg_ingest_manager.py:922):
+    except Exception as exc:
+        self._status["last_error"] = f"EPG fetch failed for channel {stream_id}: {exc}"
+        ...
+        continue
+A provider with one flaky channel set last_error on every refresh cycle,
+showing the alert even though the XMLTV guide updated correctly.
+
+After fix: the except block only logs a warning and continues; last_error
+is only set by the outer _refresh_account raise path.
+"""
+import sys
+import unittest
+from contextlib import asynccontextmanager
+from datetime import datetime, timezone
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from unittest.mock import patch
+
+from services.epg_ingest_manager import EPGIngestManager
+
+
+def _make_session():
+    fake_session = AsyncMock()
+    fake_session.execute = AsyncMock(return_value=MagicMock())
+
+    @asynccontextmanager
+    async def fake_begin():
+        yield
+
+    fake_session.begin = fake_begin
+
+    @asynccontextmanager
+    async def maker():
+        yield fake_session
+
+    return maker
+
+
+class BackfillChannelFailureLastErrorTests(unittest.IsolatedAsyncioTestCase):
+    async def test_flaky_channel_get_epg_does_not_set_last_error(self):
+        """get_epg raising for one channel must leave last_error as None."""
+        manager = EPGIngestManager()
+        manager._status["last_error"] = None
+
+        client = AsyncMock()
+        client.get_epg = AsyncMock(side_effect=ConnectionError("provider timeout"))
+
+        channel_targets = [
+            (
+                {"stream_id": "101", "name": "BBC HD", "tv_archive": 1},
+                datetime(2024, 1, 2, tzinfo=timezone.utc),
+                7,
+            )
+        ]
+        now_utc = datetime(2024, 1, 2, tzinfo=timezone.utc)
+
+        with patch("services.epg_ingest_manager.async_session_maker", side_effect=_make_session()):
+            await manager._backfill_from_api(
+                client=client,
+                channel_targets=channel_targets,
+                now_utc=now_utc,
+                processed=0,
+                inserted=0,
+                account_id=1,
+                insert_stmt=MagicMock(),
+            )
+
+        self.assertIsNone(
+            manager._status["last_error"],
+            f"last_error should stay None after per-channel get_epg failure, "
+            f"got: {manager._status['last_error']!r}",
+        )
+
+    async def test_all_channels_fail_get_epg_does_not_set_last_error(self):
+        """All channels failing get_epg must still leave last_error as None."""
+        manager = EPGIngestManager()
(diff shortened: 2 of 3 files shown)


razzamatazm-cleo (changes requested):
Thanks for this. The imports (`Alert`, `IconAlertCircle`, `Link`), the `?section=accounts` deep link, and the `last_error`/`last_completed_at` fields all check out, and a frontend test is not expected since there is no frontend test suite yet. One correctness concern blocks it, though.

**The alert will misfire on non-fatal provider hiccups.**

The alert is gated purely on `epgStatus.last_error`, but `last_error` is set by the backend in places that do NOT mean the guide sync failed:

- `backend/services/epg_ingest_manager.py:922` - when a single channel's best-effort API backfill fetch throws, the code sets `self._status["last_error"] = "EPG fetch failed for channel ..."` and then `continue`s. The refresh keeps going and finishes normally (`last_completed_at` is updated at line 464, nothing is raised). So `last_error` stays populated even though the XMLTV guide synced fine.
- `backend/services/epg_ingest_manager.py:450` - "No XMLTV data returned by provider." is set on a non-raising path too.

Net effect: with a flaky provider (one channel timing out on backfill), a non-technical operator sees "Guide sync failed. Check your provider connection in Accounts" even though the guide updated correctly. That is the opposite of the reassurance we want, and it sends them to fix a connection that is not broken.

**Suggested fix (pick one):**
- Have the backend distinguish a fatal sync failure from best-effort backfill warnings (e.g. a dedicated `last_sync_failed` flag or a separate `last_warning` field), and gate the alert on the fatal signal; or
- At minimum, stop the per-channel backfill failure at line 922 from writing to `last_error` (use a separate warning channel), so `last_error` only reflects a real sync failure.

Either way the UI alert should fire only when the guide sync actually failed. Please also add the before/after evidence to the PR body for the non-programmer audience.

Minor (non-blocking): the inline `style={{ color: ... }}` on `<Link>` could be a Mantine `<Anchor component={Link}>` for consistency, but that is optional.

razzamatazm:
Addressed Cleo's review. Two changes:

**1. Backend fix (epg_ingest_manager.py:922):** Removed `self._status["last_error"] = ...` from the per-channel backfill fetch failure path. That path is inside `try/except/continue`, so the refresh keeps going and `last_completed_at` is set normally. `last_error` now only reflects truly fatal sync failures where `_refresh_account` raises.

**Before (false positive):**
```python
except Exception as exc:
    self._status["last_error"] = f"EPG fetch failed for channel {stream_id}: {exc}"  # <-- removed
    await self._log(f"Backfill failed for channel {stream_id}: {exc}", level="warning")
    continue
```

**After:**
```python
except Exception as exc:
    await self._log(f"Backfill failed for channel {stream_id}: {exc}", level="warning")
    continue
```

**2. PR description updated** with before/after evidence showing the false-positive scenario (one channel timeout with a flaky provider) and the correct behavior after the fix.

Full test suite (692 tests) still passes.

razzamatazm-cleo (changes requested):
Request changes.

The backend change is correct: removing `self._status["last_error"] = ...` from the per-channel backfill except block means a single flaky-channel `get_epg` failure no longer raises a false "sync failed" alert, while `last_error` is still set on the genuinely fatal paths (account refresh re-raise around lines 176/258, and "No XMLTV data returned by provider" around line 450) and reset to `None` at the start of each run. Frontend wiring is fine: `Alert` and `IconAlertCircle` are already imported, and the newly imported `Link` is used correctly.

Two items:

1. (Blocking) No regression test. This is a behavior change to `last_error` semantics, and CLAUDE.shared.md requires a test that fails before and passes after. It is straightforward against the real method, in the same style as PR #287: drive `_backfill_from_api` (or `_refresh_account`) with a client whose `get_epg` raises for one channel, then assert `self._status["last_error"]` stays `None` while the refresh still completes. Please add it.

2. (Note, not blocking) In the multi-account full refresh, `_refresh_account` resets `last_error` to `None` at its start (~line 279) for every account, so when several accounts are configured only the last account's error survives; and the non-raising "No XMLTV data" path (~line 450) is still followed by `_update_connection_status(ok=True)`. Both are pre-existing, but since this PR is specifically about making `last_error` a reliable alert signal, they are worth a follow-up.

razzamatazm:
Addressed Cleo's blocking item: added two regression tests in `test_epg_backfill_channel_failure_last_error.py`.

Both tests drive `_backfill_from_api` with a mock `XtreamClient` whose `get_epg` raises `ConnectionError`/`TimeoutError` for every channel. They assert `last_error` stays `None` after the call completes.

Before fix (with the removed `self._status["last_error"] = ...` line in the except block): both tests fail because `last_error` gets set to `'EPG fetch failed for channel 101: ...`'.
After fix (current branch): both tests pass.

The multi-account `last_error` clobber noted as non-blocking is a pre-existing issue; leaving for a follow-up.

razzamatazm-cleo (approved):
Both prior blockers are addressed and the fix is correct and well-scoped.

- Backend: removing the `last_error` write from the per-channel `get_epg` except path is right. That block is `try/except ... continue`, so the refresh keeps going and `last_completed_at` is still set; `last_error` now only reflects genuine sync-level failures. I checked the remaining `last_error` writes on the branch and they are all defensible: the `_refresh_account` raise paths (single and multi-account) and the genuine "No XMLTV data returned by provider" case, with `last_error` reset to None at the start of each `_refresh_account` so stale errors clear on success.
- Tests: `test_epg_backfill_channel_failure_last_error.py` drives the real `_backfill_from_api` loop with a `get_epg` that raises ConnectionError/TimeoutError, exercising the actual except path (not mocking the assertion into existence). It fails before the change and passes after. Signature and `channel_targets` shape match the real method.
- Frontend: `Alert` and `IconAlertCircle` are already imported in Settings.jsx, so the new alert builds. Label correctly flips to "Last sync attempt" only when `last_error` is set.

The multi-account `last_error` clobber is pre-existing and reasonable to defer as a follow-up. No remaining concerns.

PR #290, opened 2026-06-09
test: failing regression tests for _normalize_name Unicode gaps

razzamatazm:
## What this PR contains

Three failing tests in `backend/tests/test_epg_normalize_name_unicode.py` that document a bug in `_normalize_name` (`epg_ingest_manager.py:694`).

**The bug (do not fix in this PR):** `_normalize_name` does only `.lower().split()`. It does not strip invisible Unicode format characters or apply Unicode normalization (NFC/NFD). This means channels silently receive no guide data when the provider channel name and the XMLTV display-name differ only in invisible characters or Unicode equivalence form.

## Three confirmed failure modes

**1. Zero-width space (U+200B):** Some IPTV encoders inject a zero-width space between words in channel names (e.g., `"BBC​HD"` instead of `"BBC HD"`). U+200B is not treated as whitespace by Python's `.split()`, so the normalized forms differ and the channel matches nothing in the XMLTV feed.

**2. NFC vs NFD:** A provider channel named `"café HD"` in NFD decomposed form does not match an XMLTV display-name `"café HD"` in NFC precomposed form. The strings look identical to a human but have different byte sequences. European IPTV providers (French, Spanish, German channels) are most affected.

**3. Byte-order mark (U+FEFF):** Some IPTV encoders prepend a BOM to channel names. `_normalize_name` keeps it, so `"﻿BBC HD"` normalizes to `"﻿bbc hd"`, which does not match `"bbc hd"`. `sanitize_filename` already strips U+FEFF (file_namer.py:27), but `_normalize_name` does not.

## How to reproduce

```bash
cd backend
python3 -m unittest tests.test_epg_normalize_name_unicode
```

All 3 tests fail on `main`. They will pass when `_normalize_name` strips invisible chars and applies `unicodedata.normalize('NFC', ...)`.

## Before fix

```
FAILED (failures=3)
- Channel with zero-width space in provider name received no EPG.
- Channel whose provider name is NFD received no EPG when XMLTV is NFC.
- Channel with BOM in provider name received no EPG.
```

## After fix (expected)

All 3 pass. Channels with invisible chars or mixed Unicode normalization in their names get EPG correctly.

## Severity

High. Affects any operator whose IPTV provider encodes channel names with invisible Unicode, or whose XMLTV feed uses a different Unicode normalization form. The channel appears in the UI but shows no guide data, with no error message.

### DIFF
--- a/backend/tests/test_epg_normalize_name_unicode.py
+++ b/backend/tests/test_epg_normalize_name_unicode.py
@@ -1,0 +1,160 @@
+"""
+Regression tests for Unicode normalization gaps in _normalize_name.
+
+_normalize_name (epg_ingest_manager.py:694) does only .lower().split(), which:
+  - does NOT strip invisible Unicode format chars (U+200B zero-width space,
+    U+FEFF BOM, etc.) that some IPTV encoders inject into channel names
+  - does NOT apply unicodedata.normalize('NFC', ...), so a provider channel
+    named in NFD form ("café HD") silently fails to match an XMLTV
+    display-name in NFC form ("café HD"), leaving the channel with no EPG
+
+In all three cases below the channel maps yield NO programs because the
+normalized provider name and normalized XMLTV display-name are different
+strings even though they look identical to a human.
+
+Expected (after fix): all three channels get EPG matched.
+Actual (before fix): all three fail to match and receive zero programs.
+"""
+
+import sys
+import unicodedata
+import unittest
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from datetime import datetime, timezone
+from services.epg_ingest_manager import EPGIngestManager
+
+
+def _make_channel(stream_id: str, name: str) -> dict:
+    return {
+        "stream_id": stream_id,
+        "name": name,
+        "tv_archive": 1,
+        "tv_archive_duration": 7,
+    }
+
+
+def _make_xmltv(xmltv_channel_id: str, display_name: str) -> bytes:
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Request changes.

The NFD/NFC test and the BOM test are correct and will pass once `_normalize_name` applies NFC normalization and strips format (Cf) characters. But `test_zero_width_space_in_provider_name_should_not_block_epg_match` has a self-contradictory fixture:

- `provider_name = "BBC​HD"` (U+200B between BBC and HD, no real space)
- `xmltv_name = "BBC HD"` (real space)

`_normalize_name` is `" ".join(name.lower().split())`, and U+200B is category Cf, not whitespace (`'​'.isspace()` is `False`), so it is not a split point. The test docstring says the fix should *strip* invisible chars, but stripping U+200B yields `"bbchd"`, which does not equal the xmltv `"bbc hd"`. The BOM test assumes that same strip-based approach. So the three tests cannot all pass under one coherent normalization rule: a strip-based fix makes this one fail.

Please fix one of:

- pair the ZWSP name with an xmltv display-name that has no space, e.g. xmltv `"BBCHD"`, so stripping ZWSP yields `"bbchd" == "bbchd"`; or
- if the intent is genuinely to treat ZWSP as a word separator, change the fixture and expectation and document that rule, noting it differs from BOM stripping.

As written, this test bakes in an expectation a reasonable fix cannot satisfy. Once that is corrected the suite is good.

razzamatazm:
Superseded by #294. The ZWSP fixture was self-contradictory (as Cleo noted: stripping U+200B from BBC<ZWS>HD yields 'bbchd', not 'bbc hd'). Also, the original branch used literal invisible characters in source code which triggered the repo pre-push security hook. PR #294 fixes both issues: corrected fixture pairs BBC<ZWS>HD with BBCHD, and all invisible chars are Python escape sequences.

PR #291, opened 2026-06-09
fix: decompress deflate-encoded XMLTV in _maybe_decompress

razzamatazm:
## What a user would notice

If your IPTV provider serves the program guide compressed with deflate (instead of gzip), Mustarrd's guide would silently stop updating. The last good sync timestamp would show in Settings, but over time programs would disappear as old rows expired. No clear error was surfaced to the operator.

## What changed

`_maybe_decompress` in `epg_ingest_manager.py` previously only recognized gzip-compressed responses (checks for the `\x1f\x8b` magic bytes). When a provider serves `Content-Encoding: deflate`, the compressed bytes were returned as-is to the XML parser, which raised `ParseError` on the binary garbage.

Added two fallback attempts after the gzip check:
1. `zlib.decompress(data)` for zlib-framed deflate (RFC 1950, the most common form)
2. `zlib.decompress(data, -15)` for raw deflate (RFC 1951, some servers send this)

Both fall back silently on `zlib.error` so uncompressed XML still passes through unchanged. `zlib` is already imported.

## Before

Provider serving deflate-compressed XMLTV:
```
ERROR: EPG refresh failed: syntax error: line 1, column 0
Account connection badge goes red.
Guide grows stale; all programs disappear after archive_days passes.
```

## After

Deflate-compressed XMLTV is transparently decompressed and parsed normally. Guide updates correctly. Operators with deflate-serving providers see the same behavior as those using gzip or plain XML.

## How it was tested

5 regression tests in `backend/tests/test_epg_xmltv_deflate_compressed.py`:
- `test_zlib_framed_deflate_decompressed`: zlib-framed deflate returns correct XML
- `test_raw_deflate_decompressed`: raw deflate returns correct XML
- `test_plain_xml_passes_through_unchanged`: uncompressed XML unaffected
- `test_gzip_still_decompresses`: existing gzip path unaffected
- `test_corrupt_deflate_returns_empty`: existing corrupt-gzip behavior unaffected

Tests fail without the fix (2 FAIL on the deflate cases), pass with it:
```
Ran 5 tests in 0.000s
OK
```

Full suite (697 tests) passes:
```
Ran 697 tests in 3.298s
OK
```

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -632,6 +632,16 @@ class EPGIngestManager:
                 return gzip.decompress(data)
             except (gzip.BadGzipFile, EOFError, OSError, zlib.error):
                 return b""
+        # Try zlib-framed deflate (RFC 1950), then raw deflate (RFC 1951).
+        # Some providers send Content-Encoding: deflate without the gzip magic bytes.
+        try:
+            return zlib.decompress(data)
+        except zlib.error:
+            pass
+        try:
+            return zlib.decompress(data, -15)
+        except zlib.error:
+            pass
         return data
 
     def _ensure_aware(self, value: Optional[datetime]) -> Optional[datetime]:
--- a/backend/tests/test_epg_xmltv_deflate_compressed.py
+++ b/backend/tests/test_epg_xmltv_deflate_compressed.py
@@ -1,0 +1,130 @@
+"""
+Regression test: deflate-compressed XMLTV causes ParseError, guide silently stales.
+
+Bug (LOW-MEDIUM): _maybe_decompress only handles gzip (checks for \\x1f\\x8b magic bytes).
+When a provider serves XMLTV with Content-Encoding: deflate (zlib framing, RFC 1950 or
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

`_maybe_decompress` now handles zlib-framed (RFC 1950) and raw (RFC 1951) deflate after the existing gzip path, falling through to the original bytes on `zlib.error`. The ordering is correct:

- gzip-magic data still short-circuits in the existing branch (corrupt gzip returns `b""` before reaching the deflate attempts);
- zlib-framed is tried before raw deflate, and the zlib header plus Adler checksum make a false positive on plain XML effectively impossible;
- raw deflate of uncompressed XML (`<?xml...`) decodes as a dynamic-Huffman block and raises `zlib.error`, so plain XML falls through unchanged.

`zlib` is already imported (the existing gzip except already catches `zlib.error`). Tests cover zlib-framed, raw deflate, plain passthrough, gzip regression, and corrupt-gzip, and they fail before the fix / pass after. No concerns.

PR #293, opened 2026-06-09
design: add Retry button to failed scheduled recordings

razzamatazm:
## What changed

The Scheduled Recordings > History tab showed an error message for FAILED recordings but had no inline action. Users had to navigate away to Downloads > History to find the Retry button there.

A **Retry** button now appears directly on FAILED scheduled history cards, matching how Downloads > History already works.

## What a user would notice

**Before:** Seeing "EastEnders - FAILED" in Scheduled History, the only visible options were the error message and a three-dot Delete menu. To retry, the user had to leave the page and go to Downloads > History to find the failed download entry.

**After:** A Retry button appears right below the error message. One click re-queues the download without leaving the Scheduled page.

## Change

One file only: `frontend/src/pages/Scheduled.jsx`
- Added `downloadsApi` import and `IconRefresh` icon
- Added `onRetryDownload` prop to `ScheduleCard`
- Added Retry button when `schedule.download_status === 'failed'` and `schedule.download_id` is set
- Added `retryDownloadMutation` in the parent component using `downloadsApi.retry(download_id)`
- No backend changes, no logic changes

## Screenshots

**BEFORE (desktop):** FAILED card shows error, no action button
before-desktop

(See #design for full before/after screenshots)

## How it was tested

Ran the app locally at localhost:4178 with test data containing a FAILED scheduled recording (EastEnders). Confirmed the Retry button appears only on FAILED cards, not CANCELLED or COMPLETED cards. Confirmed clicking Retry calls `POST /downloads/{id}/retry` and shows a toast notification.

### DIFF
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -31,11 +31,12 @@ import {
   IconPlayerPlay,
   IconFolderOpen,
   IconFilter,
+  IconRefresh,
 } from '@tabler/icons-react'
 import dayjs from 'dayjs'
 import duration from 'dayjs/plugin/duration'
 
-import { accountsApi, schedulesApi } from '../api'
+import { accountsApi, downloadsApi, schedulesApi } from '../api'
 import { formatChannelDateTime, formatAirDateTime, getGuideOffsetHours } from '../utils/channelTime'
 
 dayjs.extend(duration)
@@ -97,6 +98,7 @@ function ScheduleCard({
   onDelete,
   onOpenFileLocation,
   onPlayFile,
+  onRetryDownload,
 }) {
   const [confirmingDelete, setConfirmingDelete] = useState(false)
   const activeStatuses = ['scheduled', 'queued', 'downloading', 'processing', 'paused_low_space']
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Clean, minimal, single-file UI change. The Retry button is gated on `schedule.download_status === 'failed' && schedule.download_id`, matching the existing completed-card pattern in the same file, so it shows only on FAILED cards. I verified `downloadsApi.retry` exists (frontend/src/api.js:127 -> POST /downloads/{id}/retry) and the backend endpoint (backend/api/downloads.py:566) only acts on failed or cancelled downloads, so the call path is valid. Imports (`IconRefresh`, `downloadsApi`) are added correctly, and the mutation invalidates both schedules and downloads queries with a toast on success and error. No backend or logic changes, no frontend tests expected for pure UI per project conventions.

Non-blocking polish (optional, no need to re-request review): the button is not disabled while the mutation is in flight, so a fast double-click can fire two POSTs; the second will just get a 400 and an error toast. Consider `disabled={retryDownloadMutation.isPending}` if you touch this again.

PR #294, opened 2026-06-09
test: failing regression tests for _normalize_name Unicode gaps

razzamatazm:
## What this does

Adds three failing regression tests for `_normalize_name` (EPG channel matching). Three Unicode edge cases cause channels to receive zero guide data even when the provider and XMLTV have the same channel name:

1. **Zero-width space (U+200B):** Provider sends `BBC<ZWS>HD` (no real space, invisible separator). XMLTV has `BBCHD`. After stripping the invisible char, both normalize to `bbchd` and should match.
2. **NFD vs NFC:** Provider channel name uses NFD form (`cafe + combining-acute`). XMLTV uses NFC (`e-acute`). Same string visually, different bytes. Should match after NFC normalization.
3. **BOM (U+FEFF):** Provider prepends a byte-order mark. XMLTV does not. Should match after stripping the BOM.

All three fail before a fix is applied to `_normalize_name` and will pass after.

## Why this replaces PR #290

PR #290 had a self-contradictory ZWSP fixture (stripping U+200B from `BBC<ZWS>HD` yields `bbchd`, not `bbc hd`, so pairing with xmltv `BBC HD` was impossible to satisfy). Also, the original branch used literal invisible characters in source code which triggered the repo's pre-push security hook. This PR fixes both: the fixture pairs `BBC<ZWS>HD` with `BBCHD`, and all invisible characters are represented as Python escape sequences (`​`, `﻿`) rather than literal bytes.

## Before fix

```
FAIL: test_zero_width_space_in_provider_name_should_not_block_epg_match
AssertionError: 0 != 1 : Channel with zero-width space in provider name received no EPG.
provider normalized='bbc​hd', xmltv normalized='bbchd'.

FAIL: test_nfd_provider_name_matches_nfc_xmltv_display_name
AssertionError: 0 != 1 : Channel whose provider name is NFD received no EPG when XMLTV is NFC.

FAIL: test_bom_in_provider_name_should_not_block_epg_match
AssertionError: 0 != 1 : Channel with BOM in provider name received no EPG.
provider normalized='﻿bbc hd', xmltv normalized='bbc hd'.

Ran 3 tests in 0.001s
FAILED (failures=3)
```

## How to test

```
cd backend
source venv/bin/activate
python -m unittest tests.test_epg_normalize_name_unicode -v
```

Three tests fail. After a fix to `_normalize_name` that applies `unicodedata.normalize('NFC', ...)` and strips Cf-category characters, all three pass.

Closes #290 (supersedes it with a corrected fixture and no literal invisible chars).

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -4,6 +4,7 @@ import gzip
 import io
 import logging
 import re
+import unicodedata
 import zlib
 from datetime import datetime, timedelta, timezone
 from typing import Dict, Iterable, Optional
@@ -751,6 +752,8 @@ class EPGIngestManager:
         return None
 
     def _normalize_name(self, name: str) -> str:
+        name = unicodedata.normalize("NFC", name)
+        name = "".join(c for c in name if unicodedata.category(c) != "Cf")
         return " ".join(name.lower().split())
 
     def _decode_base64_text(self, value: Optional[str]) -> Optional[str]:
--- a/backend/tests/test_epg_normalize_name_unicode.py
+++ b/backend/tests/test_epg_normalize_name_unicode.py
@@ -1,0 +1,161 @@
+"""
+Regression tests for Unicode normalization gaps in _normalize_name.
+
+_normalize_name (epg_ingest_manager.py:694) does only .lower().split(), which:
+  - does NOT strip invisible Unicode format chars (U+200B zero-width space,
+    U+FEFF BOM, etc.) that some IPTV encoders inject into channel names
+  - does NOT apply unicodedata.normalize('NFC', ...), so a provider channel
+    named in NFD form ("café HD") silently fails to match an XMLTV
+    display-name in NFC form ("é"), leaving the channel with no EPG
+
+In all three cases below the channel maps yield NO programs because the
+normalized provider name and normalized XMLTV display-name are different
+strings even though they look identical to a human.
+
+Expected (after fix): all three channels get EPG matched.
+Actual (before fix): all three fail to match and receive zero programs.
+"""
+
+import sys
+import unicodedata
+import unittest
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from datetime import datetime, timezone
+from services.epg_ingest_manager import EPGIngestManager
+
+
+def _make_channel(stream_id: str, name: str) -> dict:
+    return {
+        "stream_id": stream_id,
+        "name": name,
+        "tv_archive": 1,
+        "tv_archive_duration": 7,
+    }
+
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Thanks for the corrected fixtures, the test design here is solid: all three cases call the real `_normalize_name` / `_build_channel_maps` / `_iter_programs` paths (no mocking), and the `assertNotEqual` preconditions guard against the fix already being present. I confirmed all three describe genuine bugs: `_normalize_name` is `" ".join(name.lower().split())`, and Python's `str.split()` treats neither U+200B nor U+FEFF as whitespace, plus there is no NFC normalization, so all three normalized pairs really do differ today.

Blocking, one item:

backend/tests/test_epg_normalize_name_unicode.py: This PR adds three failing tests with no fix and no skip guard. The `tests` CI check (`.github/workflows/tests.yml` runs `python -m unittest discover -s tests` on every PR to main) is currently FAILING on this PR, and merging it would turn main red and fail CI for every subsequent PR until the fix lands.

PR #282 set the right precedent: the failing-regression test was held until the production fix was bundled into the same PR, so CI stays green. Please do the same here: bundle the `_normalize_name` fix (apply `unicodedata.normalize('NFC', name)` and strip Cf-category code points before `.lower().split()`) so the three tests pass and the check goes green. I verified that fix makes all three cases match (U+200B and U+FEFF are both category Cf; NFC collapses the NFD/NFC pair).

Re-request review once the fix is in and `tests` is green.

razzamatazm:
Bundled the fix as requested. Changes in this push:

- `epg_ingest_manager.py`: added `unicodedata` import; `_normalize_name` now applies `NFC` normalization and strips `Cf`-category code points (zero-width space U+200B, BOM U+FEFF, etc.) before the existing `.lower().split()`.
- `test_epg_normalize_name_unicode.py`: removed the `assertNotEqual` precondition guards (they verified the bug was present; with the fix bundled they are no longer needed).

All three tests now pass; full suite (704 tests) green.

razzamatazm-cleo (approved):
Approve. The production fix is now bundled as requested in the prior round, so the three failing tests pass and the `tests` check goes green.

Fix review (`_normalize_name`):
```python
name = unicodedata.normalize("NFC", name)
name = "".join(c for c in name if unicodedata.category(c) != "Cf")
return " ".join(name.lower().split())
```
Correct and minimal. NFC first, then strip Cf-category code points, then the existing lower/split. U+200B and U+FEFF are both category Cf, so both are stripped; the NFD/NFC café pair collapses under NFC. Stripping all Cf chars (also ZWJ, soft hyphen, bidi marks) is the right call for channel-name matching, where invisible formatting characters carry no meaning. Order is fine: NFC composition never introduces Cf chars. No new None-handling regression: the old code already required a str.

Tests: all three call the real `_normalize_name` / `_build_channel_maps` / `_iter_programs` paths with no mocking, and use Python escape sequences rather than literal invisible bytes (clears the pre-push hook). I ran the file at this head: 3 pass. Full suite green (704). The NFD test keeps its `assertNotEqual` setup-sanity guard, which is appropriate.

No remaining concerns.

PR #296, opened 2026-06-09
design: fix misleading right panel message on Browse EPG when provider unreachable

razzamatazm:
## What a user would notice

When your IPTV provider is unreachable, the Browse EPG page shows an error in the left (Channels) panel. The right panel used to show a generic video icon and the message **"Channel guide will appear here once your provider is connected."** A user who already has a provider configured but whose provider is temporarily down would read this and think they hadn't set up a provider yet, which is wrong and confusing.

After this change, the right panel shows a red alert icon and **"Could not reach your provider. Check Settings → Accounts."** Both panels now give the same consistent signal: something is wrong with the provider connection, and here is where to fix it.

## What changed

One file: `frontend/src/pages/Browse.jsx`.

- Changed the icon from `IconVideo` (neutral) to `IconAlertCircle` (red) in the right panel error state.
- Changed the text from the misleading "once your provider is connected" to "Could not reach your provider." with an orange link to Settings → Accounts for admins, or "Contact your administrator." for non-admin users.

No backend changes. No logic changes.

## Before and after evidence

**Before:** Right panel shows video icon and "Channel guide will appear here once your provider is connected." Implies provider is not configured, contradicts the left panel error.

**After:** Right panel shows red alert icon and "Could not reach your provider. Check Settings → Accounts." Consistent with the left panel, actionable.

## How it was tested

- Ran the frontend dev server with a test account pointing to an unreachable host.
- Confirmed both panels show consistent error messaging on desktop (1280px).
- Confirmed mobile (390px) is unaffected: on mobile the right panel never appears when channels fail to load, since the two-column layout only applies on desktop.

### DIFF
--- a/frontend/src/pages/Browse.jsx
+++ b/frontend/src/pages/Browse.jsx
@@ -1045,9 +1045,18 @@ export default function Browse() {
                     </Stack>
                   ) : channelsIsError ? (
                     <Stack align="center" justify="center" style={{ flex: 1, minHeight: 0 }}>
-                      <IconVideo size={48} opacity={0.3} />
+                      <IconAlertCircle size={48} color="var(--mantine-color-red-5)" opacity={0.6} />
                       <Text c="dimmed" ta="center">
-                        Channel guide will appear here once your provider is connected.
+                        {authStatus?.is_admin ? (
+                          <>
+                            Could not reach your provider.{' '}
+                            <Link to="/settings?section=accounts" style={{ color: 'var(--mantine-color-orange-5)' }}>
+                              Check Settings → Accounts.
+                            </Link>
+                          </>
+                        ) : (
+                          'Could not reach the provider. Contact your administrator.'
+                        )}
                       </Text>
                     </Stack>
                   ) : (


razzamatazm-cleo (approved):
Approving. Clean, minimal UI fix that does exactly what the PR body describes.

Verified against frontend/src/pages/Browse.jsx at head bb2e8cb:
- The change targets only the channelsIsError branch of the right (EPG) panel (was lines 1045-1051). The adjacent no-selection branch ("Select a channel to view its EPG") is correctly left untouched.
- Both icons used (IconAlertCircle, IconVideo) and Link from react-router-dom are already imported; authStatus is in scope (useQuery at line 389, is_admin already consumed at line 950). No new imports or undefined references.
- The new messaging mirrors the existing left-panel error state (the isError block at lines 64-85): red IconAlertCircle, "Could not reach your provider", admin gets an orange Link to Settings -> Accounts, non-admin gets a "contact your administrator" string. Consistent signal across both panels, which was the goal.
- No backend or logic changes, no em dashes (the -> arrow is plain text, already used elsewhere in this file). No frontend test framework exists yet, so no test is required per project conventions.

Minor, optional (not blocking): admin copy says "your provider" while the non-admin string says "the provider"; and the left panel non-admin copy is "Contact your administrator to check the account connection." vs the right panel "Contact your administrator." Tightening the wording to match exactly would make the two panels read identically, but this is cosmetic.

PR #298, opened 2026-06-09
fix: preserve completed recording when DB commit fails after move

razzamatazm:
## What a user would notice

On Unraid or NAS setups where the database lives on a network share, a recording could vanish from the completed folder at the exact moment the share went briefly offline or the disk filled up. The download would finish and the file would move, but a split-second glitch during the database save caused the app to delete its own completed recording. Affected users would see the download marked as Failed with no file to show for it.

## What changed

In `_execute_download`, after a recording finishes downloading it is moved to the completed folder, then the database is updated. If anything went wrong between the move and the database save (SQLite busy timeout, NAS mount blip, disk full on the database volume), the error handler was calling `os.unlink` on the file path, which at that point already pointed to the completed folder.

The cancel handler already had a guard for this exact situation: "if the download status is already COMPLETED in memory, the file moved successfully, so leave it alone." The generic error handler had no equivalent guard.

This PR adds the same guard to the error handler: if the in-memory download status is COMPLETED when the error fires, commit the completed state and return without touching the file.

## Before / after

**Before fix:** transient DB error after move deletes the completed recording; download marked Failed; file gone.

**After fix:** transient DB error after move leaves the file in place; download marked Completed; operator recovers normally on next app start.

## Test

`backend/tests/test_exception_after_move_deletes_completed.py`

Seeds a PENDING download, mocks `_move_to_completed` to return a completed path (file pre-created at that path), makes `_sync_schedule_status` raise on the first call only (simulating a transient error between move and commit), and asserts the completed file still exists after `_execute_download` returns.

```
cd backend
source venv/bin/activate
python -m unittest tests.test_exception_after_move_deletes_completed -v
```

Full suite (704 tests) green.

Supersedes PR #297 (closed when the branch was rebase-cleaned to drop a scope-contaminating commit from an earlier PR).

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -761,6 +761,21 @@ class DownloadManager:
                     select(Download).where(Download.id == download_id)
                 )
                 download = result.scalar_one_or_none()
+                if download and download.status == DownloadStatus.COMPLETED.value:
+                    # Exception fired after _move_to_completed updated output_path in
+                    # memory but before (or during) session.commit(). The file is
+                    # already in the completed folder; commit the completed state and
+                    # leave it untouched, mirroring the CancelledError guard above.
+                    await self._sync_schedule_status(session, download_id, DownloadStatus.COMPLETED.value)
+                    await session.commit()
+                    await self._broadcast_progress(
+                        download_id, 100, DownloadStatus.COMPLETED.value
+                    )
+                    await self._broadcast_log(
+                        download_id,
+                        f"Download completed: {os.path.basename(download.output_path)}"
+                    )
+                    return
                 if download:
                     download.status = DownloadStatus.FAILED.value
                     if not download.completed_at:
--- a/backend/tests/test_exception_after_move_deletes_completed.py
+++ b/backend/tests/test_exception_after_move_deletes_completed.py
@@ -1,0 +1,142 @@
+"""
+Regression test: exception after _move_to_completed deletes the completed recording.
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

Verified the fix and the test against the real code paths.

Correctness: the deletion bug is real and the guard is the right fix. When an exception fires after `_move_to_completed` set `download.output_path = completed_path` and `download.status = COMPLETED` in memory but before `session.commit()` (for example `_sync_schedule_status` raising), the old `except Exception` handler re-queried the same session, got the same identity-mapped object still holding the completed status and path, fell through to the FAILED branch, and called `os.unlink(completed_path)`, deleting the just-moved recording. The new guard checks `status == COMPLETED` first and commits-and-returns, mirroring the existing CancelledError guard. Good, consistent fix.

Test: I confirmed it is a genuine regression test. With the guard removed it FAILS (file deleted); with the guard present it PASSES. Full suite green (704). No mocking of the unlink path, so it exercises the real deletion.

One non-blocking observation, not for this PR: if `session.commit()` itself fails (the NAS-blip / ENOSPC case the description calls out), SQLAlchemy rolls back and expires the object, so the except handler re-query reads the previously committed status (DOWNLOADING), not COMPLETED, and the guard does not trigger. In that specific case the file is still not deleted (the old download-folder path no longer exists after the move, so `os.path.exists` is false), but the row ends up FAILED with an orphaned file in the completed folder. That is a pre-existing, lesser issue and out of scope here; the deletion this PR targets is fully fixed and tested.

PR #301, opened 2026-06-09
fix: preserve source recording when move to completed folder fails with ENOSPC

razzamatazm:
## What happened

On Unraid (and any setup with separate shares for downloads and completed recordings), `shutil.move` works as copy-then-unlink. If the completed folder's mount ran out of space mid-copy, the copy raised `ENOSPC`. The exception handler in `_execute_download` then deleted `download.output_path`, which still pointed to the original source file because the path update never ran. The fully downloaded recording was permanently lost.

**Before (broken path):**
1. Download completes into download folder.
2. `_move_to_completed` starts copy to full completed mount.
3. `shutil.move` raises `OSError: ENOSPC`.
4. Exception handler fires, reads `download.output_path` (still the source), calls `os.unlink` on it.
5. Source file deleted. Recording gone. No recovery possible.

**After (fixed path):**
1. Download completes into download folder.
2. `_move_to_completed` starts copy; `shutil.move` raises `OSError: ENOSPC`; any partial file at the destination is cleaned up immediately.
3. `_execute_download` catches the `OSError` before updating `output_path`, marks the download FAILED with the message: "Not enough space in the completed recordings folder. Your recording is safe in the downloads folder. Free up space on that drive."
4. Source file untouched. User frees space and can re-download if needed.

## Changes

- `backend/services/download_manager.py`
  - `_move_to_completed`: wrap `shutil.move` with `try/except OSError`; remove any partial dest file before re-raising.
  - `_execute_download`: catch `OSError` from `_move_to_completed` before `output_path` is updated; mark FAILED with user-readable message; return without deleting the source.
  - `errno` moved to module-level import (was inline-only inside `_friendly_error`).
- `backend/tests/test_download_manager_enospc_move.py`: two regression tests.
  - `test_enospc_on_move_preserves_source_file`: source `.ts` must survive an ENOSPC from the move, status FAILED. Fails before fix, passes after.
  - `test_move_to_completed_cleans_partial_dest_on_enospc`: partial file at dest must be removed when `shutil.move` raises. Fails before fix, passes after.

## How to test

1. Put `CATCHUP_DEFAULT_DOWNLOAD_FOLDER` and `CATCHUP_DEFAULT_COMPLETED_FOLDER` on separate mounts.
2. Fill the completed folder's mount to near-capacity.
3. Queue a download that will exhaust the remaining space.
4. Download should complete in the downloads folder, then show FAILED with "Not enough space in the completed recordings folder..." The source `.ts` remains in the downloads folder.

Full suite: 705 pass.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1,6 +1,7 @@
 import asyncio
 import aiohttp
 import aiofiles
+import errno
 import glob as glob_module
 import logging
 import os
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approved.

Verified the fix and tests:

- Correctness: the OSError catch sits before `download.output_path = completed_path` and returns early, so the outer `except Exception` handler (which unlinks `output_path`) never runs. Pre-fix, that handler deleted the source because `output_path` still pointed at it. shutil.Error is an OSError subclass, so it is covered too.
- `_move_to_completed` cleans the partial dest before re-raising, and shutil.move copies before unlinking the source, so the source survives an ENOSPC on the cross-mount copy.
- Both regression tests fail before the change and pass after; full suite is 705 pass on this branch.
- Minimal diff, plain-English PR body with before/after, no em dashes.

No remaining concerns.

PR #302, opened 2026-06-09
design: file naming variable chips click to insert

razzamatazm:
## What a user would notice

On Settings > File Naming, the variable chips (`{show}`, `{season}`, etc.) below each template input were display-only. Hovering showed a tooltip, but clicking did nothing. The cursor style was `help`, which on desktop suggests a tooltip but implies you cannot interact further. On touch devices there is no hover at all, so the chips appeared purely decorative.

Now clicking any variable chip inserts it into the template field at the current cursor position. Focus and cursor are restored after insertion, so you can keep editing without extra clicks. A "Variables:" label was added before the chips so their purpose is clear at a glance.

## What changed

`frontend/src/pages/Settings.jsx` only. No backend changes.

- `TemplateSection`: added `useRef` ref on the TextInput, `insertVariable(varName)` function that reads `selectionStart`/`selectionEnd` and splices the variable into the template value, restores focus and cursor via `requestAnimationFrame`.
- Variable chips: `cursor: 'help'` changed to `cursor: 'pointer'`, added `onClick` handler.
- Added `Variables:` label (dimmed text) before the chip list.

## How it was tested

Playwright: opened File Naming, clicked `{date}` chip on the TV Shows template, confirmed the value changed from `{show} - S{season:02d}E{episode:02d} - {title}` to `{show} - S{season:02d}E{episode:02d} - {title}{date}`. PASS.

Before/after screenshots are in the Pixel iteration 39 thread in #design.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback, useContext } from 'react'
+import { useState, useEffect, useCallback, useContext, useRef } from 'react'
 import { UNSAFE_NavigationContext, useNavigate, useSearchParams, Link } from 'react-router-dom'
 import {
   Title,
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Design PR, single file Settings.jsx.

Correctness: clicking a variable chip splices `{name}` at the caret using selectionStart/selectionEnd, falls back to append when no ref, and restores caret after the controlled re-render via requestAnimationFrame + setSelectionRange. Mantine 7 TextInput forwards ref to the underlying input, so selection APIs are valid. Caret math uses start + insertion.length, correct regardless of the async template update.

No frontend test required (no frontend test suite exists yet). Minimal diff, 2-space React style, camelCase helper, no em dashes.

Minor, non-blocking: the Code chip has onClick but is not keyboard focusable, so it is mouse-only. Fine for now; could become a button later for accessibility.

PR #304, opened 2026-06-09
fix: handle bare-list response in get_epg and get_short_epg

razzamatazm:
## What a user would notice

If your IPTV provider returns EPG data in a slightly different format (a plain list instead of the standard `{"epg_listings": [...]}` wrapper), the catchup guide for affected channels would be silently empty. On the Browse EPG page it would show no programs to download. No error message, just nothing.

## What changed

`get_epg()` and `get_short_epg()` in `xtream_client.py` now handle both response shapes:
- `{"epg_listings": [...]}` (standard): extracts the list as before
- `[...]` (bare list): returns it directly

This is the same fix applied to VOD/series methods in a prior patch. `get_epg()` and `get_short_epg()` were missed.

## Before

```
data = []  # provider returned a bare list
data.get("epg_listings")  # AttributeError: 'list' object has no attribute 'get'
```

During EPG backfill: `AttributeError` caught per-channel, logged as warning, channel gets no EPG rows, catchup page empty with no explanation.

On Browse EPG live-EPG fallback path: `AttributeError` propagates as HTTP 500.

## After

```python
if isinstance(data, list):
    return data
return data.get("epg_listings") or []
```

Bare-list response returns the list directly. No error, channel EPG populated normally.

## How it was tested

8 new unit tests in `test_xtream_client_epg_list_response.py` cover both methods across four cases each: bare empty list, bare non-empty list, standard dict response, and null `epg_listings`. Bare-list tests fail before the fix (AttributeError), pass after. Full suite: 715 tests, OK.

### DIFF
--- a/backend/services/xtream_client.py
+++ b/backend/services/xtream_client.py
@@ -76,6 +76,8 @@ class XtreamClient:
             if response.status != 200:
                 raise Exception(f"Failed to get EPG: HTTP {response.status}")
             data = await response.json(content_type=None)
+            if isinstance(data, list):
+                return data
             return data.get("epg_listings") or []
 
     async def get_short_epg(self, stream_id: str, limit: int = 10) -> list:
@@ -86,6 +88,8 @@ class XtreamClient:
             if response.status != 200:
                 raise Exception(f"Failed to get short EPG: HTTP {response.status}")
             data = await response.json(content_type=None)
+            if isinstance(data, list):
+                return data
             return data.get("epg_listings") or []
 
     async def get_xmltv(self) -> bytes:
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

Correctness: get_epg / get_short_epg now return data directly when the provider sends a bare JSON list instead of {"epg_listings": [...]}. Verified against real source: the mocked session lets the real isinstance branch run, so the test exercises the actual fix, not a stub. Matches the existing get_live_streams non-dict guard, so behavior is consistent across the client.

Tests: four cases each for both methods (bare empty list, bare list with entries, dict extract, null epg_listings -> []). Covers the regression and the prior dict paths.

Minor, non-blocking and out of scope: a bare top-level `null` body would still raise on `data.get(...)`, same as before this PR. Not a regression. Leave it.

PR #305, opened 2026-06-09
docs: changelog entries for PR #294 and #301

razzamatazm:
## What changed

Adds plain-English changelog entries for two merged PRs that shipped without documentation:

- **PR #294** (fix): Channels no longer silently vanish from the guide when the provider uses invisible Unicode characters or different text encodings for the same channel name. The fix strips invisible characters and normalizes encoding before comparing channel names.
- **PR #301** (fix): A recording is now kept safe in the downloads folder when moving it to the completed folder fails because the destination drive is full. Previously the source recording was deleted and no explanation was given. The user now sees a clear message: "Not enough space in the completed recordings folder. Your recording is safe in the downloads folder."

## How tested

Verified no em dashes. Entries match the plain-English style used throughout the changelog and accurately describe the merged fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-09
 
+### Fixed: Your recording is kept safe when your completed folder runs out of space during the file move
+
+**What you would notice:** On Unraid or NAS setups where your downloads folder and your completed folder are on separate drives or shares, a full completed-folder drive could cause Mustarrd to delete the original recording after it failed to copy it across. You would end up with no file and no explanation. After this fix, if the completed folder is full, Mustarrd leaves your recording in the downloads folder untouched and marks the download Failed with the message "Not enough space in the completed recordings folder. Your recording is safe in the downloads folder."
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Docs-only changelog entries for #294 and #301, both verified against the merged code on origin/main.

Verification:
- #294: `_normalize_name` on origin/main now applies `unicodedata.normalize("NFC", ...)` and strips Cf-category code points (zero-width space U+200B, BOM U+FEFF) before the lower/split pass. The entry's description (strip invisible chars, normalize encoding before matching) is accurate. Note: #294 merged as a combined test+fix squash despite its "test:" title, so crediting it as the fix is correct.
- #301: `_execute_download` catches OSError from `_move_to_completed`, marks FAILED without deleting the source, and sets the ENOSPC message. Source-preservation behavior matches the entry.

Minor (non-blocking): the quoted #301 message is a prefix of the real string, which ends with a third sentence: "Free up space on that drive." Consider quoting it in full or trimming the quotes so the changelog does not imply the user sees a shorter message. Not a merge blocker.

Single commit, scope clean, no em or en dashes, plain-English style matches the rest of CHANGELOG.md.

PR #306, opened 2026-06-09
design: add confirm password field to Security settings

razzamatazm:
## What a user would notice

Settings > Security had two password fields: Current Admin Password and New Admin Password. There was no confirmation field. If a user mistyped their new password on a phone or while distracted, they would set an unknown password and lock themselves out of Mustarrd with no warning.

Now there is a third field: Confirm New Admin Password. The field shows a red "Passwords do not match" error inline while the user is typing, so they know before hitting Change Password. The submit handler also validates the match and shows a notification if they somehow reach it with mismatched values.

The description was also updated from "Change the admin password used to access Settings" to "Change your admin account password" since the password controls the whole app, not just Settings.

## What changed

\`frontend/src/pages/Settings.jsx\` only. No backend changes.

- Added \`confirmPassword\` state (cleared on success alongside the other fields).
- Added inline \`error\` prop on the new PasswordInput: shows "Passwords do not match" when confirm is non-empty and differs from new password.
- Added mismatch check in \`handleChangePassword\` before calling the API.
- Corrected the Security section description for admin users.

## How it was tested

Playwright: opened Security, typed different values in New and Confirm fields, confirmed the red inline error appeared. Typed matching values, confirmed no error. Before/after screenshots in #design.

## Before / After

Before: two fields, no typo protection, description says "...used to access Settings".
After: three fields, inline mismatch error, description says "Change your admin account password".

Screenshots in the Pixel iteration 40 thread in #design.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -141,6 +141,7 @@ export default function Settings() {
   const [searchParams, setSearchParams] = useSearchParams()
   const [currentPassword, setCurrentPassword] = useState('')
   const [newPassword, setNewPassword] = useState('')
+  const [confirmPassword, setConfirmPassword] = useState('')
   const [newUserName, setNewUserName] = useState('')
   const [newUserDisplayName, setNewUserDisplayName] = useState('')
   const [newUserPassword, setNewUserPassword] = useState('')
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Adds a Confirm New Password field that prevents lockout from a mistyped new password. confirmPassword state is cleared on success alongside the other fields, the inline error only shows once confirm is non-empty, and handleChangePassword re-checks the match before calling the API so the guard does not rely on UI state alone. Diff is minimal and contained to Settings.jsx, UI only with no backend or logic bundled, no em dashes. Frontend has no test harness in this repo, so the manual Playwright verification in the PR is the expected evidence for a design PR. No concerns.

PR #307, opened 2026-06-09
fix: PlexServer.enabled=False now blocks Plex login

razzamatazm:
## What a user would notice

An admin who went to Settings and disabled the Plex integration expected that Plex users could no longer log in. They couldn't. The \`enabled=False\` setting was saved to the database but had no effect on any of the three Plex login flows.

## What changed

Added a check for `plex_server.enabled` in three places in `backend/api/auth.py`:

- `plex_login_start`: returns 403 when enabled=False (Plex users can't start the PIN flow)
- `plex_login_complete`: returns 403 when enabled=False (PIN completion also blocked)
- `auth_status`: reports `plex_login_available: false` when enabled=False (UI correctly hides the Plex login button)

Three lines added, touching only `api/auth.py`.

## How it was tested

Six regression tests in `backend/tests/test_plex_enabled_guard.py`:

- The three "disabled" tests fail before this fix and pass after.
- The three "enabled" tests confirm the normal login path is unaffected.

Full suite: 715 tests, all pass.

**Before:** Admin sets enabled=False. Plex user calls `POST /auth/plex/login/start`. Receives a PIN. Completes it. Gets a session. `GET /auth/status` returns `plex_login_available: true`.

**After:** Admin sets enabled=False. Same Plex user calls `POST /auth/plex/login/start`. Receives HTTP 403. `GET /auth/status` returns `plex_login_available: false`.

### DIFF
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -158,6 +158,7 @@ async def auth_status(
     plex_server = plex_result.scalar_one_or_none()
     plex_login_available = bool(
         plex_server
+        and plex_server.enabled
         and plex_server.auto_allow_all_server_users
         and (plex_server.connection_uri or plex_server.base_url)
         and (plex_server.access_token_encrypted or plex_server.token_encrypted)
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve. The `enabled=False` guard is added to all three Plex login flows (plex_login_start, plex_login_complete, auth_status) and admin_plex.py is admin-config not a user-login bypass, so no path is missed. Model column defaults to True, so existing Plex servers stay enabled and no operator sees a behavior change. Ran tests/test_plex_enabled_guard.py locally: 6/6 pass, and the disabled-vs-enabled pairs are genuine regression coverage. Minimal diff, scope contained to api/auth.py plus the new test. No concerns.

PR #308, opened 2026-06-09
docs: changelog entries for PR #302, #304, and #306

razzamatazm:
## Summary

- PR #302: Variable chips on Settings > File Naming now click to insert. Added changelog entry under Improved.
- PR #304: Program guide now handles bare-list EPG responses from providers. Added changelog entry under Fixed.
- PR #306: Security settings now has a confirm password field. Added changelog entry under Improved.

## What a user would notice

Three new entries appear at the top of the changelog for 2026-06-09. All are written for a non-technical Unraid user with no em dashes.

## How it was tested

Diffed all three merged PRs against the existing CHANGELOG to confirm no duplicate entries. Verified diff before committing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-09
 
+### Improved: Variable chips on the File Naming settings page now click to insert
+
+**What you would notice:** On Settings > File Naming, the small chips showing available variables (like `{show}`, `{season}`, `{date}`) used to be display-only. You could hover over them to read a tooltip, but clicking did nothing. Now clicking any chip inserts it directly into the template field at your cursor position, so you can build a template without typing the variable names by hand. A "Variables:" label has also been added so the chips are easier to find at a glance.
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Changelog entries accurately describe the two merged PRs:

- The File Naming variable-chips entry matches #302: chips now have onClick that inserts {var} at the cursor (selectionStart/End), focus and caret restored via requestAnimationFrame, and a "Variables:" label added. Frontend only, as stated.
- The bare-list EPG entry matches #304: both get_epg and get_short_epg now return the response directly when it is a list instead of calling .get on it. Backend only, as stated.

Entries are not already present on main, follow the existing changelog format, and contain no em dashes. No code change, so no regression test required.

razzamatazm-cleo (approved):
Re-approved at head 7f80b54 (my earlier approval was at 0267129, before the #306 entry was added). All three entries verified: confirm-password field matches #306, variable-chip click-to-insert-at-cursor matches #302, bare-list EPG handling matches #304. None duplicated on main. Docs only, correct format, no em dashes. No concerns.

PR #309, opened 2026-06-09
design: show guidance when Plex account is not connected

razzamatazm:
## What a user would notice

Before this change, Settings > Plex Integration showed three empty dropdowns (Plex Server, Connection URI, Libraries to Refresh), a Disconnect button, and two switches even when no Plex account had been linked. A non-technical user would see what appeared to be a broken form with no indication of what to do first.

After this change, when no Plex account is connected the form is hidden and replaced with a single clear message: "No Plex account connected. Click Connect Plex Account to sign in with Plex. Your servers and libraries will appear here once connected." The Disconnect and Refresh Servers buttons are also hidden until an account is linked. The full form reappears once connected.

## What changed

One file, frontend only (`frontend/src/pages/Settings.jsx`). The `renderPlex` function now conditionally renders based on `plexConfig?.token_configured`. No backend changes.

## How it was tested

Playwright screenshots taken at 1280x800 (desktop) and 390x844 (mobile) with the app running locally. Before/after screenshots posted to #design. Both widths confirmed clean.

**Before:** Empty dropdowns, Disconnect/Refresh buttons visible, no explanation.
**After:** Single "Connect Plex Account" button and a clear guidance alert.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -1257,7 +1257,7 @@ export default function Settings() {
   const renderPlex = () => (
     <Stack gap="md">
       <Text fw={600}>Plex Integration</Text>
-      <Group justify="space-between" align="center">
+      <Group align="center">
         <Button
           variant="light"
           onClick={() => {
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. The renderPlex form is now correctly gated on plexConfig?.token_configured: the Connect button always shows, Refresh/Disconnect and the full server/library form only render once an account is linked, and a clear blue guidance Alert replaces the empty form otherwise. Verified token_configured is the same field the existing effect guards use (lines 533/540), and Alert and IconAlertCircle are both already imported. Frontend-only minimal diff with before/after Playwright evidence in the body and no em dashes.

Non-blocking note: while the plexConfig query is still loading, plexConfig is undefined so !plexConfig?.token_configured is true, meaning a connected user could see the "No Plex account connected" alert flash for a moment before data arrives. This matches the existing guard pattern in the file so it is no worse than current behavior; optional to gate on an isLoading check if you want to suppress the flash. No change required to merge.

PR #310, opened 2026-06-09
fix: CDATA with literal <programme text triggers spurious force-delete

razzamatazm:
## What a user would notice

If your IPTV provider sends XMLTV data where a program description contains the text `<programme` inside a CDATA block (for example, a description referencing an XML format), a forced EPG refresh would silently wipe your entire guide and then insert nothing. The catchup page would appear empty until the next automatic refresh.

## What changed

The code that counts `<programme` elements before deciding whether to clear the guide was using a raw byte scan. That scan could not distinguish between real XML elements and the same text buried inside a CDATA section. CDATA content is opaque character data to the XML parser, so `ET.iterparse` correctly ignores it and yields zero programme events, but the byte count still returned 1, which triggered the delete.

The fix strips CDATA sections from the byte buffer before counting, so the guard only fires when real parseable programme elements exist.

```
Before fix:
  xmltv_bytes.count(b"<programme")  # counts CDATA text too -> 1
  -> force-delete fires, guide wiped, nothing inserted

After fix:
  re.sub(CDATA pattern, b"", xmltv_bytes).count(b"<programme")  # 0
  -> force-delete skipped, existing rows preserved
```

## How it was tested

Added `test_force_refresh_cdata_only_programme_text_does_not_delete_rows` to `test_epg_force_wipe_empty_xmltv.py`. The test sends an XMLTV document containing `<programme` only inside a CDATA section, calls `_refresh_account(force=True)`, and asserts no DELETE statement was issued.

The new test fails against the old code (DELETE fires) and passes with the fix. Full suite: 718 tests, OK.

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -299,7 +299,8 @@ class EPGIngestManager:
 
             raw_xmltv = await client.get_xmltv()
             xmltv_bytes = self._maybe_decompress(raw_xmltv) if raw_xmltv else b""
-            total_programs = xmltv_bytes.count(b"<programme") if xmltv_bytes else 0
+            _xmltv_no_cdata = re.sub(rb"<!\[CDATA\[.*?\]\]>", b"", xmltv_bytes, flags=re.DOTALL) if xmltv_bytes else b""
+            total_programs = _xmltv_no_cdata.count(b"<programme")
             self._status["total_programs"] = total_programs if total_programs > 0 else None
             if xmltv_bytes:
                 await self._log(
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approved. CDATA strip before the `<programme` byte count is correct: non-greedy DOTALL pattern handles multiple CDATA blocks and a real-programme-plus-CDATA mix (real elements survive the strip, count stays accurate). `re` is imported (line 6), so no NameError. The count now matches ET.iterparse semantics, which already treat CDATA as opaque character data. Regression test tracks real DELETE statements through a patched session (it does not mock the code under test), fails before and passes after, 718 pass. Minimal diff, plain-English body with before/after, no em dashes. No concerns.

PR #311, opened 2026-06-09
fix: out-of-range start_timestamp returns HTTP 400 instead of 500

razzamatazm:
## What a user would notice

Sending a POST request to `/api/downloads` or `/api/schedules` with a timestamp value that is impossibly large (for example, from a corrupted IPTV provider EPG or a crafted API request) caused a server crash and returned HTTP 500. The server logged an unhandled `OverflowError`. It should have returned HTTP 400.

Normal frontend use is not affected: EPG timestamps from providers are always valid. This matters for API integrations and IPTV providers with corrupted far-future timestamps in their guide data.

## What changed

Two call sites used `datetime.fromtimestamp()` without catching `OverflowError` or `OSError`, which Python raises when a timestamp exceeds the platform's `time_t` range:

- `backend/services/download_builder.py`: `_parse_program_times()`, used by `POST /api/downloads`
- `backend/api/schedules.py`: `_parse_program()`, used by `POST /api/schedules`

Both functions already raised `ValueError` for other bad inputs, and their callers already converted `ValueError` to HTTP 400. The fix wraps the `fromtimestamp` calls in a `try/except (OverflowError, OSError)` and re-raises as `ValueError`, so the same HTTP 400 path is used.

`GET /api/downloads/failed-count` already used this exact pattern at line 206.

## How it was tested

Six regression tests in `backend/tests/test_timestamp_overflow.py`:

- `TestTimestampOverflowDownloadBuilder`: overflow start, overflow stop, normal timestamp guard (3 tests)
- `TestTimestampOverflowSchedules`: same three cases for the schedules path

The overflow tests fail before this fix (`OverflowError` raised, not `ValueError`) and pass after. The normal-timestamp guards confirm no regression.

**Before:**
```
POST /api/downloads {"program": {"start_timestamp": 99999999999999999999, ...}}
=> HTTP 500  OverflowError: Python int too large to convert to C long
```

**After:**
```
POST /api/downloads {"program": {"start_timestamp": 99999999999999999999, ...}}
=> HTTP 400  Invalid program payload
```

Full suite: 724 tests, all pass.

Fixes: https://discord.com/channels/1512584990910763059/1513867873800552590

### DIFF
--- a/backend/api/schedules.py
+++ b/backend/api/schedules.py
@@ -39,8 +39,11 @@ def _parse_program(program: dict) -> tuple[datetime, datetime, int, int, int, st
     stop_timestamp = _coerce_ts(program.get("stop_timestamp"))
 
     if start_timestamp and stop_timestamp:
-        start_time = datetime.fromtimestamp(int(start_timestamp), tz=timezone.utc)
-        end_time = datetime.fromtimestamp(int(stop_timestamp), tz=timezone.utc)
+        try:
+            start_time = datetime.fromtimestamp(int(start_timestamp), tz=timezone.utc)
+            end_time = datetime.fromtimestamp(int(stop_timestamp), tz=timezone.utc)
+        except (OverflowError, OSError) as exc:
+            raise ValueError(f"Program timestamp out of range: {exc}") from exc
     elif isinstance(program.get("start_time"), str) and isinstance(program.get("end_time"), str):
         start_time = datetime.fromisoformat(program["start_time"])
         end_time = datetime.fromisoformat(program["end_time"])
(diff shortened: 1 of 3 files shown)


razzamatazm-cleo (approved):
Approve. Verified:

- Both `datetime.fromtimestamp` call sites (download_builder._parse_program_times, schedules._parse_program) wrapped in try/except (OverflowError, OSError) and re-raised as ValueError. Callers already convert ValueError to HTTP 400 (downloads.py:366, schedules.py:153), so the 500 becomes a clean 400. Matches the existing pattern at downloads.py:206.
- Regression tests are genuine: OverflowError is not a ValueError subclass, so assertRaises(ValueError) fails before the fix and passes after. The normal-timestamp guards confirm no behavior change (duration 60, start_ts preserved). Tuple unpacking matches both function signatures.
- Minimal diff, two functions plus one test file, no unrelated changes, no em dashes.

Note: could not run the suite locally (deps not installed in my checkout), but verified the call paths and test logic by reading. No remaining concerns.

PR #314, opened 2026-06-09
fix: rate-limit plex_login_start to prevent Plex.tv API exhaustion

razzamatazm:
## What a user would notice

If Mustarrd is internet-exposed (NPM, Cloudflare Tunnel), an attacker could rapidly loop \`POST /api/auth/plex/login/start\` with no limit. Each call creates a real PIN on Plex.tv using the shared client identifier. If enough PINs are created, Plex.tv throttles the identifier and legitimate Plex users get \"Plex PIN create failed\" errors until the window expires.

## What changed

Added \`RATE_LIMITS["plex_start"] = 10\` and wired \`_enforce_rate_limit("plex_start", request)\` into \`plex_login_start\`, exactly matching the pattern \`plex_login_complete\` and the password login endpoints already use. The 11th request from the same client within the rate window returns HTTP 429. Requests below the limit are unaffected.

## Before

```
for i in $(seq 1 200); do curl -s -X POST http://host:4177/api/auth/plex/login/start; done
# All 200 return 200 OK and create real Plex.tv PINs
```

## After

```
# First 10 succeed; 11th returns:
{"detail":"Too many authentication attempts"}  HTTP 429
```

## Tests

4 new regression tests in \`backend/tests/test_plex_start_rate_limit.py\`:
- \`test_rate_limit_key_defined\`: asserts \`RATE_LIMITS["plex_start"]\` exists and is > 0
- \`test_eleventh_request_returns_429\`: pre-fills the log to the limit, verifies 429
- \`test_under_limit_passes_through\`: verifies calls below the limit still reach \`plex_service.create_pin\`
- \`test_different_clients_have_independent_buckets\`: exhausting client A's bucket doesn't block client B

Full suite: 728 tests, all pass.

## Steps to test

1. Configure Plex integration (or use a Mustarrd instance with \`auto_allow_all_server_users=True\`).
2. Before: 11 rapid POSTs all return 200. After: 11th returns 429.

### DIFF
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -34,6 +34,7 @@ RATE_LIMIT_WINDOW_SECONDS = 15 * 60
 RATE_LIMITS = {
     "setup": 6,
     "login": 20,
+    "plex_start": 10,
     "plex_complete": 30,
 }
 _attempt_log: dict[tuple[str, str], list[float]] = {}
@@ -369,7 +370,8 @@ async def login_download_user_legacy(
 
 
 @router.post("/plex/login/start")
-async def plex_login_start(session: AsyncSession = Depends(get_session)):
+async def plex_login_start(request: Request, session: AsyncSession = Depends(get_session)):
+    _enforce_rate_limit("plex_start", request)
     plex_result = await session.execute(select(PlexServer).limit(1))
     plex_server = plex_result.scalar_one_or_none()
     if not plex_server or not plex_server.auto_allow_all_server_users or not (plex_server.connection_uri or plex_server.base_url):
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve. Rate-limits plex_login_start before any DB work, closing an unauthenticated PIN-creation flood that could exhaust the shared Plex.tv client identifier.

- `Request` already imported (auth.py:6); `_enforce_rate_limit("plex_start", request)` runs before the PlexServer query.
- Key is (action, socket peer host) via `_client_key`, not X-Forwarded-For, so it is not trivially spoofable; per-client buckets confirmed by test.
- Limit 10 / 15 min window is consistent with the existing plex_complete (30) and login (20) limits.
- Regression test is valid: without the fix the 11th call falls through to the DB and raises a non-429 error, so test_eleventh_request_returns_429 fails before and passes after. All 4 tests pass locally.

No remaining concerns.

PR #313, opened 2026-06-09
docs: changelog entries for PR #309 and #310

razzamatazm:
## What a user would notice

Two new changelog entries added for recently merged changes.

## What changed

- **PR #309** (Plex settings guidance): Added entry explaining that the Plex Integration settings page now hides empty dropdowns and shows clear connection instructions when no account is linked.
- **PR #310** (CDATA EPG fix): Added entry explaining that a forced guide refresh no longer wipes the program guide when a provider sends program descriptions containing XML-like text.

## How it was tested

Entries checked against the merged PR descriptions. No em dashes used. Format matches existing CHANGELOG style.

PR #312, opened 2026-06-09
design: add guidance to cancelled scheduled recording cards

razzamatazm:
## What a user would notice

Before this change, cancelled scheduled recordings in Scheduled > History showed only the show title, channel, status badge, and air time. There was no explanation of what happened or any way to act on it. A user who did not deliberately cancel the recording had no idea why it was cancelled or what to do next.

After this change, cancelled cards that have no specific system message now show a short note: "Cancelled before downloading. If this program is still in your provider's catchup window, you can find it in Browse EPG." The words "Browse EPG" are a clickable link that takes the user directly there.

## What changed

Added a fallback text block to `ScheduleCard` in `Scheduled.jsx`. It renders only when `status === 'cancelled'` and no `status_message` or `download_id` is present (i.e., the recording was cancelled before a download was created). FAILED cards and cards with a specific system message are unaffected.

One file changed, no backend changes, no new dependencies.

## Before

Cancelled card shows title, channel, CANCELLED badge, and air time. Nothing else. No explanation, no next step.

## After

Cancelled card shows the same info plus: "Cancelled before downloading. If this program is still in your provider's catchup window, you can find it in Browse EPG."

## How it was tested

App started locally. Playwright screenshots taken before and after at desktop (1440x900) and mobile (390x844). Verified FAILED cards and COMPLETED cards are unchanged. Verified the Browse EPG link navigates correctly.

PR #315, opened 2026-06-09
docs: changelog entries for PR #311, #312, and #317

razzamatazm:
## What changed

Adds plain-English changelog entries for three merged PRs:

- **PR #311** (fix: out-of-range start_timestamp returns HTTP 400 instead of 500): Corrupted or far-future timestamps from a provider no longer crash the server.
- **PR #312** (design: add guidance to cancelled scheduled recording cards): Cancelled cards in Scheduled > History now show a note and a Browse EPG link when no other explanation is present.
- **PR #317** (fix: dedup schedule when program has no epg_id and no program_id): Scheduling the same program twice for a provider without EPG IDs now returns "This program is already scheduled" instead of creating a duplicate recording.

This PR was rebased onto main to resolve a conflict caused by PR #313 landing while it was open.

## What a user would notice

No visible change during normal use. These entries appear in the changelog for users reading release notes.

## How it was tested

Changelog text only, no code changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-09
 
+### Fixed: Scheduling the same program twice no longer creates duplicate recordings for providers without guide data
+
+**What you would notice:** Some IPTV providers do not include EPG IDs with their program data. Previously, if you accidentally scheduled the same program twice using such a provider, Mustarrd would create two scheduled recordings and attempt to download both at the same time, potentially corrupting the output file. Now, scheduling the same program twice returns "This program is already scheduled," matching the behavior you would see with a provider that includes full guide data.
+
+**What changed:** The schedule creation check was extended to cover programs that have neither an EPG ID nor a program ID. It now deduplicates on account, channel, and the program's start and stop times. Backend only, no frontend changes.
+
+---
+
+### Improved: Cancelled scheduled recordings now explain what happened and link to Browse EPG
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve.

Verified the changelog entry against the merged PR #311 diff:
- Two server-side functions changed: _parse_program in backend/api/schedules.py and _parse_program_times in backend/services/download_builder.py. Matches 'two server-side functions that convert provider timestamps to dates'.
- Both now catch OverflowError/OSError and raise ValueError, which callers turn into HTTP 400 instead of the prior 500. Matches the entry and PR #311's title.
- Affects both the download path (download_builder) and scheduled recording path (schedules). Matches 'queue a download or create a scheduled recording'.
- Backend only, no settings touched. Accurate.

Date heading 2026-06-09 matches the PR #311 merge date. No em dashes. Docs-only, no test needed. No remaining concerns.

razzamatazm-cleo (approved):
Approve at rebased head 6fc694e. Verified all three changelog entries against the merged PRs:
- #311 (timestamp 400 not 500): matches, backend-only, two timestamp-parse functions hardened.
- #312 (cancelled card guidance): matches the merged design PR title and scope, frontend-only.
- #317 (dedup with no epg_id/program_id): accurate. Merge diff adds an else-branch dedup on (account_id, channel_id, start_timestamp, stop_timestamp, active status), exactly as the entry describes; ships with three regression tests.
Date heading 2026-06-09 matches merge dates. The '---' lines are markdown horizontal rules matching existing changelog style, not prose em dashes. Docs-only, no test needed. No remaining concerns.

PR #316, opened 2026-06-09
design: add description to Download-Only Users settings panel

razzamatazm:
## What changed

Settings > Users had no description subtitle and an undersized title, unlike every other settings panel in the app.

Every other section (Recording, Post-Processing, File Naming, Guide, Appearance, Security) shows a dimmed subtitle explaining what the section does. Users had only a bold title with no context.

## What a user would notice

Before: clicking Settings > Users showed "Download-Only Users" with no explanation of what these accounts can or cannot do.

After: a one-line description now reads: "Can browse channels, schedule recordings, and download programs. Cannot access Settings."

The description accurately reflects the actual permission model: download-only users can schedule recordings (POST /schedules allows them), but cannot access the Settings page.

## How it was tested

Playwright screenshots at 1280px desktop and 390px mobile confirm the description renders correctly at both widths.

## Files changed

- `frontend/src/pages/Settings.jsx`: added `size="lg"` to the title and a dimmed description line under it. One file, no backend changes.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -1182,8 +1182,11 @@ export default function Settings() {
   )
 
   const renderUsers = () => (
-    <Stack gap="md">
-      <Text fw={600}>Download-Only Users</Text>
+    <Stack gap="lg">
+      <Stack gap={2}>
+        <Text fw={600} size="lg">Download-Only Users</Text>
+        <Text size="sm" c="dimmed">Can browse channels and download programs. Cannot access Settings or schedule recordings.</Text>
+      </Stack>
       <Alert color="blue" variant="light">
         <Group justify="space-between" align="center" wrap="nowrap">
           <Text size="sm">Plex Integration controls Plex-based sign-in and server/library sync.</Text>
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Request changes: the new description is factually inaccurate.

frontend/src/pages/Settings.jsx: the added line reads 'Can browse channels and download programs. Cannot access Settings or schedule recordings.' The 'cannot schedule recordings' half contradicts current backend behavior.

POST /schedules (backend/api/schedules.py:138) depends on require_admin_or_download_user, which in backend/auth.py:171 only requires an authenticated context with no is_admin check. create_schedule even has an explicit 'if not auth.is_admin: conditions.append(ScheduledRecording.requested_by_user_id == auth.user_id)' branch, which exists precisely so download users can create their own scheduled recordings. So a download-only user CAN schedule recordings today.

A non-technical operator reading this panel would be misled about what the role allows. Please either:
- correct the copy to match behavior (drop 'or schedule recordings', e.g. 'Can browse channels, download programs, and schedule recordings. Cannot access Settings.'), or
- if the intent is that download-only users should NOT be able to schedule, that is a separate backend change, and the copy must not claim a restriction that is not enforced.

The 'Cannot access Settings' half is fine (Settings endpoints use require_admin). Styling change itself is clean and has no em dashes.

razzamatazm-cleo (approved):
Approve. Prior concern resolved: the description copy now reads 'Can browse channels, schedule recordings, and download programs. Cannot access Settings.', which matches actual backend behavior (POST /schedules permits download-only users via require_admin_or_download_user; Settings endpoints require_admin). Clean one-file frontend styling change, no em dashes, no backend changes. No remaining concerns.

PR #317, opened 2026-06-09
fix: dedup schedule when program has no epg_id and no program_id

razzamatazm:
## What changed

`POST /schedules` checks for duplicate schedules before creating a new one, but only when the program payload includes `epg_id` or `id`. When both are absent (providers that strip EPG metadata), the duplicate check was silently skipped.

A second POST for the same program created a second `ScheduledRecording` row. When their `available_at` time arrived, the scheduler dispatched both, creating two `Download` rows aimed at the same output file. The two FFmpeg processes wrote to the same path simultaneously, corrupting the output.

## Fix

Added an `else` branch in `api/schedules.py` that deduplicates on `(account_id, channel_id, start_timestamp, stop_timestamp)` when both `epg_id` and `program_id` are absent. The second POST now returns HTTP 409 "This program is already scheduled," matching the behavior of the other two branches.

## Before

```
# POST /schedules twice, same program, no epg_id or id field
POST /schedules -> 201 Created
POST /schedules -> 201 Created  (should be 409)
# Two ScheduledRecording rows in DB; both dispatch; output file corrupted
```

## After

```
POST /schedules -> 201 Created
POST /schedules -> 409 This program is already scheduled
# One ScheduledRecording row; no duplicate dispatch
```

## Tests

Three new regression tests in `backend/tests/test_schedule_dedup_fallback.py`:
- First POST with no `epg_id`/`program_id` succeeds.
- Second POST with same timestamps returns 409.
- Different `channel_id` is not treated as a duplicate.

`test_duplicate_schedule_without_ids_returns_409` fails before the fix and passes after. Full suite: 727 tests, all pass.

### DIFF
--- a/backend/api/schedules.py
+++ b/backend/api/schedules.py
@@ -201,6 +201,19 @@ async def create_schedule(
         existing = await session.execute(select(ScheduledRecording).where(*conditions))
         if existing.scalar_one_or_none():
             raise HTTPException(status_code=409, detail="This program is already scheduled")
+    else:
+        conditions = [
+            ScheduledRecording.account_id == data.account_id,
+            ScheduledRecording.channel_id == data.channel_id,
+            ScheduledRecording.start_timestamp == start_ts,
+            ScheduledRecording.stop_timestamp == stop_ts,
+            ScheduledRecording.status.in_(active_statuses),
+        ]
+        if not auth.is_admin:
+            conditions.append(ScheduledRecording.requested_by_user_id == auth.user_id)
+        existing = await session.execute(select(ScheduledRecording).where(*conditions))
+        if existing.scalar_one_or_none():
+            raise HTTPException(status_code=409, detail="This program is already scheduled")
 
     custom_filename = _sanitize_filename(data.custom_filename)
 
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

Logic: the new else branch correctly mirrors the epg_id and program_id branches, deduping on (account_id, channel_id, start_timestamp, stop_timestamp, active status) with the same non-admin requested_by_user_id scoping. This closes the real gap where a provider stripping both epg_id and program_id let a second identical POST create a duplicate ScheduledRecording.

Edge cases: false-positive dedup is not a concern. A channel airs one program per time slot, so two distinct programs cannot share the exact same channel_id, start_timestamp, and stop_timestamp. start_ts/stop_ts and active_statuses are defined before all three branches.

Tests: the regression test drives the real create_schedule with a mocked session. test_duplicate_schedule_without_ids_returns_409 fails before this change (no else branch means no dedup, so no 409 is raised) and passes after. The success and different-channel cases confirm no over-blocking.

Conventions: minimal diff, matches file style, no em dashes, plain-English PR body. No remaining concerns.

PR #318, opened 2026-06-09
fix: get_auth_context validates User.role for download_only sessions

razzamatazm:
## What a user would notice

No visible change for normal users. This closes a consistency gap in how the app verifies who you are when you log in.

## What changed

When you log in, the app stores your role (admin or download-only) in a session cookie. When you make a request, it checks that cookie against the database to confirm you are still allowed in.

The admin path has always verified your role in the database: it only accepts a session if the database row also says admin. The download-only path was missing that same check. It confirmed your user ID but not your role, so the server was taking the cookie's word for it rather than checking the database.

The fix adds the matching role filter to the download-only branch. One line change in `backend/auth.py`:

**Before:**
```python
select(User).where(User.id == int(user_id))
```

**After:**
```python
select(User).where(User.id == int(user_id), User.role == "download_only")
```

## How it was tested

Three regression tests in `backend/tests/test_auth_context_role_check.py`:

1. A download-only DB user with a matching cookie authenticates correctly (unchanged behavior).
2. An admin DB user with a tampered download-only cookie is rejected. Before the fix this test would have failed because the query returned the admin row and granted access.
3. An unknown user ID is rejected.

Full suite: 730 tests, all pass.

**Before (without fix), test 2 would return:**
```
authenticated=True, role="download_only"
```

**After (with fix), test 2 returns:**
```
authenticated=False
```

### DIFF
--- a/backend/auth.py
+++ b/backend/auth.py
@@ -129,7 +129,7 @@ async def get_auth_context(
 
     if role == "download_only" and user_id:
         result = await session.execute(
-            select(User).where(User.id == int(user_id))
+            select(User).where(User.id == int(user_id), User.role == "download_only")
         )
         user = result.scalar_one_or_none()
         if user and user.status == "active":
--- a/backend/tests/test_auth_context_role_check.py
+++ b/backend/tests/test_auth_context_role_check.py
@@ -1,0 +1,88 @@
+"""
+Regression test: get_auth_context must verify User.role == "download_only" in DB.
+
+Before fix: the download_only branch queried only by User.id, never checking
+User.role. A session cookie with auth_role="download_only" for an admin user_id
+was granted download_only access without DB role confirmation.
+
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approved. The fix adds `User.role == "download_only"` to the lookup, making the download_only branch symmetric with the admin branch (which already filters on role). This closes the gap where a session claiming role=download_only authenticated against any user_id regardless of that user's actual DB role.

Verified locally: the new regression test fails on pre-fix auth.py (test_admin_user_with_download_only_cookie_rejected: `AssertionError: True is not false`) and passes after the fix. Tests exercise the real get_auth_context against an in-memory SQLite DB, no mocking of the query path, so they prove the real behavior. Minimal one-line diff, conventions met. No remaining concerns.

PR #319, opened 2026-06-09
design: add description to Accounts settings panel

razzamatazm:
## What changed

Settings > Accounts had no description subtitle, unlike every other settings panel in the app.

Every other section (Recording, Post-Processing, File Naming, Guide, Appearance, Security, and now Users) shows a dimmed subtitle explaining what the section does. Accounts showed only the bold title with no context.

## What a user would notice

Before: clicking Settings > Accounts showed just "Accounts" with action buttons and no explanation of what this section is for.

After: a one-line description reads "Connect your IPTV providers. Mustarrd reads your channel list and program guide from these accounts."

The button group alignment also changed from center to flex-start so the buttons align to the top of the title and description stack rather than vertically centering against it. flexShrink: 0 was added to the button group to prevent squishing on narrow desktop widths.

## How it was tested

Playwright screenshots at 1280px desktop and 390px mobile confirm the description renders correctly at both widths.

## Files changed

- frontend/src/components/settings/AccountsSection.jsx: wrapped title in a Stack with the new description text, adjusted Flex alignment. One file, no backend changes.

### DIFF
--- a/frontend/src/components/settings/AccountsSection.jsx
+++ b/frontend/src/components/settings/AccountsSection.jsx
@@ -515,12 +515,15 @@ export default function AccountsSection({ showTitle = true }) {
       <Flex
         direction={{ base: 'column', sm: 'row' }}
         justify={{ sm: 'space-between' }}
-        align={{ sm: 'center' }}
+        align={{ sm: 'flex-start' }}
         gap="xs"
       >
-        {showTitle ? <Title order={2}>Accounts</Title> : <Title order={4}>Accounts</Title>}
+        <Stack gap={2}>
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approved. UI-only change, one file. `Stack` and `Text` are both already imported from @mantine/core (lines 8 and 9), so no import gap. The dimmed subtitle brings Accounts in line with every other settings section, copy is accurate to the product (Xtream IPTV provider accounts feeding channel list + EPG), and the flex-start + flexShrink:0 adjustments are sensible for keeping the button group from squishing on narrow widths. No em dashes. Body documents before/after with Playwright screenshot evidence at 1280px and 390px. No concerns.

PR #321, opened 2026-06-09
fix: sanitize_filename strips bidi override and isolate controls

razzamatazm:
## What a user would notice

If a TV provider injected invisible Unicode directional characters into a program title, the downloaded file could appear to have a reversed or garbled name in your terminal, Finder, or file manager. For example, a file genuinely named `My Show S01E01.ts` could display as `st.10E10S wohS yM` in some terminals. This also broke Plex and Jellyfin library matching because the visible filename differed from the real one.

## What changed

`sanitize_filename` already stripped a set of invisible formatting characters (zero-width spaces, BOM, etc.). This extends that same strip to cover nine additional Unicode bidi (bidirectional text) control characters: the embedding controls (U+202A-202B), the directional overrides (U+202C-202E), and the isolate controls (U+2066-2069). All are silently removed before the filename reaches disk.

## How it was tested

Before the fix, inserting U+202E (RTL Override) into a title like `"Breaking‮Bad"` produced a filename with the control character still present.

After the fix:
```
$ python -m unittest tests.test_sanitize_bidi_chars -v
test_bidi_only_title_returns_fallback ... ok
test_each_bidi_char_stripped ... ok
test_mixed_bidi_chars_all_stripped ... ok
test_rtl_override_trojan_filename_cleaned ... ok
Ran 4 tests in 0.001s OK
```

Full suite: 734 tests, OK. No regressions.

### DIFF
--- a/backend/services/file_namer.py
+++ b/backend/services/file_namer.py
@@ -22,9 +22,10 @@ class FileNamer:
         # Remove stylized "Live" markers frequently injected by providers
         name = re.sub(r'[\[\(\-–—|:]*\s*ᴸᶦᵛᵉ\s*[\]\)]*\s*(?:-\s*)?', ' ', name)
 
-        # Remove invisible Unicode format/directional chars (zero-width, BOM, soft hyphen).
-        # Use empty replacement so they vanish without splitting surrounding words.
-        name = re.sub('[\u00ad\u200b\u200c\u200d\u200e\u200f\ufeff]', '', name)
+        # Remove invisible Unicode format/directional chars: zero-width, BOM, soft hyphen,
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approved. Extends existing invisible-char strip in FileNamer.sanitize_filename to the 9 bidi override/isolate controls (U+202A-202E, U+2066-2069). Verified: edited line is the staticmethod actually called by the test; regex codepoints correct; test covers each char, the RTL-override trojan filename, mixed input, and the bidi-only -> unknown-program fallback. Minimal diff, regression test fails-before/passes-after, 734 suite green. No concerns.

PR #324, opened 2026-06-09
fix: re-check schedule status after queue_download to prevent cancel race

razzamatazm:
## What a user would notice

Cancelling a scheduled recording while it was being dispatched had no effect. The recording would silently start downloading anyway, and the schedule would show as "Queued" even though the user had cancelled it. The only workaround was to delete the stuck schedule and cancel the download manually.

## What changed

When the scheduler picks up a scheduled recording to dispatch, it calls the provider API to build the download and then queues it. That takes a brief moment. If the user cancels the schedule during that window, the cancel writes `CANCELLED` to the database and commits. But the scheduler had already loaded the schedule into memory, and its final commit overwrote the cancelled status with `QUEUED`, launching the download.

The fix: after queuing the download, the scheduler now re-reads the schedule's status from the database (`session.refresh`). If the status is no longer `SCHEDULED` or `PAUSED_LOW_SPACE` (for example, because the user cancelled it), the scheduler cancels the download it just queued and leaves the schedule as-is.

**Before:** Cancel during dispatch had no effect. New download appeared on the Downloads page despite the user having cancelled the schedule.

**After:** The scheduler detects the cancellation, calls `cancel_download`, and stops the download before it can start. The schedule status is preserved as `CANCELLED`.

## How it was tested

Two regression tests added in `backend/tests/test_cancel_during_dispatch.py`:

1. `test_cancel_during_dispatch_aborts_download`: simulates a cancel mid-dispatch by having `session.refresh` return `CANCELLED`. Asserts that `cancel_download` is called with the queued download's ID and the schedule is not set to `QUEUED`. This test fails before the fix and passes after.

2. `test_no_cancel_proceeds_normally`: confirms the normal path (no cancel) still sets the schedule to `QUEUED` and does not call `cancel_download`.

Full suite: 748 tests, 744 pass. The 4 pre-existing failures (`test_plex_enabled_guard`, `test_plex_start_rate_limit` x2) are unrelated to this change and were already present on `main` before this branch was created.

### DIFF
--- a/backend/services/scheduled_manager.py
+++ b/backend/services/scheduled_manager.py
@@ -125,6 +125,17 @@ class ScheduledManager:
                     )
 
                     download = await download_manager.queue_download(download)
+
+                    # Re-read status: user may have cancelled while we were
+                    # awaiting build_download_from_program or queue_download.
+                    await session.refresh(schedule)
+                    if schedule.status not in (
+                        ScheduledStatus.SCHEDULED.value,
+                        ScheduledStatus.PAUSED_LOW_SPACE.value,
+                    ):
+                        await download_manager.cancel_download(download.id)
+                        continue
+
                     schedule.download_id = download.id
                     schedule.status = ScheduledStatus.QUEUED.value
                     schedule.status_message = None
--- a/backend/services/scheduled_manager.py
+++ b/backend/services/scheduled_manager.py
@@ -140,10 +140,12 @@ class ScheduledManager:
                     schedule.status = ScheduledStatus.QUEUED.value
                     schedule.status_message = None
                     schedule.updated_at = datetime.utcnow()
+                    await session.commit()
                 except Exception as exc:
                     schedule.status = ScheduledStatus.FAILED.value
                     schedule.status_message = str(exc)
                     schedule.updated_at = datetime.utcnow()
+                    await session.commit()
 
             await session.commit()
 
--- a/backend/tests/test_cancel_during_dispatch.py
+++ b/backend/tests/test_cancel_during_dispatch.py
@@ -168,5 +168,91 @@ class CancelDuringDispatchTests(unittest.IsolatedAsyncioTestCase):
         fake_dm.cancel_download.assert_not_called()
 
 
+    async def test_commit_fires_between_schedules(self):
+        """Per-schedule commit: session.commit() must fire after schedule A is set to
+        QUEUED and before schedule B's build starts.
+
+        Without this, a cancel of A during B's build sees download_id=NULL in the DB
+        (A not yet committed) and cannot cancel the download. The download then runs
+        after the final end-of-loop commit overwrites the CANCELLED status with QUEUED.
+        """
+        schedule_a = _make_ready_schedule()
+        schedule_a.id = 1
+        schedule_b = _make_ready_schedule()
+        schedule_b.id = 2
+
+        events = []
+        build_call = [0]
+
+        async def fake_build(*args, **kwargs):
+            build_call[0] += 1
+            dl = MagicMock()
+            dl.id = build_call[0] * 10
+            events.append(f"build_{build_call[0]}")
+            return dl
+
+        async def fake_queue(dl):
+            return dl
+
+        async def fake_commit():
+            events.append("commit")
+
+        scalars_result = MagicMock()
+        scalars_result.all.return_value = [schedule_a, schedule_b]
+        schedules_exec_result = MagicMock()
+        schedules_exec_result.scalars.return_value = scalars_result
+
+        settings_obj = AppSettings()
+        settings_obj.download_folder = "/tmp/downloads"
+        settings_obj.min_free_space_gb = 1
+        settings_exec_result = MagicMock()
+        settings_exec_result.scalar_one_or_none.return_value = settings_obj
+
+        call_count = [0]
+
+        async def execute(stmt):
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (changes requested):
Good catch on the race and the fix improves the common case, but the deferred commit leaves a residual hole in the same bug class, so requesting changes.

**Main issue: commit is outside the loop, so the re-check is not durable until all schedules are processed.**
session.commit() is at scheduled_manager.py:148, after the 'for schedule in ready' loop. The fix sets schedule.status = QUEUED at line 140 but does not persist it until line 148. So:
- Schedule A is re-checked (refresh, still SCHEDULED), set to QUEUED in-memory, download_id assigned in-memory.
- The loop moves on and awaits build_download_from_program / queue_download for schedule B (a long network/IO await, the same kind of window the original bug was about).
- During that await the user cancels A. cancel_schedule reads A from the DB (still SCHEDULED, download_id still NULL because we have not committed), sets it CANCELLED, commits. Because download_id is NULL it does NOT cancel the queued download.
- The loop finishes and session.commit() at :148 flushes A as QUEUED with download_id, overwriting the user's CANCELLED. The download runs anyway.

That is exactly the bug this PR is meant to fix, just shifted one or more iterations later. It is fully fixed only when 'ready' has a single schedule.

**Fix:** commit per schedule, immediately after setting QUEUED, inside the loop. There is no await between the refresh re-check (:131) and the status assignments (:140-142), so a per-schedule commit makes check-and-persist atomic under asyncio's single thread and closes the window completely: a cancel then either lands before the refresh (caught here) or after the commit (sees QUEUED + a real download_id and cancels the download correctly). Keep the FAILED-path commit too. Alternatively, guard with a conditional UPDATE ... WHERE status IN (SCHEDULED, PAUSED_LOW_SPACE) and check rowcount.

**Verification I ran:**
- Confirmed the test fails on current main (no fix) and passes with the PR's source. Valid regression for the targeted window.
- The test mocks session.refresh() to inject CANCELLED, which proves the post-refresh branch logic but NOT that refresh actually observes a cancel committed by another session. I checked that separately against the current sqlite+aiosqlite config (no WAL, default isolation): a second session's committed UPDATE IS visible to the first session's refresh, including when the first session already has pending DML from an earlier iteration. So the mechanism works today. This relies on pysqlite running SELECTs in autocommit; if the DB ever moves to WAL or an explicit isolation level it could change. A test using two real sessions against a temp DB would lock this in. Nice-to-have, not blocking.

**Minor:** if a schedule row is deleted (rather than set CANCELLED) between the initial SELECT and the refresh, session.refresh() will raise; the current cancel_schedule path only deletes non-active schedules so this is not reachable today, but worth a defensive note.

Please move the commit into the loop (the one-line relocation) and I will re-review.

razzamatazm:
Addressed Cleo's review. Changes in this push:

**What changed:** Moved `session.commit()` inside the loop, immediately after setting each schedule to QUEUED (and after setting FAILED in the exception path). The end-of-loop commit at line 148 is kept to persist PAUSED_LOW_SPACE updates.

**Why it matters:** Before, if two schedules were ready and schedule A was processed first (set to QUEUED in-memory), a user cancel arriving while schedule B was being dispatched would read A from the DB and see `download_id=NULL` (A not committed yet). The cancel would do nothing. The final commit would then overwrite CANCELLED with QUEUED and the download would run anyway. Now A is committed before B starts, so any cancel always sees a real `download_id` and can stop the download.

**New test:** `test_commit_fires_between_schedules` runs two schedules through the loop, intercepts the order of events (build + commit calls), and asserts a commit fires after A is set to QUEUED and before B's `build_download_from_program` is called. This test would fail on the old code.

All 3 tests pass.

razzamatazm-cleo (approved):
Approve. My earlier change request is addressed and the fix is correct for its scope.

**What I asked for, now done.** `session.commit()` is moved inside the loop, immediately after a schedule is set to QUEUED (scheduled_manager.py:148), and the FAILED exception path now commits too. There is no await between the `session.refresh(schedule)` re-check (:130) and the status assignments + commit, so check-and-persist is atomic under asyncio's single thread. A cancel arriving during the build/queue awaits is now caught one of two ways: it lands before the refresh (refresh sees the non-active status, cancel_download is called, continue), or after the commit (the cancel reads a real download_id and stops the queued download). The QUEUED-overwrite hole from my prior review is closed.

**Verification I ran:** all 3 tests pass on this branch; on main, `test_cancel_during_dispatch_aborts_download` and `test_commit_fires_between_schedules` both fail (`Got event order: ['build_1', 'build_2', 'commit']`), confirming a valid regression for the targeted window. The mock-refresh caveat from my first review still applies (it proves the post-refresh branch, not that aiosqlite's refresh observes a cross-session commit), but I previously confirmed that mechanism holds under the current no-WAL sqlite config, so this is not blocking.

**One pre-existing residual, out of scope for this PR, flagging so it is not lost:** the PAUSED_LOW_SPACE branch (:89) sets status in-memory and `continue`s without an in-loop commit, so it is persisted only by the end-of-loop commit at :153. A cancel that arrives while a *later* schedule is being dispatched can still be clobbered back to PAUSED_LOW_SPACE for a low-space schedule. This is milder than the bug fixed here (no download_id is assigned, so nothing downloads in that cycle) and predates this change, but it is the same class. Worth a follow-up task rather than expanding this diff.

PR #325, opened 2026-06-09
test: failing regression for _postproc_tmp.ts orphaned on concat failure

razzamatazm:
## What the user sees

When commercial removal is configured with TS output format (`transcode_format=ts`) and the FFmpeg concat step fails mid-run, a partial `Show_postproc_tmp.ts` file is left in the download working folder. It accumulates over repeated failures and is never cleaned up automatically.

## What changed

Adds one failing regression test that proves the bug. No production code is changed.

## Root cause

`remove_commercials` in `post_processor.py` creates `{stem}_postproc_tmp.{ext}` to protect the original file while the concat runs (lines 1190-1192). When concat fails, the `finally` block (lines 1389-1395) cleans `_seg*.ts` segment files and `.concat.txt`, but omits `work_output_path`. `_cleanup_working_files` in `download_manager.py` also has no `*_postproc_tmp.*` pattern.

Trigger conditions: `transcode_format=ts` + `comskip_enabled=True` + commercial removal detects segments + FFmpeg concat fails.

Distinct from the two existing temp-file orphan TODOs. Those cover either the path where cleanup is never called, or `{stem}.mkv`/`{stem}.mp4` missed by patterns. Here the `finally` block runs and cleans segments, but misses the temp output file.

## How to test

```
cd backend
python3 -m unittest tests.test_postproc_tmp_orphan -v
# FAIL: AssertionError: True is not false (file still on disk)
```

The test passes once the `finally` block in `remove_commercials` unlinks `work_output_path` on failure when `same_path_output=True`.

### DIFF
--- a/backend/tests/test_postproc_tmp_orphan.py
+++ b/backend/tests/test_postproc_tmp_orphan.py
@@ -1,0 +1,112 @@
+"""
+Regression test: {stem}_postproc_tmp.ts orphaned on concat failure when
+commercial removal uses TS output format (same_path_output=True).
+
+Bug: remove_commercials creates {stem}_postproc_tmp.{ext} when input and output
+share the same extension (e.g. transcode_format=ts).  If FFmpeg concat fails,
+the finally block cleans _seg*.ts and .concat.txt but NOT work_output_path.
+_cleanup_working_files in download_manager also has no *_postproc_tmp.* pattern.
+Result: partial temp file accumulates in the working directory indefinitely.
+
+Distinct from the two existing temp-file orphan bugs:
+- "Comskip and FFmpeg temp files orphaned on failure/cancel": covers the path
+  where _cleanup_working_files is never called at all.
+- "Partial transcode output orphaned on warning path": covers {stem}.mkv/.mp4
+  missed by cleanup patterns.
+Here the finally block DOES run and DOES clean segments, but misses
+work_output_path.
+
+This test FAILS on the current codebase.
+"""
+import asyncio
+import importlib.util
+import os
+import shutil
+import tempfile
+import unittest
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch, PropertyMock
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve on technical merits.

I confirmed the bug is real and the test is a valid regression:
- remove_commercials() builds work_output_path = {stem}_postproc_tmp.{ext} when input and output share an extension (same_path_output=True), e.g. transcode_format=ts.
- On concat failure the raise at post_processor.py:1387 fires; the finally block (1389-1395) removes only the _seg*.ts segments and the .concat.txt, never work_output_path.
- download_manager._cleanup_working_files patterns (1124-1136) cover _seg*.ts, .concat.txt, .mkv, .mp4, etc., but have no *_postproc_tmp.* pattern, and the stem 'Show_postproc_tmp' does not match any of them.
- Ran the test against current main source: it reaches the orphan assertion and fails exactly as documented (the partial Show_postproc_tmp.ts is left on disk). The assertRaises path also fires correctly, so the test exercises the real failure branch, not a mocked one.

Test is well-isolated (tempdir, cleaned in tearDown), loads post_processor via importlib, and the fix hint in the assertion message is accurate (unlink work_output_path in the finally when same_path_output).

Note: this is an intentionally-failing regression with no accompanying fix, so merging it turns main's test suite red until the fix lands, same pattern as #282. That trade-off is Tyler's call (Herald has already asked him to approve/reject); I am approving the test itself, which is correct and targets a genuine bug. The follow-up fix should add work_output_path to the finally cleanup and ideally a *_postproc_tmp.* pattern to _cleanup_working_files as a backstop.

razzamatazm:
PR #327 (https://github.com/razzamatazm/mustarrd/pull/327) implements the fix and includes this test. #327 supersedes this PR. Suggest closing #325 once #327 is reviewed.

razzamatazm:
PR #327 implements the fix and includes this test, making this PR redundant. Closing per Tyler's guidance.

PR #326, opened 2026-06-09
design: persist Downloads tab selection in URL

razzamatazm:
## What a user would notice

Before this change, navigating to the Downloads page always opened on the Active tab, regardless of the URL. Clicking the History or Upcoming tab, then refreshing the page, reset back to Active. Sharing a link to the History tab or bookmarking it also did not work.

After this change, each tab updates the URL to `/downloads?tab=active`, `/downloads?tab=upcoming`, or `/downloads?tab=history`. Refreshing preserves the tab. Deep links and browser Back work as expected.

## What changed

One file (`Downloads.jsx`): added `useSearchParams` from react-router-dom (already used in Settings.jsx for section routing). The `Tabs` component now reads its active value from `?tab=` and writes back on change. Unknown or missing values fall back to `active`.

## Before

`/downloads?tab=history` landed on the Active tab, showing "No active downloads" instead of the history list. Refreshing on any non-active tab reset to Active.

## After

`/downloads?tab=upcoming` opens directly on Upcoming. `/downloads?tab=history` opens History. Clicking a tab updates the URL. Refresh keeps the tab. Screenshots in #design.

## How it was tested

Playwright drove a browser to `/downloads?tab=upcoming` and `/downloads?tab=history` directly and confirmed the correct tabs loaded. Also confirmed clicking History updates the URL to `?tab=history`.

No backend changes. Design-only.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -18,7 +18,7 @@ import {
   Select,
 } from '@mantine/core'
 import { notifications } from '@mantine/notifications'
-import { useNavigate, Link } from 'react-router-dom'
+import { useNavigate, Link, useSearchParams } from 'react-router-dom'
 import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query'
 import {
   IconDotsVertical,
@@ -402,6 +402,7 @@ function DownloadCard({
 
 export default function Downloads() {
   const navigate = useNavigate()
+  const [searchParams, setSearchParams] = useSearchParams()
   const queryClient = useQueryClient()
   const [localProgress, setLocalProgress] = useState({})
   const [localLogs, setLocalLogs] = useState({})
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Persists the Downloads tab in the URL via useSearchParams.

Correctness:
- Tab value is validated against ['active','upcoming','history'] with a fallback to 'active', so a hand-edited ?tab=garbage cannot break the Tabs component.
- onChange writes {tab} with replace:true. I checked Downloads.jsx: no other query params are read on this page (the navigate() calls target /browse), so replacing the full query string drops nothing. Deep-linking ?tab=history and back/forward both work.
- No URL is written on first load until the user clicks, so the clean /downloads URL still shows the active tab. Good.

Tests: frontend has no test framework yet, same as prior design PRs, so no regression test is practical here.

Scope: minimal, single-file, UI-routing only, no logic bundled in. No em dashes. No concerns.

PR #327, opened 2026-06-09
fix: clean up _postproc_tmp file on concat failure in remove_commercials

razzamatazm:
## What a user would notice

After running commercial removal with the TS output format, a file named `Show_postproc_tmp.ts` could be left sitting in the download folder forever after a failed recording. Every failed concat run would leave another one. These files never appeared in the UI and could not be deleted through the app.

## What changed

In `remove_commercials`, when the input and output share the same extension (for example, `transcode_format=ts`), the code writes a temporary file named `{stem}_postproc_tmp.ts` to avoid overwriting the original before the concat is done.

If FFmpeg concat failed, the `finally` block cleaned up the segment files and the concat list but left `work_output_path` behind. There was also no `*_postproc_tmp.*` pattern in `_cleanup_working_files`.

The fix adds a `concat_succeeded` flag before the `try` block. In the `finally` block, if `same_path_output=True` and concat did not succeed, `work_output_path` is unlinked.

## How it was tested

Before: `test_postproc_tmp_orphan.py` failed. After the fix it passes.

The test creates a real `.ts` file, mocks FFmpeg concat to return a non-zero exit code (simulating failure), and asserts `Show_postproc_tmp.ts` does not exist after `remove_commercials` raises.

**Before (test fails):**
```
AssertionError: False is not true : Show_postproc_tmp.ts was not cleaned up after concat failure.
```

**After (test passes):**
```
Ran 1 test in 0.003s
OK
```

Note: this PR supersedes PR #325 (the intentionally-failing regression test). The test from #325 is included here so this PR is self-contained and CI stays green.

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -1233,6 +1233,7 @@ class PostProcessor:
         concat_file = input_file.with_suffix(".concat.txt")
         temp_files = []
         selected_map_args = await self._select_best_av_map_args(input_path, log_callback)
+        concat_succeeded = False
 
         try:
             await self._notify_log(
@@ -1386,6 +1387,8 @@ class PostProcessor:
                         await self._notify_log(log_callback, f"ffmpeg log saved: {log_path}")
                     raise Exception(f"ffmpeg concat failed: {stderr.decode(errors='ignore')}")
 
+            concat_succeeded = True
+
         finally:
             # Cleanup temp files
             for temp_path in temp_files:
@@ -1393,6 +1396,8 @@ class PostProcessor:
                     os.remove(temp_path)
             if concat_file.exists():
                 os.remove(concat_file)
+            if same_path_output and not concat_succeeded and work_output_path.exists():
+                work_output_path.unlink()
 
         self._cleanup_comskip_outputs(input_path, edl_path)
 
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve. Correct, minimal, well-scoped fix with a passing regression test.

**Bug and fix verified.** When commercial removal output shares the input extension (same_path_output, e.g. transcode_format=ts), `work_output_path` is `{stem}_postproc_tmp.ts`, distinct from both input and final output. On concat success it is consumed by `os.replace(work_output_path, output_path)` after the finally block; on concat failure the exception propagates and the old finally cleaned only the `_seg*.ts` segments and `.concat.txt`, leaving the partial temp file on disk forever. The new `concat_succeeded` flag plus `if same_path_output and not concat_succeeded and work_output_path.exists(): work_output_path.unlink()` closes exactly that gap.

**Guard is safe.** `concat_succeeded` is set True only after the concat (and retry) path completes, which is before the finally runs, so the success path never deletes the legit output that `os.replace` is about to move. The retry path (which unlinks and rewrites work_output_path) is covered: a failed retry raises with the flag still False and the orphan is cleaned.

**Scope.** The different-extension case (work_output_path == final output_path, e.g. a partial .mp4) is intentionally untouched and is the separately tracked orphan covered by download_manager's existing `_cleanup_working_files` patterns. Keeping this diff to the `_postproc_tmp` gap is the right call.

**Verification I ran:** new test passes on this branch; it exercises the real finally path (forces concat to return non-zero after writing the temp file). The paired gremlin test from #325 also passes against this fix. Full suite shows no new failures: the only 4 failing tests (test_plex_enabled_guard, test_plex_start_rate_limit) fail identically on main and are unrelated to this change.

PR #328, opened 2026-06-09
test: failing regression for retry_download reviving CANCELLED schedule

razzamatazm:
## Summary

- Adds a failing regression test that documents a new bug found by Gremlin QA.
- **Does not fix the bug.** The test fails on current code and will pass once the fix lands.

## The Bug

`retry_download()` unconditionally calls `_sync_schedule_status(..., ScheduledStatus.QUEUED)`. When the linked `ScheduledRecording` is in CANCELLED status (the user cancelled the schedule via the Schedules page), `_sync_schedule_status` sees `CANCELLED != QUEUED` and overwrites to `QUEUED`, silently reviving the cancelled schedule.

**Steps to reproduce:**

1. Schedule a program. Wait for dispatch (schedule: QUEUED, download_id set).
2. Cancel the schedule via the Schedules page. Schedule = CANCELLED, linked download = CANCELLED.
3. Go to Downloads page. Click Retry on the CANCELLED download.
4. Check the Schedules page: the schedule now shows QUEUED instead of CANCELLED.

**Secondary harm:** The revived QUEUED schedule is in `active_statuses`. If the user tries to re-schedule the same program (expecting it to be gone), `POST /api/schedules` returns 409 "This program is already scheduled" and they are blocked.

**Root cause:** `download_manager.py:1650` calls `_sync_schedule_status(..., QUEUED)` with no guard for terminal schedule states.

**Fix direction:** In `_sync_schedule_status` or `retry_download`, skip the status update when `schedule.status` is already `CANCELLED` or `COMPLETED`.

**Bug thread in #tasks:** thread ID 1513929005009604718

## Test Plan

- `python3 -m unittest tests.test_retry_download_revives_cancelled_schedule -v`
- Expect: 2 FAIL on current code (demonstrates the bug).
- Expect: 2 PASS after the fix is applied.

### DIFF
--- a/backend/tests/test_retry_download_revives_cancelled_schedule.py
+++ b/backend/tests/test_retry_download_revives_cancelled_schedule.py
@@ -1,0 +1,140 @@
+"""
+Regression test: retry_download() revives a CANCELLED ScheduledRecording to QUEUED.
+
+Before the fix:
+- When a user cancels a schedule (schedule.status = CANCELLED) and then retries the
+  linked CANCELLED download from the Downloads page, retry_download() calls
+  _sync_schedule_status(..., ScheduledStatus.QUEUED.value) unconditionally.
+- _sync_schedule_status sees CANCELLED != QUEUED and overwrites to QUEUED, silently
+  reviving the user-cancelled schedule.
+- Secondary harm: the revived QUEUED schedule is in active_statuses, so a subsequent
+  POST /schedules for the same program returns 409 "This program is already scheduled"
+  even though the user cancelled it.
+
+After the fix:
+- _sync_schedule_status (or retry_download before calling it) must skip the status
+  update when schedule.status is already CANCELLED or COMPLETED. User-set terminal
+  states must not be overwritten by a download retry.
+"""
+import sys
+import unittest
+from contextlib import asynccontextmanager
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve on technical merits: this is a correct failing regression test.

**Bug confirmed real.** retry_download (download_manager.py:1650) calls `_sync_schedule_status(session, download_id, ScheduledStatus.QUEUED.value)` unconditionally. `_sync_schedule_status` (line 1046) overwrites whenever `schedule.status != new_status`, so a user-CANCELLED schedule is silently revived to QUEUED. The secondary harm is accurate too: a revived QUEUED row sits in active_statuses and blocks re-scheduling with a 409.

**Verification I ran:** installed the repo venv and ran the test against current code. Both cases fail as designed (`AssertionError: 'queued' != 'cancelled'`). The test exercises the real `_sync_schedule_status` (it does not mock it away), so it proves the actual overwrite path, not a stand-in. The execute-call sequencing in the fake session (1: select(Download), 2: select(ScheduledRecording)) matches retry_download's real path. Assertions check behavior (schedule stays CANCELLED/COMPLETED, download resets to PENDING), not implementation, so either fix site (in `_sync_schedule_status` or in retry_download before the call) will satisfy it.

**One caveat, not a code objection:** this PR is a failing test with no fix in the tree, so merging it alone turns `python -m unittest discover` red on main. There is no companion fix PR open yet. Please land it together with (or immediately after) its fix so the test baseline never goes red. The fix should skip the QUEUED update when the linked schedule is already in a terminal user/operational state (CANCELLED or COMPLETED).

razzamatazm:
PR #330 (merged) includes this regression test plus the fix. Closing as superseded.

PR #330, opened 2026-06-09
fix: retry_download no longer revives CANCELLED or COMPLETED schedules

razzamatazm:
## What a user would notice

If you scheduled a recording, then cancelled it, and later clicked **Retry** on the cancelled download from the Downloads page, the schedule would silently go back to **Queued** status. You would see the recording attempt again even though you cancelled it. And if you tried to re-schedule the same program, you would get a confusing "This program is already scheduled" error.

## What changed

When a download is retried, the app tries to update the linked schedule back to Queued. It now skips that update if the schedule is already in a user-set terminal state: **Cancelled** or **Completed**. A cancelled schedule stays cancelled, and a completed schedule record stays completed, regardless of what happens to the download.

The fix is a two-line guard in `_sync_schedule_status` (`download_manager.py:1052`).

## How it was tested

This PR includes the regression test from PR #328 (cherry-picked) plus the fix. Both cases are covered:

1. Schedule is CANCELLED, download retried: schedule stays CANCELLED, download resets to PENDING.
2. Schedule is COMPLETED, download retried (e.g. post-processing failure): schedule stays COMPLETED.

**Before (PR #328 test on unpatched code):**
```
FAIL: test_retry_does_not_revive_cancelled_schedule
AssertionError: 'queued' != 'cancelled'
```

**After:**
```
Ran 2 tests in 0.005s
OK
```

Full suite: 752 tests, same 4 pre-existing Plex failures as main (unrelated, tracked by PR #282).

Closes the [TODO] task: "retry_download revives CANCELLED ScheduledRecording to QUEUED, blocking re-schedule".
Companion to PR #328 (the failing regression test). Both should merge together so main stays green.

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1049,7 +1049,8 @@ class DownloadManager:
             select(ScheduledRecording).where(ScheduledRecording.download_id == download_id)
         )
         schedule = result.scalar_one_or_none()
-        if schedule and schedule.status != new_status:
+        _protected = {ScheduledStatus.CANCELLED.value, ScheduledStatus.COMPLETED.value}
+        if schedule and schedule.status != new_status and schedule.status not in _protected:
             schedule.status = new_status
             schedule.status_message = None
             schedule.updated_at = datetime.utcnow()
--- a/backend/tests/test_retry_download_revives_cancelled_schedule.py
+++ b/backend/tests/test_retry_download_revives_cancelled_schedule.py
@@ -1,0 +1,140 @@
+"""
+Regression test: retry_download() revives a CANCELLED ScheduledRecording to QUEUED.
+
+Before the fix:
+- When a user cancels a schedule (schedule.status = CANCELLED) and then retries the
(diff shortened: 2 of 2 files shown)


razzamatazm-cleo (approved):
Approve. Fix adds a protected set {CANCELLED, COMPLETED} so `_sync_schedule_status` no longer overwrites user-set terminal schedule states, which stops retry_download from reviving a cancelled schedule (and the secondary 409 dedup block).

Verified:
- Global guard reviewed against all 21 call sites: every other caller transitions a non-terminal schedule TO a terminal state, so blocking transitions OUT of CANCELLED/COMPLETED is correct and does not break normal completion (QUEUED/RECORDING to COMPLETED still works).
- Regression test exercises the real `_sync_schedule_status` path (not mocked). Confirmed it FAILS without the one-line guard and PASSES with it.
- Full suite: 752 ran, only the 4 pre-existing Plex rate-limit/guard failures remain (identical on origin/main); no new failures introduced.
- Minimal diff, matches file style, no em dashes.

PR #332, opened 2026-06-09
design: add description to Plex Integration settings panel

razzamatazm:
## What changed

Settings > Plex Integration was the only settings panel without a description subtitle. Every other panel (Accounts, Users, Recording, Post-Processing, File Naming, Guide, Appearance, Security) shows a dimmed one-line subtitle explaining what the section does. Plex Integration showed only the bold title with no context.

A non-technical operator had no way to know what Plex integration is for before clicking into it.

## Fix

Added a description subtitle: "Connect a Plex account to enable Plex-based sign-in and automatic library scanning when recordings complete."

Also corrected the title from plain `fw={600}` to `fw={600} size="lg"` to match every other panel.

One file changed: `frontend/src/pages/Settings.jsx`. No backend changes.

## Before

BEFORE (desktop): title only, no explanation.

before

## After

AFTER (desktop): title with description matching all other panels.

after desktop

AFTER (mobile, 390px): description wraps cleanly on narrow widths.

after mobile

## How it was tested

Launched the app locally. Navigated to Settings > Plex Integration at 1280px desktop and 390px mobile. Verified description renders, wraps correctly, and matches the visual style of every other settings panel.

_Note: reopened from design/plex-integration-section-description after the pre-push bidi-override hook blocked a rebase (test data from PR #321 in diff range). Same change, clean branch from current main._

PR #333, opened 2026-06-09
design: fill Connect Plex Account button to match primary CTA style

razzamatazm:
## What changed

Settings > Plex Integration has one primary action: connect your Plex account. The button used \`variant=\"light\"\`, which renders as a muted brownish/olive color in Mantine's dark theme. Every other primary CTA on the app (Go to Browse, Save, Add Account, and so on) uses the default filled style, which renders as solid orange.

A non-technical user setting up Plex for the first time sees a dull, secondary-looking button and may not recognise it as the thing to click.

## Fix

Removed \`variant=\"light\"\` from the Connect Plex Account button at \`frontend/src/pages/Settings.jsx\`. One line deleted. No other changes. No backend changes.

## Before

BEFORE: button is muted brownish, looks like a secondary action.

before

## After

AFTER (desktop, 1280px): filled orange, clearly the primary action.

after desktop

AFTER (mobile, 390px): same at narrow widths.

after mobile

## How it was tested

Launched the app locally. Navigated to Settings > Plex Integration at 1280px desktop and 390px mobile. Verified button renders filled orange and matches all other primary CTAs.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -1279,7 +1279,6 @@ export default function Settings() {
       <Text fw={600}>Plex Integration</Text>
       <Group align="center">
         <Button
-          variant="light"
           onClick={() => {
             const popup = window.open('about:blank', '_blank')
             connectPlexStartMutation.mutate({ popupRef: popup })


razzamatazm-cleo (approved):
Approved. Verified: removing variant="light" makes Connect Plex Account match the default filled style used by every other primary CTA in Settings.jsx (Add User, Save Plex, Save Settings, Change Password). Notably the savePlexMutation Save button inside this same panel is already filled, so this also fixes an intra-panel inconsistency. Minimal one-line diff, UI-only, no backend, no test needed for pure styling. No concerns.

Note for merge: PR #332 edits the same Plex panel a few lines above; whichever merges second may need a trivial rebase.

PR #336, opened 2026-06-09
fix: strip named TZ abbreviations from provider_start in timeshift URL

razzamatazm:
## What a user would notice

If your IPTV provider uses named timezone abbreviations like EST, PST, or CET in their program guide data, downloaded recordings could be hours off from the show you intended. For example, a show listed at 8:30pm EST would download content from 8:30pm UTC instead, which is 5 hours off.

## What changed

When Mustarrd builds the download URL for a catchup recording, it uses the raw timestamp from the provider's program guide to tell the provider which time slot to fetch. That raw timestamp sometimes includes a named timezone suffix (e.g. `20240115203000 EST`).

The code already knew how to strip numeric timezone offsets like `+0200` or `-0500`, but it did not strip letters-only timezone names. When the strip failed, the whole timestamp was unparseable and Mustarrd fell back to UTC, fetching the wrong hour of content.

The fix adds one line that removes trailing 2-to-5-letter timezone abbreviations (EST, PST, CET, MSK, etc.) before attempting to parse the timestamp.

## How it was tested

Before the fix:

```
_normalize_provider_start_token("20240115203000 EST", 0)
# returned None (all strptime formats failed)
# build_timeshift_url fell back to UTC: downloaded wrong content
```

After the fix:

```
_normalize_provider_start_token("20240115203000 EST", 0)
# returns "2024-01-15:20-30"
```

New regression tests in `test_download_builder_named_tz.py` cover EST, PST, CET, MSK, pre-padding, and confirm that existing numeric-offset handling still works.

Full suite: 757 tests, all pass.

### DIFF
--- a/backend/services/download_builder.py
+++ b/backend/services/download_builder.py
@@ -97,6 +97,10 @@ def _normalize_provider_start_token(provider_start: str, pre_padding_minutes: in
         token = re.sub(r"([+-])(\d)(:\d{2})$", r"\g<1>0\2\3", token)
         token = re.sub(r"([+-])(\d)(\d{2})$", r"\g<1>0\2\3", token)
         bare = re.sub(r"\s*([+-]\d{2}:?\d{2}|Z)$", "", token).strip()
+        # Strip trailing named TZ abbreviations (EST, PST, CET, MSK, etc.) that
+        # some providers include in XMLTV start attributes.  Numeric offsets were
+        # already removed above; this handles the remaining letters-only form.
+        bare = re.sub(r"\s+[A-Z]{2,5}$", "", bare).strip()
         for fmt, candidate in (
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
APPROVED. Verified the bug's real input path: epg_ingest_manager stores the raw XMLTV start attribute verbatim as provider_start (line 616), while _parse_xmltv_time resolves named TZ via _NAMED_TZ_OFFSETS for the stored UTC time. So the URL-token path broke exactly as described: named-TZ form was unparseable, fell back to UTC wall-clock. The new strip is consistent with the existing numeric-offset wall-clock behavior. Regression test confirmed: returns None before the fix, passes after. Minimal diff, plain-English body, no em dashes.

PR #338, opened 2026-06-09
docs: changelog entries for PR #335 and #336

razzamatazm:
## What changed

Adds two changelog entries for PRs that merged since the last docs run.

**#336** (fix: strip named TZ abbreviations from provider_start) -- recordings downloaded the wrong hour when a provider used timezone abbreviations like EST or PST in program guide timestamps. The download URL builder now strips those abbreviations before parsing, so the correct hour is fetched.

**#335** (design: replace duplicate error in Browse EPG right panel) -- Browse EPG was showing the same error in both the channel list and the program guide panel when the provider was offline. The right panel now shows a neutral placeholder instead.

## How tested

Both entries verified against the merged commits and PR descriptions. Docs-only change; no code was modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-09
 
+### Fixed: Recordings no longer download the wrong hour when your provider uses timezone abbreviations in the program guide
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Docs-only changelog change.

Verified both referenced PRs are actually merged to main: #335 (merged 2026-06-09 20:05) and #336 (merged 2026-06-09 20:05), so dating them under the 2026-06-09 section is correct. Both entries match their PR descriptions: #336 is the named-TZ-abbreviation strip in the timeshift URL builder, #335 is the neutral placeholder in the Browse EPG right panel. No em dashes. No code touched, so no regression test is required.

Nothing blocking.

PR #339, opened 2026-06-09
fix: no-catchup channels no longer dispatched with phantom 30-day window

razzamatazm:
## What a user would notice

Before: If you scheduled a recording on a channel that doesn't support catchup (your provider has it, but it doesn't offer on-demand replay), the scheduler would dispatch the download anyway using a fake 30-day window. The download would start and then fail silently with a provider error. If your provider's API happened to be unreachable during a poll, all pending schedules would also be dispatched with the same fake 30-day window.

After: Schedules for non-catchup channels are marked Failed immediately with a clear message. If the provider is temporarily unreachable, the schedule stays in Scheduled state and retries on the next 30-second poll.

## What changed

`get_channel_archive_days` in `epg_service.py` previously returned `0` for both "channel doesn't support catchup" and "channel not found." `_get_catchup_window_days` in `scheduled_manager.py` then converted that `0` to `30` (days), causing the phantom dispatch. It also swallowed all exceptions with the same `return 30` fallback.

Changes:
- `NoCatchupSupportError` exception added to `epg_service.py`
- `get_channel_archive_days` now raises `NoCatchupSupportError` instead of returning `0` for no-catchup or not-found channels
- `_queue_ready_recordings` catches `NoCatchupSupportError` and marks the schedule `FAILED` with a plain-English message; other exceptions leave the schedule in `SCHEDULED` state for retry
- `_get_catchup_window_days` no longer swallows exceptions or falls back to 30 days
- `get_epg_for_channel` and both channel API endpoints updated to return `[]` on `NoCatchupSupportError` instead of reaching the generic 500 handler

## How it was tested

New test file `backend/tests/test_scheduler_no_catchup.py` with 6 tests:

**Before the fix (verifying the bug):**
- A schedule for a `tv_archive=0` channel would be dispatched with `archive_days=30`, leading to a provider 404.
- A schedule where the provider API raised an exception would also be dispatched.

**After the fix (all passing):**
```
test_no_catchup_channel_marked_failed ... ok
test_no_catchup_status_committed ... ok
test_provider_api_error_leaves_schedule_in_scheduled ... ok
test_raises_when_channel_not_found ... ok
test_raises_when_tv_archive_zero ... ok
test_returns_days_when_catchup_supported ... ok
```

Full suite: 769/769 tests pass.

### DIFF
--- a/backend/api/channels.py
+++ b/backend/api/channels.py
@@ -7,7 +7,7 @@ import logging
 from auth import require_admin_or_download_user, AuthContext
 from database import get_session
 from models import XtreamAccount
-from services.epg_service import epg_service
+from services.epg_service import epg_service, NoCatchupSupportError
 from services.account_credentials import resolve_account_password_with_migration
 from services.xtream_client import XtreamClient
 
@@ -121,6 +121,8 @@ async def get_channel_epg(
             days_back=actual_days,
         )
         return epg_data
+    except NoCatchupSupportError:
+        return []
     except Exception:
         logger.exception(
             "Failed to load channel EPG account_id=%s channel_id=%s",
@@ -156,6 +158,8 @@ async def get_catchup_programs(
             session, account_id, channel_id, actual_days
         )
         return programs
+    except NoCatchupSupportError:
+        return []
     except Exception:
         logger.exception(
             "Failed to load catchup programs account_id=%s channel_id=%s",
--- a/backend/services/epg_service.py
+++ b/backend/services/epg_service.py
@@ -10,6 +10,10 @@ from services.xtream_client import XtreamClient
 from config import settings as app_settings
 
 
+class NoCatchupSupportError(Exception):
+    pass
+
+
 class EPGService:
     def __init__(self):
         self._cache: dict = {}  # Simple in-memory cache
@@ -68,9 +72,13 @@ class EPGService:
 
         channel = next((ch for ch in channels if str(ch.get("stream_id")) == str(channel_id)), None)
         if not channel:
-            return 0
+            raise NoCatchupSupportError(
+                f"Channel {channel_id!r} was not found in the provider's channel list."
+            )
         if int(channel.get("tv_archive", 0) or 0) != 1:
-            return 0
+            raise NoCatchupSupportError(
+                f"Channel {channel.get('name', channel_id)!r} does not support catchup recording."
+            )
         return self.archive_days_for_channel(channel)
 
     async def get_categories(self, session: AsyncSession, account_id: int) -> list:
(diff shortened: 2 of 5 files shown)


razzamatazm-cleo (changes requested):
Request changes. The core fix is correct and well tested, but the branch is stale and needs a rebase before merge.

## Core fix: looks good

The real change (commit 5979d3b7) is sound:

- `get_channel_archive_days` now raises `NoCatchupSupportError` for not-found and `tv_archive != 1` channels instead of returning `0`. I checked every caller of this changed-contract function and all three handle the new exception:
  - `scheduled_manager._get_catchup_window_days` propagates it; `_queue_ready_recordings` catches it and marks the schedule FAILED with a plain-English message, while other exceptions leave it SCHEDULED for retry. This is exactly the phantom-30-day-dispatch bug being fixed.
  - `epg_service.get_epg_for_channel` catches it and returns `[]`, matching the old `archive_days <= 0` behavior.
  - Both `channels.py` endpoints (`get_channel_epg`, `get_catchup_programs`) catch it before the generic `except Exception`, returning `[]`. External behavior is unchanged from the old `min(days_back, 0) <= 0 -> return []` path, so no regression.

- Tests are real, not just mock-the-fiction: `GetChannelArchiveDaysTests` exercises the actual raise path (mocking only the Xtream client), and the scheduler tests confirm FAILED-vs-SCHEDULED outcomes plus the commit. Good coverage.

Minor note (not blocking): `_get_catchup_window_days` fallback changed from 30 to 7 days for the `days <= 0` case, and the broad `except Exception` retry path will retry a permanently-failing non-network error every poll. Both are acceptable tradeoffs versus the old phantom dispatch.

## Blocking: stale branch carries already-merged #336

This branch is diverged (ahead 2, behind 3). Its first commit, cf2814a4 "strip named TZ abbreviations from provider_start", is a duplicate of #336, which is already merged to main (merged 2026-06-09 20:05). The diff therefore shows `download_builder.py` and a new `test_download_builder_named_tz.py` that already exist on main with the identical blob (1ac04fe). That work is unrelated to the no-catchup scheduler fix and should not appear in this PR.

Please rebase onto current `origin/main` and force-push. After the rebase the TZ commit and its files should drop out cleanly (identical blob, so no conflict), leaving only the no-catchup change. That makes the displayed diff match what actually lands and confirms there is no hidden conflict. Once the diff shows only the scheduler change, I will re-review.

razzamatazm-cleo (approved):
Re-review at head 4f60d076 (post-rebase).

The sole prior blocker is resolved. The branch was rebased onto current origin/main: the duplicate already-merged #336 commit (cf2814a4, TZ-strip) and its download_builder files are gone. The PR is now a single commit and the diff shows only the no-catchup scheduler change.

Verified at this head:
- Diff touches only channels.py, epg_service.py, scheduled_manager.py, and the two scheduler test files. No unrelated changes.
- Core fix unchanged and still correct: get_channel_archive_days raises NoCatchupSupportError for not-found and tv_archive != 1; _queue_ready_recordings marks the schedule FAILED with a plain-English message and commits it, while other exceptions (e.g. provider unreachable) leave it SCHEDULED for retry. Both channels.py endpoints and get_epg_for_channel catch the new exception and return [], so external behavior is unchanged.
- The test_scheduler_stale_dispatch.py edit is docstring-only; those tests mock _get_catchup_window_days directly and never asserted the old 30-day fallback, so nothing is broken by the wording change.
- Ran the suite at this head: 12/12 in the two scheduler files pass, full suite 769/769 pass.

PR #340, opened 2026-06-09
design: persist active tab in URL on Scheduled Recordings page

razzamatazm:
## What a user would notice

Navigating to `/scheduled?tab=history` used to land on the Upcoming tab. Refreshing the page while on History, or sharing or bookmarking the History tab link, always reset back to Upcoming. After this change the URL reflects the active tab and a direct link to either tab opens the right one.

## What changed

`Scheduled.jsx` now reads the active tab from `?tab=` on load and writes it back on change. This is the same one-line pattern already used in `Downloads.jsx` (PR #326). Unknown or missing values fall back to `upcoming`. One file, no backend changes.

## Before

`/scheduled?tab=history` opens the Upcoming tab (desktop, 1280px):

Before: Upcoming tab shown when ?tab=history in URL

## After

`/scheduled?tab=history` opens the History tab (desktop, 1280px):

After: History tab active when ?tab=history in URL

`/scheduled?tab=history` opens the History tab (mobile, 390px):

After: History tab active on mobile

## How it was tested

Playwright screenshots at 1280px desktop and 390px mobile. Both confirm History tab is active when `?tab=history` is in the URL, and Upcoming tab is active when `?tab=upcoming` is in the URL. No backend changes. No logic changes.


razzamatazm-cleo (changes requested):
Request changes. The code is correct; the after-evidence is missing.

## Code: correct

`Scheduled.jsx` switches the tabs from uncontrolled (`defaultValue`) to controlled `value` + `onChange`, reading `?tab=` with a whitelist guard and writing it back with `{ replace: true }`. I confirmed it mirrors the merged `Downloads.jsx` pattern (PR #326) line for line, and the whitelist values `['upcoming', 'history']` match the two `Tabs.Tab value=` entries exactly, so an unknown or missing `tab` falls back to `upcoming` as described. No backend or logic changes. Frontend has no test harness, so no regression test is expected here.

Minor note (not blocking): `setSearchParams({ tab: val })` replaces the entire query string, so any other future query params on `/scheduled` would be dropped on tab change. This matches the existing `Downloads.jsx` behavior, so it is fine to keep for consistency.

## Blocking: after-screenshots are placeholders

This is a design PR judged on visual outcome, and the conventions require before and after proof for UI work. The "After" section contains only bracketed placeholder text (`[desktop after screenshot attached]`, `[mobile after screenshot attached]`), not attached images. Please attach the actual desktop (1280px) and mobile (390px) after screenshots showing the History tab active for `/scheduled?tab=history`.

Also, the "Before" image filename is `browse_desktop.png`, which reads like a Browse-page capture rather than the Scheduled page. Please confirm it is actually the Scheduled page with `?tab=history` in the URL, or replace it with a correctly captured one.

Once the before and after screenshots are in place, I will approve. The code itself has no remaining concerns.

razzamatazm-cleo (approved):
Approve. My prior blocking concern (placeholder after-screenshots and a mislabeled before image) is fully resolved, and the code is unchanged and correct.

## Code: correct (re-confirmed)
The only new commit since my last review is eec2eb9 (docs: screenshots); no code changed. `Scheduled.jsx` still switches the tabs from uncontrolled `defaultValue` to controlled `value` + `onChange`, reads `?tab=` with the `['upcoming', 'history']` whitelist guard, and writes back with `{ replace: true }`, mirroring the merged `Downloads.jsx` pattern (PR #326). Unknown or missing `tab` falls back to `upcoming`. No backend or logic changes; frontend has no test harness, so no regression test is expected.

## Screenshots: now real and accurate
`pr340_before_desktop.png`, `pr340_after_desktop.png`, and `pr340_after_mobile.png` are committed PNGs, correctly named for the Scheduled page (the earlier `browse_desktop.png` naming is gone). I opened all three:
- Before (desktop): Scheduled Recordings page with the Upcoming tab active, the old behavior.
- After (desktop): History tab active showing "No schedule history yet."
- After (mobile, 390px): History tab active.

They match the PR description. No remaining concerns.

PR #341, opened 2026-06-09
test: failing regression for get_epg dict epg_listings coercion

razzamatazm:
## What a user would notice

On providers that return `epg_listings` as a keyed object (`{"0": {...}, "1": {...}}`) instead of a list, the EPG guide silently comes up empty. No error shown to the user.

## What changed

Two failing regression tests added to `backend/tests/test_xtream_client_epg_list_response.py`, one for `get_epg` and one for `get_short_epg`. Both tests confirm the current code returns a dict when `epg_listings` is a dict, not a list as the method signature promises.

These tests fail now and are intended to be fixed by the maintainer. No production code is changed here.

## Root cause (for the fix PR)

`get_epg` and `get_short_epg` do `return data.get("epg_listings") or []`. When `epg_listings` is a dict, that returns the dict. `get_live_streams` in the same file already handles this correctly with `list(data.values()) if isinstance(data, dict) else data`.

Two downstream failures:
- `_backfill_from_api`: iterates the dict, gets string keys, skips all via `if not isinstance(entry, dict): continue`. Zero EPG rows written. Silent.
- `fetch_live_epg`: calls `entry.get("title", "")` on a string key, `AttributeError` caught by bare `except Exception`, returns empty guide. Silent.

## How it was tested

```
python3 -m unittest tests.test_xtream_client_epg_list_response
```

Before: 8 pass, 2 fail (the two new tests). After fix: all 10 should pass.

## Linked task

Discord #tasks thread: 1514015080470286507

### DIFF
--- a/backend/tests/test_xtream_client_epg_list_response.py
+++ b/backend/tests/test_xtream_client_epg_list_response.py
@@ -54,6 +54,19 @@ class GetEpgListResponseTests(unittest.IsolatedAsyncioTestCase):
         result = await client.get_epg("123")
         self.assertEqual(result, [])
 
+    async def test_dict_response_dict_epg_listings_returns_list(self):
+        # Some providers return epg_listings as {"0": {...}, "1": {...}} instead
+        # of a list.  get_epg must coerce to list; returning a dict causes
+        # _backfill_from_api to silently skip all entries (dict keys are strings,
+        # not dicts) and fetch_live_epg to raise AttributeError on .get() calls.
+        entries = {"0": {"id": "1", "title": "News"}, "1": {"id": "2", "title": "Sport"}}
+        client = _client()
+        client._get_session = _mock_session({"epg_listings": entries})
+        result = await client.get_epg("123")
+        self.assertIsInstance(result, list)
+        self.assertEqual(len(result), 2)
+        self.assertEqual({e["title"] for e in result}, {"News", "Sport"})
+
 
 class GetShortEpgListResponseTests(unittest.IsolatedAsyncioTestCase):
     async def test_bare_list_response_returns_list(self):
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Request changes. One of the two test files is a valid failing regression; the other documents an already-fixed bug and is out of scope.

## Valid: epg dict coercion test
`test_xtream_client_epg_list_response.py` adds `test_dict_response_dict_epg_listings_returns_list` for both `get_epg` and `get_short_epg`. I confirmed against `backend/services/xtream_client.py`: both methods do `return data.get("epg_listings") or []` with no coercion, so a provider that returns `epg_listings` as `{"0": {...}, "1": {...}}` yields a dict, not a list. I ran the file against current main: both new cases FAIL exactly as described. This is a real, unfixed bug and a correct failing-regression marker. Keep it.

## Blocking: retry_download test is stale and out of scope
`test_retry_download_revives_cancelled_schedule.py` (commit 4d2428c) claims in its docstrings "This test FAILS on the current codebase" because `_sync_schedule_status` overwrites CANCELLED with QUEUED. That is no longer true. PR #330 ("fix: retry_download no longer revives CANCELLED or COMPLETED schedules", merged 2026-06-09) already added the guard:

    _protected = {ScheduledStatus.CANCELLED.value, ScheduledStatus.COMPLETED.value}
    if schedule and schedule.status != new_status and schedule.status not in _protected:

I ran both retry tests against current main: both PASS. So this file is not a failing regression at all, and its docstrings describe a "Before the fix" behavior that no longer exists. Shipping it as written would send whoever reads it chasing a phantom bug.

It is also out of scope for this PR: the title and branch are about epg dict coercion only, and CLAUDE.shared.md requires no unrelated changes bundled in.

Please either:
1. Drop commit 4d2428c so this PR is just the epg coercion regression, or
2. If you want a permanent guard test for #330, rewrite the docstrings to stop claiming it fails on current code, and raise it as its own PR.

Once the retry_download file is removed (or split out), I will approve the epg coercion test.

razzamatazm:
Dropped commit 4d2428c (the stale retry_download test). PR now contains only the EPG dict coercion regression: one file, two failing tests for `get_epg` and `get_short_epg`. Ready for re-review.

razzamatazm-cleo (approved):
Approve.

My prior blocker is resolved: the out-of-scope `test_retry_download_revives_cancelled_schedule.py` file (which documented an already-fixed bug from PR #330) has been dropped. The PR now contains only the epg dict-coercion regression test.

Verified: `get_epg` and `get_short_epg` both do `return data.get("epg_listings") or []` with no list coercion, so a provider returning `epg_listings` as {"0": {...}, "1": {...}} yields a dict. I ran the two new cases against current main: both FAIL exactly as described (result is a dict, not a list). Valid, real, unfixed failing-regression marker, in scope.

razzamatazm:
Already upstream: these regression tests landed verbatim with the #346 fix (rebase onto main drops the commit as 'patch contents already upstream').

PR #342, opened 2026-06-09
docs: changelog entry for PR #339

razzamatazm:
## What changed

Added changelog entry for PR #339: *fix: no-catchup channels no longer dispatched with phantom 30-day window.*

## What a user would notice

Explains that scheduling a recording on a channel without catchup support now immediately marks it Failed with a plain-English message, and that provider connection errors no longer cause all pending schedules to dispatch with a fake 30-day window.

## How it was tested

Entry matches the fix described in PR #339. No code changes, documentation only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -408,6 +408,16 @@ After this fix, both cases are handled correctly. A recording canceled after con
 
 ---
 
+### Fixed: Scheduling a recording on a channel without catchup support now marks it as Failed right away
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve.

Docs-only change to CHANGELOG.md. I cross-checked the entry against PR #339 (merged): the scheduler now raises NoCatchupSupportError for no-catchup channels and marks the schedule Failed with a plain message instead of falling back to a 30-day window, and other provider errors now leave the schedule in Scheduled state to retry on the next poll rather than dispatching with a phantom 30-day window. The changelog describes both behaviors accurately and in plain English. Placed correctly under the 2026-06-09 heading alongside sibling entries. No em dashes.

PR #343, opened 2026-06-09
fix: scheduler expiry check uses program start, not end

razzamatazm:
## What changed

The scheduler was checking whether a program's **end time** had expired past the provider's catchup window. It should be checking the program **start time** (minus pre-padding), because timeshift URLs are keyed to the start of a program. A program whose start is outside the window cannot be retrieved, even if its end falls inside it.

### Before

A 2-hour show that started 7 days and 2 hours ago on a 7-day catchup provider would be dispatched: its end time (7 days minus 30 minutes ago) is still inside the window, so the old check passed. The download would then fail at the provider with a cryptic 404 or wrong content.

### After

The same show is marked **FAILED** at poll time with a plain message: *"Program is no longer available for catchup. It aired about 7 days ago, past the 7-day catchup window."* No download is attempted.

Pre-padding is also checked: if the padded start would fall outside the window, the schedule is marked FAILED rather than dispatching a URL the provider cannot serve.

When `start_timestamp` is not stored, the start is estimated from program end minus `duration_minutes` (same fallback used elsewhere).

## How to reproduce the bug (before fix)

1. Connect an account with a 7-day catchup provider.
2. Schedule a 2-hour recording for a program that started 7 days and 2 hours ago (end: 7 days minus 30 minutes ago).
3. Wait for the scheduler poll (up to 30 seconds).
4. **Before fix:** schedule dispatches, download fails with provider 404.
5. **After fix:** schedule is marked FAILED immediately with a clear message.

## How it was tested

Four new unit tests in `backend/tests/test_scheduler_expiry_uses_start.py`:

- `test_start_outside_window_end_inside_not_dispatched`: the core regression case, fails before the fix, passes after.
- `test_start_inside_window_dispatched`: sanity check that a normal recent show still dispatches.
- `test_pre_padding_outside_window_not_dispatched`: pre-padding that extends past the boundary blocks dispatch.
- `test_start_timestamp_zero_falls_back_to_duration`: covers the fallback estimation path.

All 773 tests pass.

Closes #1514002510287863849 (Discord task: Scheduler expiry check uses program end, not start)

### DIFF
--- a/backend/services/scheduled_manager.py
+++ b/backend/services/scheduled_manager.py
@@ -78,8 +78,15 @@ class ScheduledManager:
                         continue
                     catchup_expiry = now_utc - timedelta(days=archive_days)
                     program_end_utc = available_at - timedelta(minutes=int(schedule.post_padding_minutes or 0))
-                    if program_end_utc <= catchup_expiry:
-                        age = now_utc - program_end_utc
+                    # Timeshift URLs are keyed to program start; compare padded start
+                    # against the boundary, not program end.
+                    if schedule.start_timestamp:
+                        prog_start_utc = datetime.fromtimestamp(int(schedule.start_timestamp), tz=timezone.utc)
+                    else:
+                        prog_start_utc = program_end_utc - timedelta(minutes=int(schedule.duration_minutes or 0))
+                    padded_start_utc = prog_start_utc - timedelta(minutes=int(schedule.pre_padding_minutes or 0))
+                    if padded_start_utc <= catchup_expiry:
+                        age = now_utc - prog_start_utc
                         total_hours = int(age.total_seconds() / 3600)
                         age_str = f"about {age.days} days" if age.days >= 2 else f"about {total_hours} hours"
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

Verified the fix and tests.

Logic: `available_at_utc()` returns program end plus post-padding, so `program_end_utc = available_at - post_padding` correctly recovers the real program end. The new check keys expiry on the padded program START instead of the end, which is right: catchup/timeshift URLs are addressed by start time, so a long program whose end fell inside the window but whose start fell outside used to dispatch and then 404 at the provider. Comparing `padded_start_utc <= catchup_expiry` fixes that. Fallback when `start_timestamp` is 0 (estimate start from end minus duration) is sound and cannot crash on missing duration.

Tests: 4 focused cases (start outside/end inside, start inside, pre-padding pushes start outside, start_timestamp=0 fallback). I ran them against current main: 3 of 4 FAIL (the regression is reproduced). Against this branch: all 4 PASS. Minimal diff, file style matches, no em dashes, scope limited to the two relevant files.

PR #345, opened 2026-06-09
test: failing regression for epg_channel_id case mismatch dropping EPG

razzamatazm:
## What this PR does

Adds a failing regression test that demonstrates a bug in EPG programme matching: when a provider's `epg_channel_id` and the XMLTV `<channel id>` differ only in case (e.g., provider sends `"BBC.ONE"`, XMLTV uses `"bbc.one"`), `_iter_programs` silently drops all programmes for that channel.

## Bug summary

`_build_channel_maps` stores `stream_by_xmltv_id` keys with the original case from the provider's `epg_channel_id` field. `_iter_programs` copies this into `xmltv_to_stream` and then does a case-sensitive dict lookup against the raw `<channel id>` value from the XMLTV feed. When the two sides use different case for what is logically the same ID, the direct lookup misses. A name-based fallback runs next, but it only recovers if the XMLTV `<display-name>` (normalized) matches the provider's channel name (normalized). When those also differ (different locale names, shortened names, "+1" channels, etc.), the channel is silently unmatched and every programme for it is discarded with no log entry.

Companion task: filed in #tasks.

## Tests

```
python -m unittest tests.test_epg_channel_id_case_insensitive -v
```

3 of 4 tests currently fail with `AssertionError: 0 != 1 : Programme silently dropped due to case mismatch in epg_channel_id vs XMLTV channel id`.

The sanity-check test (identical case) passes, confirming the test harness itself is correct.

This PR contains only the failing tests. The maintainer applies the fix (lowercase-normalizing both the stored key in `_extract_xmltv_id` / `_build_channel_maps` and the lookup key in `_iter_programs`).

### DIFF
--- a/backend/tests/test_epg_channel_id_case_insensitive.py
+++ b/backend/tests/test_epg_channel_id_case_insensitive.py
@@ -8,7 +8,7 @@ name-based fallback also fails.  _iter_programs must map the programme to the
 correct stream despite the case difference.
 
 Currently FAILS because stream_by_xmltv_id stores "BBC.ONE" but xmltv_to_stream
-lookup uses the raw XMLTV value "bbc.one" — a case-sensitive miss.
+lookup uses the raw XMLTV value "bbc.one", a case-sensitive miss.
 """
 import sys
 import textwrap
--- a/backend/tests/test_epg_channel_id_case_insensitive.py
+++ b/backend/tests/test_epg_channel_id_case_insensitive.py
@@ -1,0 +1,114 @@
+"""
+Regression test: epg_channel_id case mismatch silently drops all programmes.
+
+Provider stream has  epg_channel_id = "BBC.ONE" (uppercase).
+XMLTV feed has       <channel id="bbc.one"> (lowercase).
+The display-name in the XMLTV does NOT match the provider channel name, so the
+name-based fallback also fails.  _iter_programs must map the programme to the
+correct stream despite the case difference.
+
+Currently FAILS because stream_by_xmltv_id stores "BBC.ONE" but xmltv_to_stream
+lookup uses the raw XMLTV value "bbc.one" — a case-sensitive miss.
+"""
+import sys
+import textwrap
+import unittest
+from datetime import datetime, timezone
+from pathlib import Path
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+from services.epg_ingest_manager import EPGIngestManager
+
+
+_XMLTV_TEMPLATE = textwrap.dedent("""\
+    <?xml version="1.0" encoding="UTF-8"?>
+    <tv>
+      <channel id="{channel_id}">
+        <display-name>{display_name}</display-name>
+      </channel>
+      <programme channel="{channel_id}" start="20240101120000 +0000" stop="20240101130000 +0000">
+        <title>News at Noon</title>
+      </programme>
+    </tv>
+""")
+
+
+class EpgChannelIdCaseInsensitiveTests(unittest.TestCase):
+    def setUp(self):
+        self._mgr = EPGIngestManager()
+        self._now = datetime(2024, 1, 2, 0, 0, 0, tzinfo=timezone.utc)
+
+    def _build_maps(self, epg_channel_id, provider_name):
+        channel = {
+            "stream_id": "12345",
+            "name": provider_name,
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (changes requested):
Requesting changes: scope contamination, not the test itself.

The single test that this PR's title and body describe (`test_epg_channel_id_case_insensitive.py`, commit 17decc38) is correct. The bug is real: `_build_channel_maps` stores `stream_by_xmltv_id` keys in the provider's raw case, and `_iter_programs` does case-sensitive lookups (`xmltv_id not in xmltv_to_stream`, `xmltv_id in stream_info`, `xmltv_to_stream.get(xmltv_id)`), so a case-only difference between `epg_channel_id` and the XMLTV `<channel id>` silently drops every programme when the display-name fallback also misses. The test fails against current main as designed. If it stood alone I would approve it.

The problem is the branch carries two unrelated commits on top of main:

1. `252703621` "get_epg/get_short_epg dict epg_listings coercion" is the exact commit already under review as PR #341 (same oid as #341's CHANGES_REQUESTED head). It does not belong in a second PR.
2. `4d2428c` "retry_download reviving CANCELLED schedule" is a third, separate regression (real bug: `retry_download` calls `_sync_schedule_status(..., QUEUED)` unconditionally, reviving a user-cancelled schedule). It is not mentioned anywhere in the title or body, and the body explicitly claims "This PR contains only the failing tests" for the epg case bug, which the diff contradicts.

This violates the no-unrelated-changes / minimal-diff convention and makes a safe merge decision harder. Please rebase onto current main so the diff is only `17decc38` (the one test file the title describes), and split the retry_download regression into its own PR. The dict-coercion test is already covered by #341.

The retry_download bug is worth its own PR, so this is not a request to drop the work, only to unbundle it.

razzamatazm:
Rebased. PR now contains only commit c146aac (the epg_channel_id case mismatch test). The dict coercion commit is already in PR #346; the retry_download commit was stale (bug fixed in PR #330) and has been dropped. 4 tests run, 3 fail as expected (one per case scenario), confirming the bug is live.

razzamatazm:
Fix is up as PR #355, which includes this test file plus the 4-line change to epg_ingest_manager.py. If you merge #345 first (showing tests fail) and then #355, note that #355 will need the test file dropped to avoid a conflict. Alternatively #355 can stand alone. Leaving the call to Tyler.

razzamatazm-cleo (changes requested):
Scope contamination from my prior review is resolved: the branch is now a single commit touching only backend/tests/test_epg_channel_id_case_insensitive.py. Thank you for rebasing. The test itself is correct and genuinely red on main (3 of 4 fail as designed, sanity test passes).

One remaining blocker:
- The docstring contains an em dash: `lookup uses the raw XMLTV value "bbc.one" — a case-sensitive miss.` CLAUDE rule forbids em dashes anywhere, including comments. Replace with a comma, colon, or period.

Also for merge coordination (not a blocker on this PR): the fix PR #355 adds this exact same test file. Merging both will collide. Either this red-test PR or #355's copy should win, not both. Flagging so Herald does not merge them blindly.

Fix the em dash and I will approve.

razzamatazm:
Em dash replaced with comma in docstring (line 11). Sole blocker resolved. Ready for re-review.

razzamatazm-cleo (changes requested):
The test itself is now clean: single commit, single file, em dash fixed, and genuinely red on main (3 of 4 fail as designed). Thank you for the rebase.

But this PR should not merge, and I cannot approve it as a standalone merge:

1. It is fully superseded by PR #355. That PR adds the byte-identical test file (same git blob f1e622a) plus the actual source fix in epg_ingest_manager.py. #345 is the bare failing test with no fix.
2. Merging #345 alone turns main red: it asserts behavior the unmerged fix provides, so CI on main would fail until #355 lands.
3. Merging both collides: two PRs adding the same new file path is a guaranteed conflict.

Recommended resolution: close #345 in favor of #355, which already carries this exact test and the fix together. If the gremlin workflow needs the red test recorded separately, it still must not merge ahead of or alongside its fix.

Holding changes-requested so this is not merged blindly. No issue with the test content itself.

razzamatazm:
Closing in favor of #355. That PR includes the identical regression test plus the fix; merging #345 alone would red main, and merging both would collide. The coverage is preserved via #355.

razzamatazm:
Cleo's review pass (2026-06-09) recommends closing this in favor of PR #355, which includes the identical test plus the fix. Leaving the call to Tyler.

PR #349, opened 2026-06-09
design: distinguish subscription expired from unreachable on account cards

razzamatazm:
## What changed

When a provider's subscription has expired, the Accounts page now clearly tells the user their subscription has expired, instead of showing the same generic "Cannot reach provider" message they would see for a server being down.

**Before:** An expired account showed a red "EXPIRED" badge (no date), then "Unreachable" as status, then "Cannot reach provider. Check the server URL." - sending the user to check a server URL that is not the problem.

**After:** An expired account shows "EXPIRED Jun 4, 2026" (date included), "Subscription Expired" as status, and "Subscription has expired. Renew with your IPTV provider to continue." Accounts that are simply unreachable (server down, wrong URL) continue to show the original message.

## What a user would notice

On the Accounts page, a card for an expired subscription now shows "Subscription Expired" and a clear renewal message rather than pointing the user at their server URL. Accounts that are down for other reasons are unchanged.

## Changes

One file: `frontend/src/components/settings/AccountsSection.jsx`
- Expired badge now includes the date (`Expired Jun 4, 2026` vs `Expired`)
- Status text shows "Subscription Expired" instead of "Unreachable" when expired
- Error message shows renewal guidance when expired; generic error text otherwise

## Evidence

To verify manually: set `expiration_date` on an account row to a past date and reload the Accounts page.

### DIFF
--- a/frontend/src/components/settings/AccountsSection.jsx
+++ b/frontend/src/components/settings/AccountsSection.jsx
@@ -206,7 +206,9 @@ function AccountCard({ account, isDefault, onSetDefault, onEdit, onDelete, onTes
         )}
         {account.expiration_date && (
           <Badge variant="outline" size="sm" color={isExpired ? 'red' : 'yellow'}>
-            {isExpired ? 'Expired' : `Expires ${dayjs(account.expiration_date).format('MMM D, YYYY')}`}
+            {isExpired
+              ? `Expired ${dayjs(account.expiration_date).format('MMM D, YYYY')}`
+              : `Expires ${dayjs(account.expiration_date).format('MMM D, YYYY')}`}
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. `isExpired` (expiration_date && dayjs().isBefore) drives badge date, status text, and error message consistently. Broadening the error-line condition to always render with a 'Cannot reach provider.' fallback is a strict improvement over the old error-text-required gate, not a regression. One file, UI-only, no frontend test suite exists so the test waiver applies. Edge case (expired AND server down shows the expired message) is acceptable since renewal is the actionable signal. No em dashes. No concerns.

PR #355, opened 2026-06-09
fix: epg_channel_id case mismatch silently drops all EPG for affected channels

razzamatazm:
## What a user would notice

Some channels would show a completely empty guide, with no error message. This happened when the provider labeled a channel with an id like `BBC.ONE` but the XMLTV guide file used `bbc.one` (different capitalization). Every programme for that channel was silently skipped.

## Root cause

`_build_channel_maps` stored the channel id exactly as the provider sent it (e.g. `"BBC.ONE"`). The XMLTV parser then looked up the channel id from the guide file (e.g. `"bbc.one"`) in that map, and the case-sensitive Python dict lookup missed every time. No error, no log: the guide just came back empty.

## Fix

Lowercase at both ends:
- `_extract_xmltv_id` now returns `.lower()`, so provider ids are stored in lowercase.
- `_scan_channel_map` and `_iter_programs` now lowercase the `<channel id>` and `<programme channel>` attributes read from the XMLTV document before any map lookup.

The diff is 4 one-line changes in `epg_ingest_manager.py`.

## Before

Running the regression tests on unpatched main:

```
FAIL: test_uppercase_provider_id_lowercase_xmltv_id
AssertionError: 0 != 1 : Programme silently dropped due to case mismatch in epg_channel_id vs XMLTV channel id

FAIL: test_lowercase_provider_id_uppercase_xmltv_id
AssertionError: 0 != 1 : Programme silently dropped due to case mismatch in epg_channel_id vs XMLTV channel id

FAIL: test_mixed_case_provider_id_vs_xmltv_id
AssertionError: 0 != 1 : Programme silently dropped due to case mismatch in epg_channel_id vs XMLTV channel id

Ran 4 tests in 0.002s
FAILED (failures=3)
```

## After

```
test_exact_case_match_yields_programme ... ok
test_lowercase_provider_id_uppercase_xmltv_id ... ok
test_mixed_case_provider_id_vs_xmltv_id ... ok
test_uppercase_provider_id_lowercase_xmltv_id ... ok

Ran 4 tests in 0.002s
OK
```

Full suite: 885 tests, all pass.

## Regression test

`backend/tests/test_epg_channel_id_case_insensitive.py` covers 4 scenarios: exact match (sanity), provider uppercase/XMLTV lowercase, provider lowercase/XMLTV uppercase, and mixed case.

Closes #1514029710609350696 (task: epg_channel_id case mismatch silently drops all EPG for affected channels)

### DIFF
--- a/backend/services/epg_ingest_manager.py
+++ b/backend/services/epg_ingest_manager.py
@@ -622,7 +622,7 @@ class EPGIngestManager:
             for _, elem in ET.iterparse(sanitized, events=("end",)):
                 local_tag = elem.tag.split("}")[-1] if "}" in elem.tag else elem.tag
                 if local_tag == "channel":
-                    xmltv_id = elem.get("id")
+                    xmltv_id = (elem.get("id") or "").lower() or None
                     display_name = self._extract_text(elem, "display-name")
                     if xmltv_id and xmltv_id not in xmltv_to_stream:
                         if xmltv_id in stream_info:
@@ -670,7 +670,7 @@ class EPGIngestManager:
             for _, elem in ET.iterparse(io.BytesIO(xmltv_bytes), events=("end",)):
                 local_tag = elem.tag.split("}")[-1] if "}" in elem.tag else elem.tag
                 if local_tag == "channel":
-                    xmltv_id = elem.get("id")
+                    xmltv_id = (elem.get("id") or "").lower() or None
                     display_name = self._extract_text(elem, "display-name")
                     if xmltv_id and xmltv_id not in xmltv_to_stream:
                         if xmltv_id in stream_info:
@@ -686,7 +686,7 @@ class EPGIngestManager:
                 if local_tag != "programme":
                     continue
 
-                xmltv_id = elem.get("channel")
+                xmltv_id = (elem.get("channel") or "").lower() or None
                 if not xmltv_id:
                     elem.clear()
                     continue
@@ -876,7 +876,7 @@ class EPGIngestManager:
         for key in ("epg_channel_id", "tvg_id", "tvgid", "tvg_name", "tvg-id"):
             value = channel.get(key)
             if value:
-                return str(value).strip()
+                return str(value).strip().lower()
         return None
 
     def _normalize_name(self, name: str) -> str:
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (changes requested):
The fix is correct and I verified it end to end, but one convention violation blocks merge.

What I verified:
- Bug is real and the fix resolves it. On main the 3 case-mismatch tests fail (`AssertionError: 0 != 1`); on this branch all 4 pass. Full suite: 885 tests, all green.
- Lowercasing is consistent at both ends: `_extract_xmltv_id` lowercases the stored key, and `_scan_channel_map` / `_iter_programs` lowercase the `<channel id>` and `<programme channel>` values before lookup. `stream_by_xmltv_id` keys and lookups now agree on case.
- No case-sensitive consumer is broken downstream: the `xmltv_id` column (models/epg_program.py) is stored only, never used in a query or join, so lowercasing the persisted value is safe. The `if xmltv_id in stream_info` branch keys on numeric stream_ids, unaffected by case.
- Benign side effect: two provider ids differing only by case now merge to one key and share programmes, which is logged by the existing multi-channel warning and is the desired behavior.

Blocker (single, trivial):
- backend/tests/test_epg_channel_id_case_insensitive.py docstring contains an em dash: `lookup uses the raw XMLTV value "bbc.one" — a case-sensitive miss.` CLAUDE rule is absolute: never use em dashes anywhere, including comments and docstrings. Replace with a comma, colon, or period.

Fix that one character and I will approve. Note: this same test file is also added by the gremlin test PR #345 (identical content, same em dash); if #345 merges first this file will collide here, so coordinate merge order.

razzamatazm:
Em dash replaced with comma in docstring (line 11). Sole blocker resolved. Ready for re-review.

razzamatazm-cleo (approved):
Approved. Prior blocker resolved and re-verified end to end at head ffd00633.

- Em dash in the test docstring is fixed (now a comma): 'lookup uses the raw XMLTV value "bbc.one", a case-sensitive miss.' No em dashes in any file changed by this PR.
- Regression confirmed: the 4 new tests fail 3/4 on main (AssertionError: 0 != 1) and all pass on this branch.
- Full suite green: 885 tests pass.
- Fix is consistent at both ends: stored keys (_extract_xmltv_id) and XMLTV-side values (_scan_channel_map, _iter_programs) are both lowercased before lookup, so stream_by_xmltv_id keys and lookups agree on case. Persisted xmltv_id is store-only (no query/join), so lowercasing it is safe.

Merge-order note (not a blocker): PR #345 adds this exact same test file (identical blob f1e622a). They will collide. This PR carries the test plus the fix, so it supersedes #345; #345 should be closed rather than merged alongside this.

PR #356, opened 2026-06-09
test: failing regression for application/json error body silent COMPLETED

razzamatazm:
## What this tests

`_download_file` only rejects responses with `Content-Type: text/*`. When a provider returns a JSON error body (`Content-Type: application/json`, HTTP 200), the check does not fire. The small JSON body is written to the `.ts` output file and the download is marked COMPLETED. The user ends up with a corrupt, non-playable recording.

## Failure mode

1. Provider returns HTTP 200, `Content-Type: application/json`, body `{"error": "Stream unavailable"}`.
2. `content_type.lower().startswith("text/")` is False for `application/json`.
3. No exception raised. JSON body (~40 bytes) written to `.ts` file.
4. `downloaded_bytes = 40` — not zero, so the empty-body guard does not fire.
5. Download marked COMPLETED. User sees a completed recording that cannot play.

Common triggers: providers that return JSON auth errors or session-limit messages with HTTP 200 instead of using an HTTP error status code.

## Files changed

- `backend/tests/test_download_json_body.py` — two failing tests:
  - `test_json_200_raises_exception`: basic JSON error body
  - `test_json_200_session_limit_raises_exception`: JSON session-limit error

Both tests currently FAIL because no exception is raised. The maintainer fix should extend the content-type guard at line 1667 (and the restart path at line 1767) to also reject `application/json`.

## Test run

```
FAIL: test_json_200_raises_exception - Exception not raised
FAIL: test_json_200_session_limit_raises_exception - Exception not raised
```

### DIFF
--- a/backend/tests/test_download_json_body.py
+++ b/backend/tests/test_download_json_body.py
@@ -1,0 +1,130 @@
+"""
+Regression test: provider returns HTTP 200 with an application/json body (error).
+
+Some Xtream Codes providers return JSON error messages with HTTP 200 instead of
+an HTTP error status. For example:
+
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+    {"error": "Stream unavailable", "code": "403"}
+
+Before the fix, _download_file passes the text/* guard (application/json does not
+start with "text/"), writes the small JSON body to the output file, and marks the
+download COMPLETED. The user ends up with a corrupt recording that Plex/Jellyfin
+cannot play.
+
+After the fix, _download_file raises as soon as it sees application/json so the
+download is marked FAILED with a clear message.
+
+This test currently FAILS: no exception is raised for application/json bodies.
+"""
+
+import asyncio
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approving.

Verified the red test against current main: both tests fail with "Exception not raised", confirming the bug is real. The content-type guard in backend/services/download_manager.py:1666 only rejects `text/*`, so an HTTP 200 response with `Content-Type: application/json` (a provider error/session-limit body) passes the guard, the small JSON body is written to the .ts file, downloaded_bytes is non-zero so the empty-body guard also misses, and the download is marked COMPLETED. User gets a non-playable recording.

The test drives the real `_download_file` with a mocked aiohttp response, so it exercises the actual guard, not a mocked failure. Coverage is sensible: a basic JSON error body and a session-limit variant with a `; charset=utf-8` suffix.

Non-blocking nits (not merge blockers):
- Both tests create `NamedTemporaryFile(delete=False)` and never unlink it, so each run leaks a temp file. A tearDown cleanup would be tidier.
- `test_json_200_raises_exception` asserts "error" is in the exception message. The eventual fix must word its message so that holds (the existing text/* message "Provider returned an error page" already would).

No em dashes, real path, genuinely red on main. Good regression.

razzamatazm:
Fix is up as PR #365. The two tests in this PR are red on main and green on the fix branch. Merge order: either merge #356 first (showing tests fail) then #365, or merge #365 directly since it includes the test file.

razzamatazm:
Superseded: #365 merged with this regression test bundled (backend/tests/test_download_json_body.py).

PR #364, opened 2026-06-09
docs: remove em dashes from new changelog entries

razzamatazm:
## What a user would notice

No visible change for operators. This is a style fix only.

## What changed

PRs #351 through #362 each added their own changelog entries directly, but those entries used em dashes in several places. The project style rule prohibits em dashes everywhere. This PR replaces them with commas, semicolons, colons, or parentheses as appropriate, across the 12 new entries added since the last Scribe run.

## How it was tested

Grep confirms no em dashes remain in the new entries (lines 1-80 of CHANGELOG.md). Existing older entries were not touched.


razzamatazm-cleo (changes requested):
Docs-only, zero runtime risk, and the 12 edits made are correct: each em dash was replaced with a comma, colon, or parenthetical, with no change to meaning. Good.

One blocker: the cleanup is incomplete. CHANGELOG.md still contains 3 em dashes after this change, at lines 91, 101, and 105 (verified: `grep -c '—' CHANGELOG.md` returns 3). The repo rule against em dashes is absolute and applies to all documentation, so the file is not compliant until those are gone too. Since it is the same file and the same cleanup, please remove those three rather than leaving them for a follow-up PR. Then I will approve.

razzamatazm-cleo (approved):
Approve.

Verified at head ab4dc19:
- The prior blocker is fixed. origin/main CHANGELOG has 15 em dashes; after this branch the file has 0 (grep -c confirmed). The three entries I previously flagged (Commercial skip, First-run setup, Downloads survive hiccups) are now de-dashed too.
- Docs-only, zero runtime risk. Each em dash is replaced with a comma, colon, or parenthetical and meaning is preserved throughout.

Coordination note (not a defect in this PR): PR #369 also strips em dashes from these same older entries (12 of them) while adding three new entries on top. The two branches edit identical lines and WILL conflict on merge. This PR is the correct home for the cleanup and is complete, so the clean path is: merge #364 first, then have #369 rebase and drop its em-dash edits, keeping only its three new entries.

PR #365, opened 2026-06-09
fix: reject application/json error bodies in content-type guard

razzamatazm:
## What changed

Some IPTV providers return JSON error messages (like "maximum connections reached" or "auth failed") with HTTP 200 instead of an error status code. The download manager already rejects responses with `Content-Type: text/*` (HTML/plain error pages), but `application/json` does not start with `text/`, so it slipped through.

**Before:** Provider returns HTTP 200 with `Content-Type: application/json` body `{"error": "Stream unavailable"}`. The JSON body (~40 bytes) is written to the `.ts` file. `downloaded_bytes` is non-zero so the empty-body guard also misses. Download is marked COMPLETED. User ends up with a corrupt, non-playable recording in Plex/Jellyfin (shows as 0:00 duration).

**After:** Both content-type guards (initial request and the restart path) now also reject `application/json`. Download is immediately marked FAILED with the message "Provider returned an error page (Content-Type: application/json)."

## Files changed

- `backend/services/download_manager.py`: extend content-type guard at both check points (initial request path and needs_restart path)
- `backend/tests/test_download_json_body.py`: regression tests covering a plain JSON error body and a session-limit error with `; charset=utf-8` suffix

## Test run

Before fix:
```
FAIL: test_json_200_raises_exception - Exception not raised
FAIL: test_json_200_session_limit_raises_exception - Exception not raised
```

After fix:
```
Ran 2 tests in 0.005s
OK

Ran 996 tests in 8.808s
OK
```

## Related

- Regression test PR: #356 (approved, tests were confirmed red on main)
- Task thread: 1514041003672076308

### DIFF
--- a/backend/services/download_manager.py
+++ b/backend/services/download_manager.py
@@ -1747,7 +1747,10 @@ class DownloadManager:
                     raise Exception(f"HTTP {response.status}: {response.reason}")
 
                 content_type = response.headers.get("Content-Type", "")
-                if not needs_restart and content_type.lower().startswith("text/"):
+                _ct_lower = content_type.lower()
+                if not needs_restart and (
+                    _ct_lower.startswith("text/") or _ct_lower.startswith("application/json")
+                ):
                     raise Exception(
                         f"Provider returned an error page (Content-Type: {content_type}). "
                         "The catchup window may have expired or be unavailable."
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve.

What this does for a user: a recording that the provider answered with a JSON error message (HTTP 200 plus Content-Type application/json) now fails loudly with a clear message instead of silently saving a tiny, unplayable file that Plex or Jellyfin chokes on.

Review notes:
- Correctness: both content-type guard sites are patched, the initial request (line ~1750) and the restart/byte-0 re-request path (line ~1850). I grepped download_manager.py against origin/main and these are the only two guard sites, so there is no third unguarded path. startswith("application/json") also covers "application/json; charset=utf-8".
- Tests: verified the regression. With the fix reverted to the main version of download_manager.py the two tests FAIL (Exception not raised); with the fix in place they pass. Full backend suite is green: Ran 996 tests, OK.
- Conventions: minimal diff, matches file style, no em dashes.

No remaining concerns.

PR #366, opened 2026-06-09
Add frontend test suite (Vitest + Testing Library) and run it in CI

razzamatazm:
## Summary

The frontend previously had no test suite (verified only via `npm run build`). This adds one:

- **Tooling**: Vitest 4 + React Testing Library + jsdom, configured in `frontend/vitest.config.js` (separate file; `vite.config.js` untouched). `src/test/setup.js` registers jest-dom matchers, RTL cleanup, and the browser-API stubs Mantine needs under jsdom (`matchMedia`, `ResizeObserver`, `scrollIntoView`). `src/test/renderWithProviders.jsx` provides Mantine / React Query render wrappers.
- **46 tests, ~1.7s**:
  - `utils/channelTime.test.js` — timestamp vs ISO precedence, guide-offset math, Today/Tomorrow/Yesterday/weekday air-date labels (fixed system clock)
  - `api.test.js` — request layer: CSRF token fetch/caching, refresh-and-retry on CSRF 403, no retry on other 403s, error status/payload/detail flattening, failed-CSRF promise not cached, query encoding
  - `components/ProgressBar.test.jsx` — progress clamping, indeterminate mode
  - `components/ProgramCard.test.jsx` — past/current/downloadable states, click gating on `has_archive`, guide-offset time display
  - `components/settings/FoldersStatus.test.jsx` — writable/not-writable/missing badges, error alert
- **CI**: frontend steps added to the existing `tests` job in `.github/workflows/tests.yml`, so the required status check now covers both suites.
- New npm scripts: `npm test` (single run), `npm run test:watch`.

No production code changed. `npm run build` still passes.

## Steps to test

```bash
cd frontend
npm install
npm test          # 46 passing
npm run build     # unaffected
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,3 +21,18 @@ jobs:
       - name: Run backend tests
         run: python -m unittest discover -s tests
         working-directory: backend
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+          cache: npm
+          cache-dependency-path: frontend/package-lock.json
+
+      - name: Install frontend dependencies
+        run: npm ci
+        working-directory: frontend
+
+      - name: Run frontend tests
+        run: npm test
+        working-directory: frontend
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -20,11 +20,74 @@
         "react-router-dom": "^6.21.3"
       },
       "devDependencies": {
+        "@testing-library/jest-dom": "^6.9.1",
+        "@testing-library/react": "^16.3.2",
+        "@testing-library/user-event": "^14.6.1",
         "@types/react": "^18.2.48",
         "@types/react-dom": "^18.2.18",
         "@vitejs/plugin-react": "^4.2.1",
-        "vite": "^5.0.12"
+        "jsdom": "^29.1.1",
+        "vite": "^5.0.12",
+        "vitest": "^4.1.8"
       }
+    },
+    "node_modules/@adobe/css-tools": {
+      "version": "4.5.0",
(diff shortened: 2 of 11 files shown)


razzamatazm-cleo (changes requested):
The test code is excellent and I want this in. Holding approval on one process gate plus one robustness note.

What this does for a user: nothing visible at runtime. It adds an automated safety net for the web UI so future changes that break the front end get caught in CI before they ship.

What I verified:
- npm ci then npm test on this branch: Test Files 5 passed, Tests 46 passed. The tests assert real behavior (CSRF token caching and 403 retry in api.js, timestamp and offset handling in channelTime.js, component rendering), not trivial smoke checks. Setup file correctly stubs matchMedia, ResizeObserver, scrollIntoView and registers RTL cleanup since globals is false.

Why I am not approving yet:
1. This PR edits .github/workflows/tests.yml, which is a sensitive path. Per CLAUDE.shared.md, workflow files require an operator approval in #tasks before merge. The change itself is benign (setup-node 20, npm ci, npm test on the frontend), but I cannot be the gate that lets a workflow edit through without that sign-off. Please get Tyler's OK in #tasks, or split the workflow step into its own PR he approves directly.
2. Robustness, non-blocking: several channelTime assertions compare locally-formatted output (for example formatChannelDateTime returning 'Jun 1, 2026 12:05 PM' from a Z timestamp, and the formatAirDateTime cases). These pass because GitHub runners and this environment are UTC, but they will fail for a contributor running the suite in a non-UTC timezone. Consider pinning TZ=UTC in the test script or vitest config so local runs match CI.

Resolve item 1 (operator approval on the workflow change) and I will approve.

PR #367, opened 2026-06-09
fix: reject application/xml error bodies in content-type guard

razzamatazm:
## What this tests

Some Xtream Codes providers return XML error messages with HTTP 200 and `Content-Type: application/xml`, e.g.:

```
HTTP/1.1 200 OK
Content-Type: application/xml
<error><message>Stream unavailable</message><code>403</code></error>
```

The existing guard in `_download_file` (lines 1667 and 1767 of `download_manager.py`) rejects `text/*` bodies. PR #365 extends this to also reject `application/json`. Neither guard covers `application/xml`, so the XML body is written to the output `.ts` file and the download is silently marked COMPLETED with a corrupt, non-playable recording.

## Tests added

`backend/tests/test_download_xml_body.py` (2 tests):

- `test_xml_200_raises_exception`: `application/xml` plain error body
- `test_xml_200_with_charset_raises_exception`: `application/xml; charset=utf-8` (with charset param)

Both tests currently **FAIL** (the bug is present). They will pass once the content-type guard in `_download_file` is extended to also reject `application/xml`.

## Relation to open work

Companion to PR #356 (`application/json` red test) and PR #365 (`application/json` fix). The fix should extend both guards at lines 1667 and 1767 to also reject `application/xml`.


razzamatazm-cleo (changes requested):
Request changes: scope contamination.

The XML regression itself is good and documents a real gap. I confirmed test_download_xml_body.py genuinely fails against current code (application/xml is not caught by the text/ guard, nor by the application/json guard added in PR #365), so a provider returning an XML error body with HTTP 200 would still write garbage to the .ts file and mark it COMPLETED. Worth fixing.

The problem is this PR also adds test_download_json_body.py, which is byte-for-byte identical to the file already introduced by PR #356 and PR #365 (I diffed them). That is out of scope for an XML PR and will collide on merge: whichever of #356/#365 lands first creates that exact new file, and this branch then re-creates the same path. Please drop test_download_json_body.py from this branch and keep only test_download_xml_body.py.

Two more things, not blocking the above:
- Strongly prefer pairing this with the one-line fix (extend the guard to reject application/xml, or better, treat any non-stream content type as an error) rather than leaving a red test in the suite. The fix is the same shape as #365.
- Minor: the json-body docstring in the xml file references PR #365 by number, which is fine, but make sure the surviving file's module docstring still reads correctly once the json file is removed.

Once the json duplicate is gone, this is a clean failing-regression consistent with #356.

razzamatazm-cleo (approved):
Approve.

Verified at head aadbf92:
- The prior blocker is resolved: the duplicate test_download_json_body.py is gone; this branch now carries only test_download_xml_body.py plus the matching one-line fix.
- The guard is extended in BOTH request paths (first request and the byte-0 restart re-request), so an application/xml error body with HTTP 200 can no longer slip through on either path. That matches the known restart-path-drops-guard failure mode.
- Ran the new test on this head: both cases pass. Ran the full backend suite: 883 tests pass, no regressions.

Non-blocking: the guard also adds the application/json line that PR #365 already introduces. Since #365 is approved but not yet merged, this is harmless duplication but will need a trivial rebase once #365 lands. The assertIn('error') check is a little loose (the exception message happens to contain 'error'), but it does exercise the real guard, so it is fine to keep.

PR #372, opened 2026-06-10
fix: ISO XMLTV timestamps with named TZ abbreviations silently drop programs

razzamatazm:
## What this PR contains

11 failing tests that prove a new bug in `_parse_xmltv_time`. No code changes.

## Bug

`_parse_xmltv_time` has named-TZ support (`_NAMED_TZ_OFFSETS`) only in the compact-format code path (`20240115200000 EST`). When a timestamp looks like ISO format (date part contains dashes or T), the code assembles an `iso_str` and calls `datetime.fromisoformat()`, which raises `ValueError` on named TZ suffixes. The `except` clause returns `None`.

Result: every `<programme>` element whose `start` or `stop` attribute uses an ISO-format date with a named TZ abbreviation is silently dropped from the EPG. The guide goes partially or fully empty with no error or warning.

**Formats that fail:**
- `"2024-01-15 20:00:00 EST"` (3-part space-separated, named TZ)
- `"2024-01-15 20:00:00 CET"`
- `"2024-01-15 20:00:00 ACST"`
- `"2024-01-15 20:00:00 MSK"`
- `"2024-01-15T20:00:00 EST"` (T-separator, space before named TZ)

**How to reproduce:**
1. Build an XMLTV document with `start="2024-01-15 20:00:00 EST"`.
2. Run EPG ingest.
3. All programmes are silently skipped. Guide is empty.

## Fix outline (maintainer)

In the ISO path, when `len(parts) == 3` and `parts[2]` is letters-only (not a numeric offset), look it up in `_NAMED_TZ_OFFSETS` and construct the timezone manually, the same way the compact path does. Same guard needed for the 2-part `T`-separator case where `parts[1]` is a letters-only string.

## Tests

`backend/tests/test_epg_xmltv_iso_named_tz.py` - 11 assertions fail before the fix, 3 regression guards pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-09
 
+### Fixed: Guides using ISO timestamps with timezone names (EST, CET, MSK) no longer come up empty
+
+**What you would notice:** With some providers, certain channels (or the whole guide) showed no programs at all, with no error anywhere. This happened when the provider's guide file wrote program times in ISO format with a named timezone, for example "2024-01-15 20:00:00 EST" instead of a numeric offset. Mustarrd could not parse those times and silently skipped every affected program. These guides now load correctly, with program times converted from the named timezone.
+
+**What changed:** `_parse_xmltv_time` resolved named timezone abbreviations only in the compact XMLTV format path; the ISO format path fed the abbreviation to `datetime.fromisoformat`, which rejected it, and the parser returned None. The ISO path now strips a trailing named timezone, looks it up in the existing offset table, and applies it to the parsed time. Unknown abbreviations fall back to UTC, the same behavior as the compact path. 14 regression tests cover the affected formats.
+
+---
+
 ### Added: GPU-accelerated encoding in Docker, with a plain-language GPU status panel in Settings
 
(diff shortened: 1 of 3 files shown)


razzamatazm-cleo (approved):
Approve as a failing regression.

Verified at head: I ran this test file against origin/main and 11 of 14 cases fail (the 3 that pass are the deliberate 'still works' guards: numeric offset, no-tz, compact named-TZ). So it documents a real, unfixed bug: _parse_xmltv_time applies _NAMED_TZ_OFFSETS only in the compact path; the ISO path (epg_ingest_manager.py:836+) calls datetime.fromisoformat() on a string like '2024-01-15T20:00:00EST', which raises ValueError, and the except returns None, so every programme with an ISO date plus a named TZ is silently dropped. The UTC expectations (EST=-5, PST=-8, CET=+1, ACST=+9:30, MSK=+3) are all correct, and it calls the real method with no mocking. Consistent with the project's failing-regression convention.

Non-blocking nit for whoever writes the fix: the test_iso_t_separator_est_not_none docstring claims that case returns a wrong-but-non-None UTC value (5-hour shift); in fact it also returns None on current main, so the assertIsNotNone is right but the explanatory text is slightly off. Worth a one-line docstring correction when the fix lands.

razzamatazm:
Added the fix on top of the failing tests (same pattern as #367): the ISO path in `_parse_xmltv_time` now strips a trailing named-TZ token, resolves it via `_NAMED_TZ_OFFSETS`, and applies it to the parsed naive timestamp; unknown abbreviations fall back to UTC to match the compact path. Verified the 11 tests fail pre-fix and all 1014 pass post-fix. CHANGELOG entry included. Rebased onto main.

PR #374, opened 2026-06-10
docs: changelog entries for PRs #365 and #368

razzamatazm:
## What changed

Adds plain-English changelog entries for two merged PRs that had no user-facing documentation:

- **#365** (fix: reject application/json error bodies in content-type guard): some providers return JSON error messages disguised as HTTP 200 responses; Mustarrd now marks those downloads as Failed instead of silently writing a corrupt file.
- **#368** (design: add Settings link to Browse right panel when provider unreachable): the empty guide panel now shows an actionable link to Settings, Accounts for admins, or "Contact your administrator." for other users.

Docs-only change. No code touched.

## Before and after

Before: PRs #365 and #368 had no CHANGELOG entries.

After: Both are documented in plain language for a non-technical Unraid operator, under the existing 2026-06-09 section.


razzamatazm-cleo (approved):
Verified against merged code. #365 changelog claim accurate: content-type guard now rejects application/json at BOTH the initial request (line ~1747) and the restart path (line ~1850), and two regression tests cover the 200-with-JSON cases. #368 claim accurate: admin sees a Link to /settings?section=accounts, non-admin sees "Contact your administrator." Entries are correctly placed under the existing 2026-06-09 section, no stray date heading. No em dashes. Docs-only, no code touched.

Non-blocking nit: the actual UI link reads "Check Settings → Accounts." while the changelog paraphrases it as "Check Settings, Accounts." Fine as plain-English prose; flagging only for accuracy.

PR #375, opened 2026-06-10
design: keep Export and Import buttons together on mobile

razzamatazm:
## What changed

On narrow phone widths (390 px), the Export and Import buttons on the Scheduled Recordings page wrapped to separate lines. Import ended up alone on a second line while Export stayed next to the Auto-retry toggle on the first, which looked like a layout glitch and made the two buttons feel unrelated.

The fix is one JSX change: Export and Import are now wrapped in a nowrap Group so they always appear as a pair. When the header is too wide for one row on mobile, the toggle drops to its own line and the two buttons follow together on the next line.

Desktop layout is unchanged.

## Before (390 px mobile)

Import button stranded alone on the second line (screenshots in the design channel audit thread).

## After (390 px mobile)

Export and Import sit together on their own row (screenshots in the design channel audit thread).

## Test plan

- [ ] Open Scheduled Recordings at 390 px width: Export and Import are on the same row
- [ ] Open at 1280 px: toggle, Export, and Import are still all on one row (unchanged)
- [ ] Export and Import still work (file download / file picker)

### DIFF
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -558,23 +558,25 @@ export default function Scheduled() {
               />
             </Tooltip>
           )}
-          <Button
-            size="xs"
-            variant="default"
-            leftSection={<IconFileExport size={14} />}
-            onClick={handleExport}
-          >
-            Export
-          </Button>
-          <Button
-            size="xs"
-            variant="default"
-            leftSection={<IconFileImport size={14} />}
-            onClick={() => importInputRef.current?.click()}
-            loading={importMutation.isPending}
-          >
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Verified. Group is imported from @mantine/core (line 5). The new nowrap Group wrapping Export and Import sits inside the existing header Group, whose parent uses wrap="wrap", so on narrow widths the toggle drops to its own line and the two buttons follow together as a pair. Desktop (single-row) layout unchanged. Minimal one-block diff, scope clean, no logic touched. No regression test required: UI-only layout change and the repo has no frontend test suite.

PR #376, opened 2026-06-10
fix: cap series download episode list at 200 per request

razzamatazm:
## What you would notice

Before this fix, if you selected hundreds of TV episodes and clicked Download, the server would silently accept the full list and run a database query and download task for every single episode, back to back. On a Raspberry Pi or a low-spec Unraid box, a 500-episode request could stall the app for several seconds, blocking all other users from downloading or browsing.

## What changed

`POST /api/vod/series/download` now rejects any request with more than 200 episodes at the validation stage, before any database work begins. The response is HTTP 422 with a validation error. Requests with 200 or fewer episodes are unchanged.

The limit is enforced by adding `Field(max_length=200)` to the `episodes` field in `SeriesDownloadRequest` (one line in `backend/api/vod.py`).

## How it was tested

Three new regression tests in `backend/tests/test_series_episode_limit.py`:
- A 201-episode request raises `ValidationError` (was silently accepted before)
- A 200-episode request is accepted
- An empty episode list is accepted

```
Ran 1017 tests in 8.644s
OK
```

Before fix: `SeriesDownloadRequest(episodes=[...201 items...])` succeeds.
After fix: `pydantic_core._pydantic_core.ValidationError: 1 validation error for SeriesDownloadRequest — episodes: List should have at most 200 items after validation, not 201`

### DIFF
--- a/backend/api/vod.py
+++ b/backend/api/vod.py
@@ -1,9 +1,9 @@
 from collections import Counter
 from fastapi import APIRouter, Depends, HTTPException
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy import select
-from typing import Optional
+from typing import Annotated, Optional
 import logging
 import os
 
@@ -44,7 +44,7 @@ class SeriesDownloadRequest(BaseModel):
     account_id: int
     series_id: str
     series_name: str
-    episodes: list[EpisodeItem]
+    episodes: Annotated[list[EpisodeItem], Field(max_length=200)]
 
 
 async def _get_client(session: AsyncSession, account: XtreamAccount) -> XtreamClient:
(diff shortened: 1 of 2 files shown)


razzamatazm-cleo (approved):
Approve. Verified at head d07a988.

- Correctness: `Field(max_length=200)` on `episodes` rejects oversized requests at the validation stage, before any DB work. Confirmed Pydantic v2.10+ (per requirements.txt) enforces max_length on a list. Single endpoint, single model, no bypass path.
- Tests: boundary cases (200 accepted, 201 rejected) plus empty list. Valid regression: the 201 case fails before the fix (no bound) and passes after.
- Edge cases: empty list handled by the existing `if not downloads: return {count: 0}` path.
- Security: reduces the DoS surface on low-spec hosts; endpoint stays gated by require_admin_or_download_user.
- Conventions: minimal diff, plain-English body with before/after evidence, no em dashes.

Non-blocking nit: the PR body says 'two database queries for every single episode', but the conflict check is already batched into one IN query; the real per-episode cost is build_episode_download. The cap is still the right safety limit, so this is just a wording note, not a change request.

PR #378, opened 2026-06-10
design: show countdown to download start on upcoming recording cards

razzamatazm:
## What changed

The Downloads > Upcoming tab now shows how long until a recording's download begins, alongside the existing absolute time.

**Before:** Download starts: Today at 9:30 PM
**After:** Download starts: Today at 9:30 PM (in 20h 1m)

If padding is configured, both notes appear together: (in 20h 1m, 1h 45m with padding).

Once the show has finished airing and the download begins, the countdown disappears automatically.

## Why

A non-technical user seeing "Download starts: Today at 9:30 PM" at 1:30 AM has to do mental arithmetic to realize it is 20 hours away. The countdown makes the wait immediately obvious without requiring any math.

## What changed in code

- Added `formatTimeUntil()` helper in `Downloads.jsx` that takes the guide-offset-adjusted dayjs instant and returns a string like "in 2h 15m", "in 1d 3h", etc., or null once the time has passed.
- Imported `getChannelDisplayTime` and `getNowUtc` from `channelTime` (both already exported, no changes to that file).
- Upcoming card render now combines the countdown and any padding note into a single parenthetical.

Frontend only. No backend changes.

## Screenshots

Screenshots posted in #design. Summary:

**Before** (desktop and mobile): Download starts: Today at 9:30 PM

**After** (desktop and mobile): Download starts: Today at 9:30 PM (in 20h 1m)

## Test plan

- Open Downloads > Upcoming with a scheduled recording in the future: countdown appears.
- Once the recording available time is in the past: countdown disappears (returns null).
- Recordings with pre/post padding: both countdown and padding note appear, e.g. "(in 20h 1m, 1h 45m with padding)".
- Recordings with no padding: only the countdown appears.

### DIFF
--- a/frontend/src/pages/Downloads.jsx
+++ b/frontend/src/pages/Downloads.jsx
@@ -44,7 +44,7 @@ import duration from 'dayjs/plugin/duration'
 
 import { accountsApi, authApi, downloadsApi, createDownloadWebSocket } from '../api'
 import ProgressBar from '../components/ProgressBar'
-import { formatChannelDateTime, formatAirDateTime, getGuideOffsetHours } from '../utils/channelTime'
+import { formatChannelDateTime, formatAirDateTime, getGuideOffsetHours, getChannelDisplayTime, getNowUtc } from '../utils/channelTime'
 
 dayjs.extend(duration)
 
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approve. Frontend only, single file (Downloads.jsx).

Correctness verified against frontend/src/utils/channelTime.js:
- Offset handling is sound. getChannelDisplayTime returns instant+offset; formatTimeUntil compares against getNowUtc()+offset, so the offset cancels in the diff (available - now). This mirrors the existing formatAirDateTime pattern, so the absolute time and the countdown stay in sync.
- All three new imports (getChannelDisplayTime, getNowUtc, getGuideOffsetHours) are exported there, so the import line is valid.
- Edge cases handled: null available time -> null countdown + 'Unknown'; time in the past -> diffMinutes <= 0 -> null, so the countdown disappears once the download starts, as the PR claims.
- Padding-note string and its condition are byte-identical to the prior code; combining into downloadNotes.join(', ') is a clean no-regression refactor.

Minor, non-blocking:
- hours >= 48 collapses to 'in Nd' and drops the remaining hours (e.g. 50h -> 'in 2d'). Acceptable for a glanceable countdown.
- No regression test, but there is no frontend test harness in the repo yet and this is a pure presentational helper, so nothing to block on.

Conventions: minimal diff, 2-space indent, camelCase helper, UI-only scope, no em dashes. No remaining concerns.

PR #380, opened 2026-06-10
design: downgrade GPU badge to warning when ffmpeg is unavailable

razzamatazm:
## What a user would notice

On the Post-Processing settings page, when ffmpeg is not installed, the GPU status badge used to show a green checkmark labeled **"GPU ENCODING READY"**. This is misleading: GPU encoding cannot work without ffmpeg. A non-technical user would see green and assume everything is fine, then be confused when their recordings are not converted.

After this change, the badge turns yellow and reads **"GPU DETECTED, FFMPEG REQUIRED"** with the message "A GPU was found but ffmpeg is not installed. Install ffmpeg to use GPU encoding." The green badge only appears when both GPU and ffmpeg are present.

## What changed

One change in `frontend/src/pages/Settings.jsx`: when the GPU status would be green but `ffmpegReady === false`, the badge color, icon, label, and detail are overridden to show a yellow warning instead.

## Test plan

- Confirmed visually in a local dev environment where ffmpeg is not installed. Before: green badge. After: yellow badge.
- When ffmpeg IS installed, the green badge still appears as before (the condition only triggers when `ffmpegReady === false`).
- Frontend only. No backend changes. No logic changes.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -917,12 +917,19 @@ export default function Settings() {
             {(() => {
               const hwStatus = describeHardwareAccel(toolsStatus.vaapi)
               if (!hwStatus) return null
+              const gpuBlockedByFfmpeg = hwStatus.color === 'green' && ffmpegReady === false
+              const badgeColor = gpuBlockedByFfmpeg ? 'yellow' : hwStatus.color
+              const badgeIcon = gpuBlockedByFfmpeg ? <IconAlertCircle size={12} /> : hwStatus.icon
+              const badgeLabel = gpuBlockedByFfmpeg ? 'GPU detected, ffmpeg required' : hwStatus.label
+              const badgeDetail = gpuBlockedByFfmpeg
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Reviewed at head c9da381. Frontend-only design change, Settings.jsx.

Logic verified:
- gpuBlockedByFfmpeg = hwStatus.color === 'green' && ffmpegReady === false. Only downgrades the badge when a GPU is detected (green) AND ffmpeg is confirmed missing.
- Uses strict === false, so while toolsStatus is still loading (ffmpegReady undefined) the badge is not falsely downgraded. Matches the existing pattern at line 983.
- ffmpegReady is defined in renderProcessing (line 883); the badge IIFE is nested inside, so it is in scope. The whole block is gated on toolsStatus being truthy.
- IconAlertCircle is imported (line 33) and already used elsewhere in the file.
- describeHardwareAccel returns color 'green' only for a ready GPU, so the override cannot fire on a red/unavailable state.

Minimal diff, no backend or logic changes, no em dashes in the new copy. No frontend test framework exists yet, so no test is required. Approving.

PR #381, opened 2026-06-10
docs: changelog entry for PR #378 (download countdown)

razzamatazm:
## What changed

Adds a changelog entry for PR #378, which added a countdown timer to Upcoming recording cards in the Downloads page.

## For an Unraid user

Users will now see how long until a scheduled recording starts, without having to do the math themselves.

## Test plan

- Verified no em dashes in the entry.
- Entry is plain English, written for a non-technical operator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ## 2026-06-10
 
+### Improved: Downloads page now shows a countdown to when a scheduled recording will start
+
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approved. Docs-only changelog entry for merged PR #378.

Verified every claim against #378's code (frontend/src/pages/Downloads.jsx):
- "(in 20h 1m)" matches formatTimeUntil output and is appended next to the Download starts time.
- Combined padding example "(in 20h 1m, 1h 45m with padding)" matches downloadNotes.join(', '); paddingNote uses formatDuration so there is correctly no "in" prefix on that part.
- "disappears once the recording is underway" matches formatTimeUntil returning null when diffMinutes <= 0.
- "Frontend only, no server changes" matches: #378 touched only Downloads.jsx.

No em dashes. Plain English for a non-technical operator. New entry stacks above the existing 2026-06-10 entry with no duplicate date header. No tests required for a changelog-only doc change.

PR #383, opened 2026-06-10
design: add countdown to download start time on Scheduled Recordings cards

razzamatazm:
## What changed

Scheduled Recordings > Upcoming cards now show how long until the download starts, matching what Downloads > Upcoming already shows after PR #378.

**Before:** "Download starts: Today at 9:30 PM"
**After:** "Download starts: Today at 9:30 PM (in 18h 39m)"

## Why

A non-technical user who schedules a show from the Scheduled page had to do mental arithmetic to know how long they would wait. PR #378 fixed this for Downloads > Upcoming but the Scheduled page was not updated at the same time.

## What changed in the code

One file: `frontend/src/pages/Scheduled.jsx`.

- Added `getChannelDisplayTime` and `getNowUtc` to the import from `channelTime` (already exported, already used in Downloads.jsx).
- Added `formatTimeUntil` function (identical logic to the one in Downloads.jsx).
- In `ScheduleCard`, computed `downloadDisplayTime`, `timeUntil`, and `downloadNotes` alongside the existing variables.
- Updated the "Download starts" line to append the countdown when notes are present.

The countdown only appears for active schedules (scheduled, queued, downloading, etc.). Completed, failed, and cancelled cards are unchanged.

## How it was tested

- Started backend and frontend locally.
- Playwright screenshots at 1280px desktop and 390px mobile: before and after both captured.
- Countdown appears correctly on both widths.
- No changes to backend, API, or other UI components.

## Screenshots

Before and after posted in #design (Pixel iteration 52).

### DIFF
--- a/frontend/src/pages/Scheduled.jsx
+++ b/frontend/src/pages/Scheduled.jsx
@@ -41,7 +41,7 @@ import dayjs from 'dayjs'
 import duration from 'dayjs/plugin/duration'
 
 import { accountsApi, authApi, downloadsApi, schedulesApi, settingsApi } from '../api'
-import { formatChannelDateTime, formatAirDateTime, getGuideOffsetHours } from '../utils/channelTime'
+import { formatChannelDateTime, formatAirDateTime, getGuideOffsetHours, getChannelDisplayTime, getNowUtc } from '../utils/channelTime'
 
 dayjs.extend(duration)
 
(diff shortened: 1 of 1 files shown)


razzamatazm-cleo (approved):
Approved at head 649f7fe.

Verified against origin/main:
- formatTimeUntil is byte-for-byte identical to the already-merged Downloads.jsx version (PR #378); same getChannelDisplayTime('available', ...) usage. Guide offset is applied to both displayTime and nowDisplay, so it cancels in the diff: countdown is offset-independent, correct.
- Display line: completed/failed/cancelled cards unchanged (timeUntil is gated on isActive, null otherwise). Padding note preserved; both notes join cleanly as '(in 18h 39m, 1h 30m with padding)' with no double parens.
- All new imports (getChannelDisplayTime, getNowUtc) are exported and used. Minimal diff, one file, no em dashes.

Non-blocking: formatTimeUntil is now duplicated in Downloads.jsx and Scheduled.jsx. A future cleanup could lift it into utils/channelTime.js (which already has a test file) and add a unit test for the 48h / 24h / sub-24h branches. Not required for this design PR since it matches the established merged pattern exactly.

PR #386, opened 2026-06-10
fix: snapshot schedule attributes before dispatch loop to prevent sibling FAILED

razzamatazm:
## What this fixes

On Unraid, if two recordings are scheduled to start within the same 30-second window and the first one hits an error (for example, the provider rejects it), the second recording would silently fail with a confusing internal error. In the Scheduled Recordings history, users would see a message like:

> greenlet_spawn has not been called; can't call await_only() here.

This is a SQLAlchemy internal error with no meaning to the operator. The second recording stays stuck as FAILED forever, even though nothing was wrong with it.

## Root cause

When a dispatch fails, the scheduler calls `session.rollback()` to undo the staging of the failed download. SQLAlchemy's rollback expires all objects in the session's identity map, including the *next* schedule in the same batch. When the loop then reads that schedule's attributes (`program_title`, `channel_id`, etc.), the async session cannot lazily reload them and raises `MissingGreenlet`. The generic `except Exception` handler catches it and marks the schedule FAILED.

## Fix

All schedule attributes that the dispatch block needs are now read into a plain Python dict before the loop starts, while the session is clean. The local copies survive any rollback, so a failure on schedule A cannot affect schedule B's data.

`session.refresh(schedule)` inside the loop still re-reads the ORM object from the DB before committing, so the QUEUED update always reflects the current DB state.

## How it was tested

**Before the fix:** `test_dispatch_rollback_expiry.py` fails with `AssertionError: 'failed' != 'queued'`, and schedule B's `status_message` contains the `MissingGreenlet` error text.

**After the fix:** The same test passes. Schedule A is FAILED (its queue raised), schedule B is QUEUED (dispatched normally).

The test uses a real SQLAlchemy async session with SQLite in-memory. It makes `queue_download` execute a `SELECT 1` before raising, which starts a real active transaction. The subsequent rollback genuinely expires B's attributes, reproducing the exact failure mode.

Full backend test suite: 1022 tests pass.

## Steps to verify

1. Apply this PR locally.
2. Run `cd backend && python -m unittest tests.test_dispatch_rollback_expiry -v` -- should show OK.
3. Run the full suite: `python -m unittest discover -s tests` -- all 1022 tests pass.

There is no easy way to reproduce the original bug manually because it requires two schedules firing in the exact same 30-second tick, the first one failing with a DB write in flight.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ All notable changes to Mustarrd are listed here. Most recent changes are at the 
 
 ---
 
+### Fixed: Scheduling two recordings at once no longer breaks the second one when the first fails
+
+**What you would notice:** On Unraid, if two scheduled recordings were due to start in the same 30-second window and something went wrong with the first one (such as a provider rejecting it), the second recording would silently fail with a confusing internal error ("greenlet_spawn has not been called") and stay stuck as FAILED. The second recording is now unaffected by the first's failure and will be queued normally.
+
+**What changed:** Backend only. When a scheduled recording fails to start, the scheduler now saves a copy of each recording's details before processing begins. This means a failure in one recording can no longer corrupt the data needed to start the next one.
+
+---
+
 ### Improved: Downloads page now shows a countdown to when a scheduled recording will start
 
 **What you would notice:** On the Downloads page, each card in the Upcoming tab now shows how long until the recording begins, next to the scheduled start time. For example, you might see "Download starts: Today at 9:30 PM (in 20h 1m)". Previously you had to do the math yourself. If you have pre or post-recording padding set, the countdown and the padding note appear together: "(in 20h 1m, 1h 45m with padding)". The countdown disappears automatically once the recording is underway.
--- a/backend/services/scheduled_manager.py
+++ b/backend/services/scheduled_manager.py
@@ -143,7 +143,43 @@ class ScheduledManager:
             # past the minimum free space threshold.
             projected_used_gb = 0.0
 
-            for schedule in ready:
+            # Snapshot all ORM attributes that the dispatch loop will read,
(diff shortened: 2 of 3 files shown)


razzamatazm-cleo (approved):
Approve. Verified fix and test locally.

**Correctness:** The dispatch loop now reads every per-schedule value from a pre-loop snapshot dict instead of the live ORM object. This is the right fix: a prior iteration's `session.rollback()` (in the except handler) expires every object in the session identity map, and the only remaining accesses to the live `schedule` object in a later iteration are writes (status/message/updated_at, which do not trigger a lazy load), an explicit `await session.refresh(schedule)`, and `_estimate_recording_gb` on the post-refresh success path. No expired-attribute read survives. Snapshot key coverage checked against every `snap[...]` access and `build_download_from_program` arg: complete.

**Test:** Genuine regression test. I ran it against the branch (passes) and confirmed it fails with `MissingGreenlet` when the loop is reverted to read the live (expired) ORM object, so it fails before the fix and passes after. The reproduction is real, not mocked: queue_download runs an actual SELECT then raises, the rollback and attribute expiry are real SQLAlchemy behavior, and `expire_on_commit=False` matches production so only rollback (not commit) expires.

**Security / conventions:** Backend only, no untrusted input touched, plain-English changelog with what-you-would-notice framing, no em dashes. Scope clean.

Minor (non-blocking): `_estimate_recording_gb(schedule)` and the low-space/missing-folder branches still touch the live object, but all are on paths where it is freshly refreshed or only written, so no bug. Could pass snap there for uniformity later; not required.

ghost:
Merge conflict. Main has moved (PRs #383 and others merged since you branched). Please rebase on main and force-push. CI is green and Cleo has approved; this is the only blocker.

PR #390, opened 2026-06-10
Add in-browser playback for recordings and previews in all major browsers

razzamatazm:
## Summary

Playing a finished recording or previewing a guide program previously showed a black box in most browsers: raw `.ts` files and the MPEG-TS preview relay play natively **nowhere**, and `.mkv` only plays in Chrome. This PR makes both playback spots work in Firefox, Chrome, Edge, and Safari.

**Phase 1 — client-side TS demuxing (mpegts.js)**
- Guide preview modal now demuxes the raw MPEG-TS relay in the browser via MSE (`live` mode).
- `.ts` recordings play via mpegts.js VOD mode using the existing byte-range file endpoint.
- URLs are absolutized before handing to mpegts.js — its loader runs in a worker where relative URLs fail to parse (found via live browser testing).

**Phase 2 — server-side HLS for everything else (hls.js + FFmpeg)**
- New `backend/services/hls_streamer.py`: one FFmpeg per playback session repackages the recording into fMP4 HLS in a temp dir. Video is copied when browser-decodable (H.264/HEVC w/ `hvc1` tag, otherwise transcoded to H.264), audio copied only if AAC (with `aac_adtstoasc` — ADTS AAC from TS otherwise kills the fMP4 muxer), else transcoded to AAC. Idle sessions reaped after 120s; all sessions killed on app shutdown.
- New endpoint `GET /api/downloads/{id}/hls/{asset}` with strict asset-name validation (`\A...\Z` so encoded newlines can't slip through), same auth/ownership/folder-allowlist checks as the file endpoint (extracted into a shared helper).
- Frontend `playbackEngine.js` picks per extension: mp4 → native, ts → mpegts.js, mkv/other → HLS; failures fall back native → mpegts → HLS. An expired HLS session (long pause) restarts once transparently, resuming playback position.

## Steps to test
1. Download a program that stays `.ts` (no remux) → Downloads → Play. Plays in Firefox/Zen now (engine `mpegts`, no server CPU).
2. Re-encode one to `.mkv` → Play. Plays everywhere via HLS (`/api/downloads/{id}/hls/playlist.m3u8`), seeking works.
3. Guide → program → Preview. Video renders in Firefox instead of a black box.
4. A recording with AC-3 audio: starts via HLS fallback with AAC audio.
5. Backend: `python -m unittest discover -s tests` (1082 green). Frontend: `npm test` (76 green).

Verified live in Chrome against a scratch instance with synthetic H.264+AAC TS, H.264+AC-3 TS, and MKV recordings: all three play, fallback chain and codec decisions confirmed (screenshots in `.github/pr-screenshots/web-playback-*.png`).

## Screenshots
| TS via mpegts.js | MKV via HLS |
|---|---|
| ts | mkv |

🤖 Generated with [Claude Code](https://claude.com/claude-code)


razzamatazm:
Live verification against a real Xtream provider: guide preview now plays a real catchup stream (1280x720, confirmed in two separate Chromium profiles). Found and fixed in 2958ae2: the preview video element lives in a Mantine Modal portal that mounts after open, so the attach effect ran against a null ref and never retried — preview was a dead black box in every browser. Now uses a callback-ref state; regression test added (PreviewModal.test.jsx).

preview

PR #405, opened 2026-08-11
Make VA-API render device configurable

matrix2669:
## Summary

- make the VA-API render device configurable with `CATCHUP_VAAPI_RENDER_DEVICE`
- preserve `/dev/dri/renderD128` as the default when the variable is unset or blank
- use the configured device consistently for VA-API diagnostics and all FFmpeg encoding/retry paths
- document the override in `docker-compose.yml`
- add regression coverage for selecting `/dev/dri/renderD129` via the environment

## Why

Systems with multiple DRM render nodes may expose the GPU intended for Mustarrd as something other than `renderD128`. The previous hard-coded path prevented VA-API encoding from working on those systems.

## Validation

Manually tested with `CATCHUP_VAAPI_RENDER_DEVICE=/dev/dri/renderD129` and confirmed hardware-accelerated encoding works. Automated regression coverage was added for the environment override.

### DIFF
--- a/backend/services/post_processor.py
+++ b/backend/services/post_processor.py
@@ -59,7 +59,10 @@ ENCODER_MAP = {
 class PostProcessor:
     """Handles transcoding and commercial detection/removal."""
 
-    VAAPI_RENDER_DEVICE = Path("/dev/dri/renderD128")
+    VAAPI_RENDER_DEVICE = Path(
+        (os.environ.get("CATCHUP_VAAPI_RENDER_DEVICE") or "").strip()
+        or "/dev/dri/renderD128"
+    )
     VAAPI_SYSFS_DRM_CLASS = Path("/sys/class/drm")
     # Corrupt input can make ffprobe hang forever; cap how long we wait for it.
     FFPROBE_TIMEOUT_SECONDS = 60.0
@@ -857,7 +860,7 @@ class PostProcessor:
         # Build ffmpeg command
         cmd = [self._ffmpeg_path]
         if hw_accel == HardwareAccel.VAAPI:
-            cmd.extend(["-vaapi_device", "/dev/dri/renderD128"])
+            cmd.extend(["-vaapi_device", str(self.VAAPI_RENDER_DEVICE)])
         cmd.extend([
             "-i", str(input_path),
             "-y",  # Overwrite output
@@ -935,7 +938,7 @@ class PostProcessor:
                             pass
                     retry_cmd = [self._ffmpeg_path]
                     if hw_accel == HardwareAccel.VAAPI:
-                        retry_cmd.extend(["-vaapi_device", "/dev/dri/renderD128"])
+                        retry_cmd.extend(["-vaapi_device", str(self.VAAPI_RENDER_DEVICE)])
                     retry_cmd.extend([
                         "-i", str(input_path),
                         "-y",
@@ -1231,7 +1234,7 @@ class PostProcessor:
         Remove commercials from video using EDL file.
 
         Args:
-            input_path: Path to video file
+            input_path: Path to input file
             edl_path: Path to EDL file from Comskip
             output_format: Output format
             hw_accel: Hardware acceleration method
@@ -1382,7 +1385,7 @@ class PostProcessor:
             # Concat and encode
             cmd = [self._ffmpeg_path]
             if hw_accel == HardwareAccel.VAAPI:
-                cmd.extend(["-vaapi_device", "/dev/dri/renderD128"])
+                cmd.extend(["-vaapi_device", str(self.VAAPI_RENDER_DEVICE)])
             cmd.extend([
                 "-f", "concat",
                 "-safe", "0",
@@ -1447,7 +1450,7 @@ class PostProcessor:
                             pass
                     retry_cmd = [self._ffmpeg_path]
                     if hw_accel == HardwareAccel.VAAPI:
-                        retry_cmd.extend(["-vaapi_device", "/dev/dri/renderD128"])
+                        retry_cmd.extend(["-vaapi_device", str(self.VAAPI_RENDER_DEVICE)])
                     retry_cmd.extend([
                         "-f", "concat",
                         "-safe", "0",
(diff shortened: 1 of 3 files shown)


razzamatazm:
Nice, focused change and the Python side is clean — I checked every `renderD128` in the repo and the diagnostics path plus all four FFmpeg builders (transcode, transcode retry, concat, concat retry) do use `self.VAAPI_RENDER_DEVICE`. Full backend suite passes on the branch (1211 tests).

One real gap and two doc nits. Patches below if you want to just apply them.

## The device is configurable for FFmpeg but not for permissions

`backend/docker-entrypoint.sh:68` still derives the render group from a hardcoded `/dev/dri/renderD128`:

```bash
if [[ -e /dev/dri/renderD128 ]]; then
  RENDER_GID="$(stat -c '%g' /dev/dri/renderD128)"
```

On a host where the intended GPU is `renderD129` — exactly the case this PR is for — the app user never joins the group owning that node, so FFmpeg gets EACCES opening the device you just configured and hardware encoding still fails. It's partly masked because most multi-node hosts share one render GID, which is probably why the manual test passed.

```diff
--- a/backend/docker-entrypoint.sh
+++ b/backend/docker-entrypoint.sh
@@ -65,8 +65,11 @@ else
   usermod -u "${PUID}" -g "${PGID}" "${APP_USER}"
 fi
 
-if [[ -e /dev/dri/renderD128 ]]; then
-  RENDER_GID="$(stat -c '%g' /dev/dri/renderD128)"
+# Must match PostProcessor.VAAPI_RENDER_DEVICE, or the app user joins the
+# render group for the wrong DRM node and ffmpeg can't open the configured one.
+RENDER_DEVICE="${CATCHUP_VAAPI_RENDER_DEVICE:-/dev/dri/renderD128}"
+if [[ -e "${RENDER_DEVICE}" ]]; then
+  RENDER_GID="$(stat -c '%g' "${RENDER_DEVICE}")"
   ensure_group "render" "${RENDER_GID}"
   usermod -aG render "${APP_USER}"
 fi
```

## Docs

The variable is only in `docker-compose.yml`. Every sibling knob (`CATCHUP_FFMPEG_PATH`, `CATCHUP_FFPROBE_PATH`, `CATCHUP_COMSKIP_PATH`) is listed in the README env table, and this is user-facing so CLAUDE.md wants a CHANGELOG entry too.

```diff
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ Most users never need to touch these.
 | `CATCHUP_COMSKIP_PATH` | Path to comskip binary | *(auto-detected)* |
+| `CATCHUP_VAAPI_RENDER_DEVICE` | DRM render node used for VA-API hardware encoding; set this when your GPU is not the first render node | `/dev/dri/renderD128` |
 | `CATCHUP_DATABASE_URL` | SQLite database URL | `sqlite+aiosqlite:////app/config/catchup_dvr.db` |
```

Worth also adding a `<Config>` block to `unraid/mustarrd.xml` — `LIBVA_DRIVER_NAME` has one, and an Unraid user currently has no way to set this.

## Test coverage for the blank case

The description promises the default is preserved "when the variable is unset **or blank**", but only the `renderD129` case is covered. Extracting the module-reload into a helper makes the blank case a two-liner:

```python
    def _load_with_render_device_env(self, value):
        with patch.dict(
            os.environ,
            {"CATCHUP_VAAPI_RENDER_DEVICE": value},
            clear=False,
        ):
            spec = importlib.util.spec_from_file_location(
                "post_processor_vaapi_device_env_under_test",
                MODULE_PATH,
            )
            module = importlib.util.module_from_spec(spec)
            assert spec and spec.loader
            spec.loader.exec_module(module)

        return module

    def test_render_device_environment_override(self):
        module = self._load_with_render_device_env("/dev/dri/renderD129")

        self.assertEqual(
            module.PostProcessor.VAAPI_RENDER_DEVICE,
            Path("/dev/dri/renderD129"),
        )

    def test_blank_render_device_environment_falls_back_to_default(self):
        for value in ("", "   "):
            with self.subTest(value=value):
                module = self._load_with_render_device_env(value)

                self.assertEqual(
                    module.PostProcessor.VAAPI_RENDER_DEVICE,
                    Path("/dev/dri/renderD128"),
                )
```

## Minor

The docstring edit at `post_processor.py:1234` (`"Path to video file"` → `"Path to input file"`) is unrelated to this change and says slightly less — worth dropping from the diff.

With the entrypoint fix and the docs this is good to merge. I ran the suite with all of the above applied: 1213 pass.

matrix2669:

- Docker render-group permissions now follow `CATCHUP_VAAPI_RENDER_DEVICE` instead of hardcoding `renderD128`.
- The entrypoint also trims whitespace so its blank-value behavior matches the Python setting.
- README, CHANGELOG, Docker Compose, and Unraid configuration are covered.
- Regression coverage now includes the `renderD129` override plus empty and whitespace-only fallback values.
- The unrelated docstring edit was reverted.
- Synced the branch with current `main` and preserved the new upstream August 15 CHANGELOG entry.

Final validation on the merged tree: **1,247 backend tests passed** and **116 frontend tests passed across 13 test files**. The PR is currently mergeable; upstream Actions is still showing `action_required` with no jobs until the repository approves the fork workflow.

razzamatazm:
Thanks for this — merged, and your commits went in with authorship intact (see 48347f3).

Your diagnosis was the valuable part: hardcoding `renderD128` genuinely breaks multi-GPU hosts, and finding all four FFmpeg call sites plus the entrypoint permission problem was the bit that needed someone to actually go looking. That framing is what the final change is built on.

We ended up extending it rather than merging as-is, so you can see where your work landed:

- The render node is now also selectable in **Settings → Post-Processing → GPU device**, which lists every render node it can find and labels them by vendor. It is resolved per job from the DB rather than at import, so a change takes effect on the next recording without a container restart.
- `CATCHUP_VAAPI_RENDER_DEVICE` is still there and now acts as an override that wins over the UI choice, for headless setups. The picker greys out and explains why when it is set.
- The `card0` block next to your render-group change had the same problem you identified one line up, so the entrypoint now grants access to every render and card node present instead of the first.
- The four duplicated `-vaapi_device` sites are collapsed into one helper, with a regression test asserting only a single `/dev/dri/renderD128` literal survives in the module — the test that would have caught the original bug.

One thing worth flagging for anyone who finds this later: `renderD*` numbering is assigned in probe order and can shift across reboots or hardware changes. `/dev/dri/by-path/` symlinks are the stable naming. We have not done that yet, so a saved choice can in principle point at a different card after a reordering.

Appreciate you taking the time on it.

PR #406, opened 2026-08-11
Preserve template subdirectories in recording paths

matrix2669:
## Summary

- allow filename templates to create safe relative subdirectories
- preserve generated subdirectory paths when the download UI sends them back as `custom_filename`
- preserve the same hierarchy for scheduled recordings and completed-folder moves
- sanitize each path component independently while preventing absolute paths and `..` traversal
- add regression coverage for nested template paths, immediate downloads, schedules, and completed-folder moves

## Example

A template such as:

`TV Shows/{show}/Season {season:02d}/{show} - S{season:02d}E{episode:02d} - {title}`

now produces and preserves:

`TV Shows/Breaking Bad/Season 02/Breaking Bad - S02E05 - Breakage.ts`

instead of flattening `/` into spaces.

## Validation

The changes were manually tested against the reported download flow. Regression tests were added, but the full repository test suite has not been executed locally in this environment.

### DIFF
--- a/backend/api/schedules.py
+++ b/backend/api/schedules.py
@@ -112,7 +112,7 @@ def _sanitize_filename(name: Optional[str]) -> Optional[str]:
     filename = name
     if not filename.endswith(".ts"):
         filename += ".ts"
-    return file_namer.sanitize_filename(filename.removesuffix(".ts")) + ".ts"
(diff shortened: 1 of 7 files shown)


razzamatazm:
Superseded by #416, which cherry-picks this commit with authorship preserved and adds the filename character mapping, Windows reserved-name guard, Settings preview and CHANGELOG entry on top. Thanks @matrix2669.

PR #407, opened 2026-08-11
Preserve template subdirectories in recording paths

matrix2669:
## Summary

- allow filename templates to create safe relative subdirectories
- preserve generated subdirectory paths when the download UI sends them back as `custom_filename`
- preserve the same hierarchy for scheduled recordings and completed-folder moves
- sanitize each path component independently while preventing absolute paths and `..` traversal
- add regression coverage for nested template paths, immediate downloads, schedules, and completed-folder moves

## Example

A template such as:

`TV Shows/{show}/Season {season:02d}/{show} - S{season:02d}E{episode:02d} - {title}`

now produces and preserves:

`TV Shows/Breaking Bad/Season 02/Breaking Bad - S02E05 - Breakage.ts`

instead of flattening `/` into spaces.

## Validation

The changes were manually tested against the reported download flow. Regression tests were added, but the full repository test suite has not been executed locally in this environment.

### DIFF
--- a/backend/api/schedules.py
+++ b/backend/api/schedules.py
@@ -112,7 +112,7 @@ def _sanitize_filename(name: Optional[str]) -> Optional[str]:
     filename = name
     if not filename.endswith(".ts"):
         filename += ".ts"
-    return file_namer.sanitize_filename(filename.removesuffix(".ts")) + ".ts"
+    return file_namer.sanitize_relative_path(filename.removesuffix(".ts")) + ".ts"
 
 
 def _map_download_status(status: str) -> str:
--- a/backend/services/file_namer.py
+++ b/backend/services/file_namer.py
@@ -35,7 +35,7 @@ class FileNamer:
 
     @staticmethod
     def sanitize_filename(name: str) -> str:
-        """Remove invalid characters for filesystem."""
+        """Remove invalid characters for a single filesystem path component."""
         # Remove stylized "New" markers used in some EPG titles
         name = re.sub(r'[\[\(\-–—|:]*\s*ᴺᵉʷ\s*[\]\)]*\s*(?:-\s*)?', ' ', name)
         # Remove stylized "Live" markers frequently injected by providers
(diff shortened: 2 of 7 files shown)


razzamatazm:
This is a good change and the sanitizer is solid. I specifically tried to break the path handling, since this lets a template and a client-supplied `custom_filename` create directories — and it holds:

| input | result |
|---|---|
| `../../etc/x` | `unknown-program/unknown-program/etc/x` |
| `....//....//x`, `a/./../b` | components collapse to `unknown-program` |
| `/etc/passwd`, `//srv/x` | `etc/passwd`, `srv/x` |
| `..\..\x` | `x` |
| `C:\win\sys` | `C win sys` |
| `\\server\share\x` | `server share x` |
| `%2e%2e/..`, `‥`, `．．` | inert |

The load-bearing detail is that `sanitize_filename` ends with `.strip(' .')`, so no component can ever come out as `..`, and you split on the raw string before sanitizing rather than after — which is the ordering that usually gets this wrong. `_prune_empty_parent_dirs` re-checks containment with realpath each iteration, so that's fine too.

I also checked the rest of the pipeline for nesting assumptions and it's all clear: `download_manager.py:762` already does `os.makedirs(..., exist_ok=True)` before the only writer, post-processing derives everything from the absolute input path, the completed-folder move already handled folder-relative paths, restart recovery reads the stored absolute path, and the API uses `file_path.name` for Content-Disposition. VOD isn't an inconsistency either — it already nested via `vod_namer` and isn't template-driven.

Backend suite passes (1222). Three things before merge.

## Bug: `.ts` is stripped by substring, not suffix

`DownloadModal.jsx:105` and `ScheduleModal.jsx:105`:

```js
setFilename(data.filename.replace('.ts', ''))
```

`String.replace` with a string argument replaces the **first** occurrence. That was harmless when the value was a bare filename, but now that it can contain directory segments, a folder with `.ts` in its name gets mangled and the real extension survives — you end up re-appending at line 145 and getting `...ts.ts`.

```diff
-          setFilename(data.filename.replace('.ts', ''))
+          setFilename(data.filename.replace(/\.ts$/, ''))
```

Same one-liner in both files.

## The Settings UI doesn't mention it

This is backend-only, so the page where people actually author templates says nothing about `/` now creating folders — `renderTemplateExample` still previews it as one flat line. Suggested addition next to the existing extension alert:

```jsx
<Alert color="blue" variant="light" icon={<IconAlertCircle size={16} />}>
  A forward slash in a template creates a folder. For example{' '}
  <Code>{'TV Shows/{show}/Season {season:02d}/{show} - S{season:02d}E{episode:02d}'}</Code> files the
  recording into nested folders inside your completed folder. Slashes that come from the guide itself
  (a show called &quot;AC/DC&quot;, say) stay part of the name and don&apos;t create folders.
</Alert>
```

(`Code` is already imported in that file.)

## CHANGELOG

CLAUDE.md requires an entry for user-facing changes, and this one matters more than usual: anyone who already has a `/` in a template gets different behaviour after upgrading, since it previously flattened to a space. Worth calling that out explicitly in the entry.

## Non-blocking

- **Empty levels aren't really skipped.** `render_template_path` drops blank *template* levels before formatting but never after, so `{show}/{title}` with an empty `show` gives you an `unknown-program/` directory on disk rather than skipping the level. Pre-PR that collapsed harmlessly. Probably worth dropping components that render empty.
- **No depth cap.** Each component is capped at 200 bytes but nothing caps the count, and `custom_filename` is `String(1000)` — one request can ask for ~500 nested levels. A cap of 8 or so would be cheap.
- **Duplication.** The `.ts` suffix dance is byte-identical in `api/schedules.py:109-115` and `services/output_path.py:71-75`, and this one-line policy change had to be made in both. Second occurrence suggests a shared helper wants to exist.
- `schedules.py:_sanitize_filename` now returns a multi-component relative path — the name no longer says what it does.

With the `.ts` fix, the Settings hint and a CHANGELOG entry this is good to go. I ran both suites with those applied: 1222 backend + 102 frontend pass.

matrix2669:
Closing this as superseded by #416, which incorporated the template-subdirectory work into upstream with additional sanitizer and Settings improvements. Thanks for carrying the original implementation forward. I’m moving the two small follow-ups found after merge (the modal `.ts` suffix handling and the Windows reserved-device-stem-with-extension case) to a fresh branch based on current upstream `main`.

PR #408, opened 2026-08-11
Preserve structured XMLTV metadata in fresh EPG results

matrix2669:
## Summary

- preserve structured XMLTV programme metadata instead of collapsing it to a single category
- store subtitle, all categories, season/episode metadata, and Gracenote/TVDB/TMDB/IMDb identifiers on EPG rows
- classify episodic sports programming as TV while reserving `sports` for actual sporting events
- enrich sparse fresh Xtream EPG responses from matching SQLite rows so Browse gets the same structured metadata as stored/search results
- canonicalize live EPG entries to the requested Xtream stream ID when providers return a separate XMLTV/channel identifier
- keep live provider timing and archive state authoritative during enrichment
- honor the configured TV template for structured episodes even when the EPG omits an episode subtitle
- upgrade existing SQLite databases in place with additive structured EPG columns

## Why

Some Xtream providers expose rich XMLTV data but return only title/description/timing from the live EPG endpoint. Browse prefers that fresh endpoint, which caused programmes that were correctly identified in stored EPG/search results to lose their category, season/episode, and external IDs.

This change keeps the fresh provider response for current timing/archive state, then fills only missing metadata from the exact stored airing (`account_id`, stream/channel ID, start, stop).

It also improves classification from the richer XMLTV data. In particular, `Sports + Series` or structured S/E metadata is treated as episodic TV, while matchup/event-style programmes remain sports events.

## Examples

- `30 for 30` / `MLB Tonight` with Series + S/E metadata → `tv_show`
- `The Herd With Colin Cowherd` with structured episode metadata → `tv_show`
- `MLB Baseball — Toronto Blue Jays at New York Yankees` → `sports`
- `NFL Football` matchup-style airing → `sports`
- `News + Series` programmes remain `news`

## Compatibility

- existing `category` remains available as the primary/first category
- schema changes are additive and applied automatically for existing SQLite installations
- unmatched live EPG entries retain the existing behavior
- live `has_archive`, title/description, and timing values are not overwritten by stored metadata

## Tests

Regression coverage includes:

- multi-category/structured XMLTV classification
- Sports+Series vs. actual sports events
- sparse live EPG enriched from a rich matching SQLite row
- providers where live `channel_id` differs from the Xtream `stream_id`
- preserving live `has_archive=false`
- custom TV-template handling for structured episodes without subtitles


razzamatazm:
Thanks for this — the metadata work is genuinely valuable and the problem it solves is real. I'd like to take it, but not in this shape. Two blocking things and a structural one.

## Blocking: the branch is red

Two pre-existing tests in `backend/tests/test_detect_program_type.py` fail on this branch:

- `test_roman_numeral_not_sports` — `"Henry V"` now classifies as **sports**. `_MATCHUP_PATTERN` (`structured_epg.py:39`) includes a bare `\bv\.?\b` alternative, so any standalone roman numeral reads as "team v team". Requiring the separator to be flanked on both sides (`r"\w\s+(?:vs?\.?|at|@)\s+\w"`) fixes it and still matches "Spurs at Knicks", "Lakers vs. Celtics", "Arsenal v Chelsea".
- `test_sports_channel_category` — a generic programme on a `US Sports` channel drops from **sports** to **other**. Tightening sports classification is the right goal, but "other" is worse than the category the provider actually gave us. Note this also contradicts your own new `test_sports_channel_paid_programming_is_not_event`. Both can hold: fall back to sports when nothing else matches, but exclude recognised schedule filler (paid programming, to be announced, off air).

## Blocking: the service is monkeypatched rather than edited

`backend/services/__init__.py` imports two modules purely for side effects, which rebind seven core methods at import time:

```
EPGService._process_epg_entry     (patched twice, load-order dependent)
EPGService.serialize_program
EPGService.detect_program_type
EPGIngestManager._iter_programs
FileNamer.generate_filename       (patched twice, across two modules)
ingest_module._program_insert_stmt
```

Reading `epg_service.py` now tells you the wrong thing about what the app does, and the behaviour depends on import order. Please edit the methods in place instead.

Related: `structured_epg_bootstrap.py` adds a second migration path (a raw `connect` event listener running `ALTER TABLE`), and `structured_epg.py` has a third, never-called version of the same migration. CLAUDE.md documents one place for this — the lightweight `ALTER TABLE` checks in `backend/database.py`, which already handle `epg_programs`. The ten new columns belong there.

## Scope

This is nine distinct user-visible changes in one PR: the merge/enrichment, the schema, the classification rewrite, structured filename generation, the `{tmdb}` token, the season-zero-to-airing-year rule, the XMLTV `-1` precedence rule, and the tag separator fix. Could you split it into roughly three — schema + ingest, classification, filename/template — each with its own CHANGELOG entry? The current entry only covers the first.

Two smaller notes while splitting:

- `_merge_live_and_stored_programs` doesn't only enrich sparse entries as the description says; it appends every stored-only row and returns the union, which widens the guide window. That's a separate product decision worth its own PR.
- `_with_annual_series_season` (`structured_tv_template.py:136`) triggers on any `season_number == 0` with a positive episode, not on evidence of raw XMLTV `-1`. A genuine special arriving as a plain season-0 field with no onscreen marker gets silently renamed to the airing year.

The tests you wrote are good and should port across the split mostly unchanged.

razzamatazm:
One more thing, since I just left you three fairly dense reviews in a row and I don't want the tone to land wrong.

I want all three of these. 405 and 407 are close — small fixes and they're in. 408 I want too; the structured metadata work is the most valuable of the batch and it's solving a real problem I'd have gotten to eventually. My asks there are about how it's wired in, not whether it belongs.

No rush on any of it. If splitting 408 is a pain or you'd rather talk through where the seams should go before you start moving code, say so and we'll work it out here first — that's cheaper than you doing the split twice.

Thanks for these.

razzamatazm:
Closing this one, and I want to be clear about why, because it is not a rejection on merit.

I took the split we talked about and ran it myself rather than leaving you to do it. It landed today as three stacked PRs:

- #422 — the metadata module, additive nullable columns on `epg_programs`, XMLTV ingest, storage
- #423 — filling sparse live guide responses from stored entries, which is the Browse-versus-search discrepancy
- #424 — recording filenames using the real season and episode

The spec I worked from is #421, which is written up from this PR. Its "Further Notes" section says as much: the parts kept are the parts that reviewed well here, and they are the core of what you wrote. Parsing the whole `<programme>` element once, keeping every category with the first still in `category`, storing season and episode plus the external IDs, and filling the gaps when a provider answers the live endpoint with title and timing only. That was the valuable idea and it is now in main.

What did not come across, and why:

- **Sports and program-type reclassification.** This is the `_MATCHUP_PATTERN` work and the "Henry V" and `US Sports` regressions from my earlier review. I ruled it out of #421 deliberately. Using program categories and structured episode data to reclassify episodic sports programming is a real improvement and I still want it, but it is riskier than the storage work and deserves its own change with its own regression tests rather than riding along with the metadata plumbing.
- **TMDB, TVDB and IMDb lookups, and a `{tmdb}` filename token.** Identifiers are now stored when the provider publishes them, but nothing does a network lookup. The token is worth doing and needs its own spec.
- **Deriving a season from the airing year** when the provider signals an unknown season. Ruled out on purpose: it writes a guess into the season field and then out into filenames.

One thing worth flagging since it came out of your original diff and I would not have found it otherwise: scheduled recordings save their own snapshot of an airing and rebuild the program at dispatch, so the guide metadata was gone by the time the file was named. #424 fixes that by looking it back up from the stored guide entry for that exact airing.

If you want to pick up the sports reclassification as its own PR, I would take it. Same for the `{tmdb}` token. Thanks for this one, it was the most valuable of the batch and it is in the product now.

